PR #807, opened 2014-10-07
Fix up icid-values  some more.

### LINKED ISSUE #486: P-Charging-Vector icid-value parameters are rejected unnecessarily
If bono needs to build a P-Charging-Vector, it uses the Call-ID as an icid-value.  (P-CSCFs are free to choose their own identifier, as long as it has the correct lifecycle.)

Jitsi generates a Call-ID including colons (maybe a hypothetical IPv6 address), and bono puts this on the P-Charging-Vector.

    P-Charging-Vector: icid-value=a2bb639b437cd5827a8f54fe39f3987c0:0:0:0:0:0:0:0;icid-generated-at=ec2-54-237-198-247.compute-1.amazonaws.com

sprout fails to parse this - it expects icid-value to be a normal parameter when actually it's defined to be a `gen-value`.  sprout should parse this.

rkday:
Fixes #486.

Most of the issues have already been resolved, by https://github.com/Metaswitch/sprout/commit/e0e9830b3a3af390bc9a3da205c7e4b9f742aa14#diff-2ef3671b84bf856b95dce4da51471524 - which quotes the icid-value in all outgoing headers.

However, I've:
* added a couple more unit tests for various icid-values (a hostname, an IPv6 address, and the original example from #486)
* fixed up the case where the icid-value was just an IPv6 address (where its enclosing [ and ] weren't preserved)
* fixed up the case where the Call-ID can contain one double quote (which previously got transposed to the icid-value and messed up the quoting), and updated the unit tests to cover that case

UTed but not tested live.


sprout/custom_headers.cpp
```diff
@@ -588,7 +588,17 @@ pjsip_hdr* parse_hdr_p_charging_vector(pjsip_parse_ctx* ctx)
 
   // Parse the required icid-value parameter first.
   pjsip_parse_param_imp(scanner, pool, &name, &value,
-                        PJSIP_PARSE_REMOVE_QUOTE);
+                        NULL);
+
+  // Strip the quotes of manually instead of using
+  // PJ_PARSE_REMOVE_QUOTE. This preserves the square bracket on IPv6
+  // addresses.
+  if (value.ptr[0] == '"' && (value.slen > 2))
```
eleanor-merry: Is this safe? Can the last character ever not be a "? 
Also *off

sprout/ut/custom_headers_test.cpp
```diff
@@ -165,8 +165,100 @@ TEST_F(CustomHeadersTest, PChargingVectorQuotedIcidValue)
...
-  EXPECT_EQ(written, 140);
-  EXPECT_STREQ("P-Charging-Vector: icid-value=\"4815162542\";orig-ioi=homedomain;term-ioi=remotedomain;icid-generated-at=edge.proxy.net;other-param=test-value", buf);
+  EXPECT_EQ(written, 177);
+  EXPECT_STREQ("P-Charging-Vector: icid-value=\"a2bb639b437cd5827a8f54fe39f3987c0:0:0:0:0:0:0:0\";orig-ioi=homedomain;term-ioi=remotedomain;icid-generated-at=edge.proxy.net;other-param=test-value", buf);
+}
+
+TEST_F(CustomHeadersTest, PChargingVectorHostnameIcidValue)
+{
+  string str("INVITE sip:6505554321@homedomain SIP/2.0\n"
+
```
eleanor-merry: Extra line (and in test below as well)

sprout/ut/custom_headers_test.cpp
```diff
@@ -165,8 +165,100 @@ TEST_F(CustomHeadersTest, PChargingVectorQuotedIcidValue)
...
+  EXPECT_PJEQ(pcv->icid_gen_addr, "edge.proxy.net");
+  EXPECT_EQ(1u, pj_list_size(&pcv->other_param));
+
+  // Test the icid-value has its quotes replaced after printing.
+  char buf[1024];
+  hdr = (pjsip_hdr*)pcv;
+  int written = hdr->vptr->print_on(hdr, buf, 0);
+  EXPECT_EQ(written, -1);
+  int i = 1;
+  while ((written == -1) && (i <= 1024)) {
```
eleanor-merry: Bracket style

eleanor-merry:
Looks good, minor comments inline

sprout/ut/custom_headers_test.cpp
rkd-msw: This is wrong throughout - I'll fix up in a separate PR.

PR #912, opened 2014-12-23
[Reviewer: Mike] Handle 6xx responses by cancelling outstanding requests

### LINKED ISSUE #757: Sprout doesn't handle 6xx responses correctly
[RFC 3261](https://www.ietf.org/rfc/rfc3261.txt), section 16.7, point 5 says

> If a 6xx response is received, it is not immediately forwarded, but the stateful proxy SHOULD cancel all client pending transactions as described in Section 10, and it MUST NOT create any new branches in this context.

Sprout handles 6xx responses just like 4xx and 5xx responses and allows the other client pending transactions to continue.

mirw:
Mike,

Please can you review my fix to handle 6xx responses by cancelling outstanding requests?  This fixes #757.

I've UT-ed this, but one thing I wanted to check with you - the old code prioritized 487 responses over all other responses.  I can't see anything in RFC 3261 that says it should do.  Do you know why the old code did that?

Matt

### DIFF
--- a/sprout/basicproxy.cpp
+++ b/sprout/basicproxy.cpp
@@ -1071,6 +1071,18 @@ void BasicProxy::UASTsx::on_new_client_response(UACTsx* uac_tsx,
         LOG_DEBUG("%s - All UAC responded", name());
         on_final_response();
       }
+      else if (PJSIP_IS_STATUS_IN_CLASS(status_code, 600))
+      {
+        // From RFC 3261, section 16.7, point 5:
+        // > If a 6xx response is received, it is not immediately forwarded,
+        // > but the stateful proxy SHOULD cancel all client pending
+        // > transactions as described in Section 10, and it MUST NOT create
+        // > any new branches in this context.
+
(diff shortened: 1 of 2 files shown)


mike-evans:
Matt

The rationale for favouring 487 responses was that, if the user has cancelled the call, 487 is what they are expecting to see.  With your change I think that this case won't work as expected as if one leg returns a 4xx response less than 487 it will get priority over other legs which are cancelled.  It used to be the case that this simple prioritising of all 487 responses gave the desired behaviour as the proxy code would only generate CANCEL requests locally when the upstream transaction had already ended, but your 600 handling code breaks this.  I therefore think we need a flag to record whether the CANCEL processing has been triggered locally or whether it was a response received on the upstream UAS transaction, and to prioritize 487s in the latter case only.

Otherwise fix looks good.

Mike

mirw:
Thanks!  As discussed, there isn't a strong argument either way but in the interests of retaining the previous behavior (except where we know it's explicitly wrong), I'll prioritize 487 over 3-5xx, but 6xx takes precedence over it.

PR #925, opened 2015-01-28
Reviewer Matt - Create new tdata before re-try if TCP is super congested

### LINKED ISSUE #919: Call traffic impacted when a Sprout is isolated from the network
While running 2 CPS traffic on a duplex Clearwater configuration, Sprout-1 was isolated from the network (port disconnected from vSwitch). Calls start failing mostly with 503 Service Unavailable errors. This happens when Sprout-2 tries to use Sprout-1 as the terminating S-CSCF after an SRV lookup. The attempted TCP connection times out after 32S with a transport failure. Sprout-2 attempts to lookup another S-CSCF from Homestead, but only the local clustered record is provisioned (which was just attempted and failed) so a 503 is issued. These errors persist as long as Sprout-1 remains isolated from the network.
	
The same behaviour is also observed if Sprout-1 is simply powered off as well.

bossmc:
As discussed, if the TCP transport is really congested it's possible that the original send is still pending in the transport layer and thus the `tdata` is not (exclusively) owned by the transaction layer or the basic proxy layer.  In this case we simply clone the old `tdata` and replace it with it's own clone, dropping our reference (the transaction has it's own reference).  The transaction will eventually terminate at some point and clean up the abandoned `tdata`.

Tested live by @plfrancis to confirm that this fixes #919.

### DIFF
--- a/sprout/basicproxy.cpp
+++ b/sprout/basicproxy.cpp
@@ -1981,6 +1981,21 @@ bool BasicProxy::UACTsx::retry_request()
     // future events from it.
     LOG_DEBUG("Attempt to retry request to alternate server");
     pjsip_transaction* retry_tsx;
+
+    // In congestion cases, the old tdata might still be held by PjSIP's
+    // trasport layer waiting to be sent.  Therefore it's not safe to re-send
+    // the same tdata, so we should clone it first.
+    // LCOV_EXCL_START - No congestion in UTs
+    if (_tdata->is_pending)
+    {
+      pjsip_tx_data* old_tdata = _tdata;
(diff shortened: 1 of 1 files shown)


mirw:
As discussed, I don't think this is the right long-term fix, but I agree it's worth having in the short-term.  Can we leave the original issue open (or clone it) to fix this properly?

bossmc:
I've commented in #919 that this is an incomplete fix and (depending on how #919 continues) I'll either leave that as the place-holder or split it into a specific issue (and possibly raise with Teluu).

PR #926, opened 2015-01-29
[Reviewer: Ellie] Memcached schema upgrade

AlexHockey:
Hi Ellie, please can you review my changes to migrate Sprout's RegStore to using JSON for its records. 

This PR contains the bulk of the changes. These should match the design you reviewed, with two changes (which we discussed):

* I haven't implemented the change to produce two different packages based on a built time option (this is not needed for the Project Clearwater release plan). 
* We want to continue use the binary format for the time being. As our packages never supply `--memcached-write-format` I have done this by simply defaulting the write format to binary. 

## Testing

* UTs - in this PR.
* Live tests - I am currently running through the live test plan. This is as you reviewed but with the build-time options test removed, and a couple of tests to check the config option is honored when specified. 

## Docs

I plan to write the upgrade instructions (for future release notes). I will send these to you for review when I have written them.


include/regstore.h
```diff
@@ -250,10 +308,39 @@ class RegStore
...
+           std::vector<SerializerDeserializer*>& deserializers,
+           ChronosConnection* chronos_connection);
+
+  /// Alternative RegStore constructor that creates a RegStore using just the
+  /// default (de)serializer.
+  ///
+  /// @param data_store         - Pointer to the underlying data store.
+  /// @param chronos_connection - Chronos connection used to set timers for
+  ///                             expiring registrations and subscriptions.
+  RegStore(Store* data_store,
```
eleanor-merry: Why do we still need this version?

sprout/main.cpp
```diff
@@ -108,62 +108,64 @@ enum OptionTypes
...
+  { "max-call-list-length",         required_argument, 0, OPT_MAX_CALL_LIST_LENGTH},
+  { "memento-threads",              required_argument, 0, OPT_MEMENTO_THREADS},
+  { "call-list-ttl",                required_argument, 0, OPT_CALL_LIST_TTL},
+  { "alarms-enabled",               no_argument,       0, OPT_ALARMS_ENABLED},
+  { "log-level",                    required_argument, 0, 'L'},
+  { "daemon",                       no_argument,       0, 'd'},
+  { "interactive",                  no_argument,       0, 't'},
+  { "help",                         no_argument,       0, 'h'},
+  { "memcached-write-format",       required_argument, 0, OPT_MEMCACHED_WRITE_FORMAT},
   { NULL,                0, 0, 0}
```
eleanor-merry: Can you tidy up this line as well

sprout/main.cpp
```diff
@@ -1496,8 +1539,27 @@ int main(int argc, char* argv[])
     }
 
     // Create local and optionally remote registration data stores.
-    local_reg_store = new RegStore(local_data_store, chronos_connection);
-    remote_reg_store = (remote_data_store != NULL) ? new RegStore(remote_data_store, chronos_connection) : NULL;
+    RegStore::SerializerDeserializer* serializer;
```
eleanor-merry: It's worth a comment here to explain why it's OK to reuse the serializer/deserializers

sprout/regstore.cpp
```diff
@@ -369,179 +399,40 @@ void RegStore::expire_subscriptions(AoR* aor_data,
...
-
-    // Now extract the various fixed binding parameters.
-    getline(iss, b->_uri, '\0');
-    getline(iss, b->_cid, '\0');
-    iss.read((char *)&b->_cseq, sizeof(int));
-    iss.read((char *)&b->_expires, sizeof(int));
+    SerializerDeserializer* deserializer = *it;
 
-    iss.read((char *)&b->_priority, sizeof(int));
+    LOG_DEBUG("Try to deserialize record with '%s' deserializer",
```
eleanor-merry: These debug logs would be more helpful if they included the aor_id

sprout/regstore.cpp
```diff
@@ -810,3 +716,463 @@ std::string RegStore::AoR::Binding::pub_gruu_quoted_string(pj_pool_t* pool) cons
...
+//
+
+RegStore::AoR* RegStore::BinarySerializerDeserializer::
+  deserialize_aor(const std::string& aor_id, const std::string& s)
+{
+  std::istringstream iss(s, std::istringstream::in|std::istringstream::binary);
+
+  // First off, try to read the number of bindings.
+  int num_bindings;
+  iss.read((char *)&num_bindings, sizeof(int));
```
eleanor-merry: Can you use char* instead

sprout/regstore.cpp
```diff
@@ -810,3 +716,463 @@ std::string RegStore::AoR::Binding::pub_gruu_quoted_string(pj_pool_t* pool) cons
...
+
+  if (iss.eof())
+  {
+    // Hit an EOF which means the record is corrupt.
+    LOG_INFO("Could not deserialize AOR - EOF reached");
```
eleanor-merry: Why is this a LOG_INFO?

sprout/regstore.cpp
```diff
@@ -810,3 +716,463 @@ std::string RegStore::AoR::Binding::pub_gruu_quoted_string(pj_pool_t* pool) cons
...
+  {
+    JSON_ASSERT_CONTAINS(doc, JSON_BINDINGS);
+    JSON_ASSERT_OBJECT(doc[JSON_BINDINGS]);
+    const rapidjson::Value& bindings_obj = doc[JSON_BINDINGS];
+
+    for (rapidjson::Value::ConstMemberIterator bindings_it = bindings_obj.MemberBegin();
+         bindings_it != bindings_obj.MemberEnd();
+         ++bindings_it)
+    {
+      LOG_DEBUG("  Binding: %s", bindings_it->name.GetString());
```
eleanor-merry: Does the binding ID not need asserting on as well?

sprout/ut/regstore_test.cpp
```diff
@@ -50,11 +50,30 @@
...
-class RegStoreTest : public SipTest
+#include "mock_store.h"
+
+using ::testing::_;
+using ::testing::DoAll;
+using ::testing::Return;
+using ::testing::SetArgReferee;
+
+// These tests use "typed tests" to run the same tests over different
+// (de)serializers. For more infomration see:
```
eleanor-merry: *information

eleanor-merry:
Looks good - minor comments inline

include/regstore.h
AlexHockey: Yes, for UT. I will add a comment.

sprout/regstore.cpp
AlexHockey: As discussed, I think this is more interesting than a plain debug log would suggest, but shouldn't be any higher severity as it is on the call path.

PR #928, opened 2015-02-04
[Reviewer: Ellie] I-CSCF should route to BGCF when appropriate

### LINKED ISSUE #927: I-CSCF never routes to BGCF
If Sprout as an I-CSCF finds that the subscriber is not known to the platform, it rejects the call.  If the URI is a TEL URI (after ENUM) or a non-local SIP URI, Sprout should instead route to the BGCF.

This breaks tandem function.

mirw:
Ellie,

Please can you review my enhancement to fix #927?  It should be reasonably self-explanatory.

I've UT-ed and live-tested.

Thanks,

Matt

### DIFF
--- a/sprout/icscfsproutlet.cpp
+++ b/sprout/icscfsproutlet.cpp
@@ -50,6 +50,7 @@ extern "C" {
 
 #include "log.h"
 #include "pjutils.h"
+#include "stack.h"
 #include "sproutsasevent.h"
 #include "icscfsproutlet.h"
 #include "scscfselector.h"
@@ -60,7 +61,8 @@ extern "C" {
 #define MAX_ENUM_LOOKUPS 2
 
 /// Constructor.
-ICSCFSproutlet::ICSCFSproutlet(int port,
+ICSCFSproutlet::ICSCFSproutlet(const std::string& bgcf_uri,
+                               int port,
                                HSSConnection* hss,
                                ACRFactory* acr_factory,
                                SCSCFSelector* scscf_selector,
@@ -68,6 +70,7 @@ ICSCFSproutlet::ICSCFSproutlet(int port,
                                bool enforce_global_only_lookups,
                                bool enforce_user_phone) :
   Sproutlet("icscf", port),
+  _bgcf_uri(NULL),
   _hss(hss),
   _scscf_selector(scscf_selector),
   _acr_factory(acr_factory),
@@ -75,6 +78,13 @@ ICSCFSproutlet::ICSCFSproutlet(int port,
   _global_only_lookups(enforce_global_only_lookups),
   _user_phone(enforce_user_phone)
 {
+  // Convert the BGCF routing URI to a form suitable for PJSIP, so we're
+  // not continually converting from a string.
+  _bgcf_uri = PJUtils::uri_from_string(bgcf_uri, stack_data.pool, false);
+  if (_bgcf_uri == NULL)
+  {
+    LOG_ERROR("Invalid BGCF URI %s", bgcf_uri.c_str()); //LCOV_EXCL_LINE
+  }
 }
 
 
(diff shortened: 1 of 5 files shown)


rkd-msw:
Does this work for the case where the I-CSCF and S-CSCF aren't colocated? I think we've previously only needed the BGCF Sproutlet to be running alongside the S-CSCF Sproutlet (and treated it as effectively an internal mechanism, so we don't document bgcf.homedomain as something you need to set up in DNS).

Does this change make it necessary to install a BGCF on I-CSCF nodes (with associated bgcf.json), or for the I-CSCF on node A to route directly to the BGCF on node B (bypassing the S-CSCF)?

mirw:
Good spot!  I think we need to

*   make the BGCF URI default to `bgcf.<sprout cluster>` not `bgcf.<scscf uri>`
*   document that if you separate I- and S-CSCF, you must still put `bgcf.json` on both
*   talk to the Product Owner to see if we should have a user story to support separate BGCF.

mirw:
Actually, it turns out that the BGCF URI should be `bgcf.<scscf uri>` because that's what the BGCF service registers as, even when it's running on I-CSCF (and likewise the I-CSCF service registers as `icscf.<scscf uri>`).  I'm going to raise a separate issue about that, and just fix up the docs.

mirw:
@eleanor-merry, please can you review?

sprout/icscfsproutlet.cpp
```diff
@@ -75,6 +78,13 @@ ICSCFSproutlet::ICSCFSproutlet(int port,
   _global_only_lookups(enforce_global_only_lookups),
   _user_phone(enforce_user_phone)
 {
+  // Convert the BGCF routing URI to a form suitable for PJSIP, so we're
+  // not continually converting from a string.
+  _bgcf_uri = PJUtils::uri_from_string(bgcf_uri, stack_data.pool, false);
+  if (_bgcf_uri == NULL)
+  {
+    LOG_ERROR("Invalid BGCF URI %s", bgcf_uri.c_str()); //LCOV_EXCL_LINE
```
eleanor-merry: Shouldn't this cause the I-CSCF to exit? If we try and route to the BGCF we'll crash (as we don't check it isn't NULL)

sprout/icscfsproutlet.cpp
```diff
@@ -506,8 +517,8 @@ void ICSCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
         (!PJUtils::is_uri_gruu(uri)))
     {
       LOG_DEBUG("enforce_user_phone set to false, try using a tel URI");
-      req->line.req.uri =
```
eleanor-merry: This log makes no sense

sprout/icscfsproutlet.cpp
```diff
@@ -527,9 +538,21 @@ void ICSCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
...
           (translate_tel_uri(req, pool)))
       {
-        // If we successfully translate the req URI, we should look for an
-        // S-CSCF again.
-        status_code = (pjsip_status_code)_router->get_scscf(pool, scscf_sip_uri);
+        // If we successfully translate the req URI and end up with either another TEL URI or a
+        // local SIP URI, we should look for an S-CSCF again.
+        uri = req->line.req.uri;
+        if ((PJSIP_URI_SCHEME_IS_TEL(uri)) ||
+            ((PJSIP_URI_SCHEME_IS_SIP(uri)) &&
```
eleanor-merry: Doesn't this need to check that the SIP URI is numeric?

eleanor-merry:
Looks good, some comments inline

sprout/icscfsproutlet.cpp
mirw: As discussed, it's actually pretty accurate, and the comment above elaborates on it, so I'm leaving this as-is.

sprout/icscfsproutlet.cpp
mirw: No.  What we're doing here is looking it up in the HSS - it could have been resolved to `sip:alice@example.com` and we'd still want to do this lookup.

sprout/icscfsproutlet.cpp
mirw: It's a valid point, but we have no way of cleanly exiting here.  This is a common problem across all our sproutlets.  I've raised #935 to track this.

PR #930, opened 2015-02-06
Use ipc instead of tcp for 0MQ alarm interface

dtheobald:


AlexHockey:
Looks good. I will hold off merging until https://github.com/Metaswitch/cpp-common/pull/234 is merged.

PR #933, opened 2015-02-09
[Reviewer: Ellie] Add missing command line option

AlexHockey:
Hi Ellie. When I enhanced the RegStore to support JSON, I forgot to update sprout's usage statement with the new command line argument. This PR fixes that. 

I tested that sprout builds and includes the option when running `sprout -h`


eleanor-merry:
Looks good

PR #936, opened 2015-02-10
[Reviewer: Ellie] Make Bono always route to Sprout on initial access-proxy requests

### LINKED ISSUE #842: Bono does not forward all requests upstream
Suppose I have a user, `alice@some-network.com` who is registered with Sprout through Bono. If she calls `bob@another-network.com`, the INVITE is received by Bono and routed directly to `another-network.com`. Instead Bono should route to Sprout to apply Alice's originating services. 

From discussion with @bossmc I understand that Bono is trying to act as a SIP proxy, and will only forward the request upstream to Sprout if there are no other hops to process and the request targets the home domain. However bypassing the originating network entirely is wrong.

rkd-msw:
Fixes #842. The key change is https://github.com/Metaswitch/sprout/compare/upstream_routing?expand=1#diff-c70d752a0ba2b041697dce70b56ab72eR1480 (which forces Sprout to be the target for initial requests).

I've also had to rearrange some of the set_target logic - PJSIP doesn't like creating a transaction for tel: Request-URIs with no Route headers, so I had to make sure that the Route headers were added first.

UTed, about to live test (by retrying my failing tel: live test).

### DIFF
--- a/sprout/stateful_proxy.cpp
+++ b/sprout/stateful_proxy.cpp
@@ -181,6 +181,8 @@ PJUtils::host_list_t trusted_hosts(&PJUtils::compare_pj_sockaddr);
 static pj_bool_t proxy_on_rx_request(pjsip_rx_data *rdata );
 static pj_bool_t proxy_on_rx_response(pjsip_rx_data *rdata );
 
+void set_target_on_tdata(const struct Target& target, pjsip_tx_data* tdata);
+
 static pjsip_module mod_stateful_proxy =
 {
   NULL, NULL,                         // prev, next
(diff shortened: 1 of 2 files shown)


sprout/stateful_proxy.cpp
```diff
@@ -1474,6 +1476,15 @@ int proxy_process_access_routing(pjsip_rx_data *rdata,
               PJUtils::add_asserted_identity(tdata, aid2);
             }
           }
+
+          bool initial_request = (rdata->msg_info.to->tag.slen == 0);
+          if (initial_request)
```
eleanor-merry: Why not just `if (rdata->msg_info.to->tag.slen == 0)` with an appropriate comment?

eleanor-merry:
Looks good, one minor comment

PR #939, opened 2015-02-13
[Reviewer: Matt] Check whether the plugins load

### LINKED ISSUE #935: Misconfigured Sproutlets crash
Both SCSCFSproutlet and ICSCFSproutlet parse SIP URIs and, if they fail, log, but otherwise leave NULL pointers around (so will probably crash when they're invoked).  It would be tidier if they signaled back to the parent that they have been misconfigured and caused clean termination.  Unfortunately, there doesn't seem to be a way to do this at present.

eleanor-merry:
Matt, can you please review this change to check whether a plugin has loaded correctly, and exiting if not. 

Testing by making the scscf-uri invalid, and checking that Sprout died

Fixes #935

### DIFF
--- a/include/icscfsproutlet.h
+++ b/include/icscfsproutlet.h
@@ -80,6 +80,8 @@ public:
 
   virtual ~ICSCFSproutlet();
 
+  bool init();
+
   SproutletTsx* get_tsx(SproutletTsxHelper* helper,
                         const std::string& alias,
                         pjsip_msg* req);
@@ -136,6 +138,9 @@ private:
   bool _global_only_lookups;
   bool _user_phone;
   bool _override_npdi;
+
+  /// String versions of cluster URIs
+  std::string _bgcf_uri_str;
 };
 
 
--- a/include/pluginloader.h
+++ b/include/pluginloader.h
@@ -52,8 +52,10 @@ public:
   PluginLoader(const std::string& path, struct options& opt);
   ~PluginLoader();
 
-  /// Loads the plug-ins and returns the resulting list of Sproutlets.
-  void load(std::list<Sproutlet*>& sproutlets);
+  /// Loads the plug-ins and populates the resulting list of Sproutlets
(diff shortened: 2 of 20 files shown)


mirw:
Looks good.  Are there any plugins outside sprout (e.g. gemini/memento) that we need to fix up?

mirw:
When testing, did you confirm that the correct log was generated?

eleanor-merry:
The gemini/memento plugin updates are in this review. The memento plugin will return an error on load if the cassandra store hits an unexpected error (which isn't just that cassandra hasn't started yet), the gemini plugin won't (currently) return an error on load. 

If a plugin fails to load, /var/log/sprout/ contains:
```
Debug scscfsproutlet.cpp:96:   S-CSCF cluster URI = sip:sprout.em7.cw-ngv.com:5054;transport=TCP
Debug scscfsproutlet.cpp:97:   S-CSCF node URI    = sip:10.191.159.82:5054
Debug scscfsproutlet.cpp:98:   I-CSCF URI         = sip:sprout.em7.cw-ngv.com:5052;transport=TCP
Debug scscfsproutlet.cpp:99:   BGCF URI           = sip:bgcf.sprout.em7.cw-ngv.com:5054;transport=TCP
Error scscfsproutlet.cpp:125: Invalid BGCF URI sip:bgcf.sprout.em7.cw-ngv.com:5054;transport=TCP
Error pluginloader.cpp:101: Failed to successfully load plug-in /usr/share/clearwater/sprout/plugins/sprout_scscf.so
Status pluginloader.cpp:148: Finished loading plug-ins
Error main.cpp:1637: Failed to successfully load plug-ins
```
and syslog has:
```
2038 - Description: One or more plugins failed to load. @@Cause: The service is no longer available. @@Effect: The application will exit and restart until the problem is fixed. @@Action: Check the configuration in /etc/clearwater/config
```

Issue #749, opened 2014-09-01
Notify XML bodies are not encoded


rkd-msw:
The PJSIP XML methods we use don't do entity encoding of things like `<`. This means that if any of our fields contain a literal `<` or `>`, the XML becomes invalid.

We fixed one occurrence of angle brackets in https://github.com/Metaswitch/sprout/issues/637, and I've just patched another common one - but we're still vulnerable to e.g. Contact URIs that contain `;param="</uri>"`, so we really need to manually entity-encode everything before using it, or switch to an XML builder that does. (I couldn't see anything indicating RapidXML does this.)

PR #942, opened 2015-02-17
Reviewer RKD: [DO NOT MERGE] Escape necessary fields in reg-event XML

graemerobertson:
Hi Rob,

Please can you review my changes to escape characters correctly in our XML that we build for the NOTIFY reg-event package? This fixes https://github.com/Metaswitch/sprout/issues/749.

I've also changed the pub_gruu_pj_str method to be pub_gruu_str method since it only seemed to be return a pj_str for the purpose of building this xml, and I now want it in std::string form first. I hope this will become clear when you look at the code, and I hope this is sensible.

I think I'm escaping all the necessary fields, and no unnecessary fields. I've been using the following website to help - http://www.in2eps.com/fo-abnf/tk-fo-abnf-siprules.html.

Testing wise, all I've done is run the existing UTs. I had to fix up one UT that was expecting unescaped '<' and '>'. Otherwise, the xml_escape method has its own UTs so I deemed that sufficient.

I won't merge until after the release has been cut now.

Thanks,
Graeme

### DIFF
--- a/include/regstore.h
+++ b/include/regstore.h
@@ -114,7 +114,7 @@ public:
       bool _emergency_registration;
 
       pjsip_sip_uri* pub_gruu(pj_pool_t* pool) const;
-      pj_str_t pub_gruu_pj_str(pj_pool_t* pool) const;
+      std::string pub_gruu_str(pj_pool_t* pool) const;
       std::string pub_gruu_quoted_string(pj_pool_t* pool) const;
     };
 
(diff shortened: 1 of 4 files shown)


rkd-msw:
Looks good, but suggest you test by running this against a Perimeta that's configured to do this subscribe/notify flow to make sure we haven't broken interop.

graemerobertson:
Fine, I'll check that on the staging system once I've checked it in.

graemerobertson:
Checked.

PR #944, opened 2015-02-18
[Reviewer: Rob] Make load monitor options configurable

eleanor-merry:
Rob, can you please review this change to make the load monitor options configurable. As discussed, I've not made the default initial token rate based on the number of cores, but I have bumped it up to 100 (which was the original plan). 

Tested by spinning up a Sprout/Bono node, and checking that the values get passed through correctly. 

If this is OK, I'll make the same change to Homestead, Ralf and Memento. 

This is part of the fix for https://github.com/Metaswitch/cpp-common/issues/199


sprout/main.cpp
```diff
@@ -265,6 +266,12 @@ static void usage(void)
        "                            The session expiry period to request (in seconds)\n"
        "     --target-latency-us <usecs>\n"
        "                            Target latency above which throttling applies (default: 100000)\n"
+       "     --max-tokens N         Maximum number of tokens allowed in the token bucket (used by\n"
```
rkd-msw: Can you list the default in the help text?

rkd-msw:
Looks good, one picky comment. Probably worth announcing in the release note that we've enlarged the initial token rate, which will improve our behaviour when we get a large initial spike of load?

PR #945, opened 2015-02-19
[Reviewer: Rob] Add new option for memento load monitor

eleanor-merry:
Matching change for https://github.com/Metaswitch/memento/pull/48


rkd-msw:
Looks good. (As discussed, though, let's not document `cass_target_latency_us` as an option yet.)

Issue #853, opened 2014-11-07
Sprout node quiesce causes a signal 11 crash


tcaterisano:
The sprout upgrade procedure expects the user to quiesce the sprout node before stopping the node.
With a call active, I tried this using this command:  sudo /etc/init.d/sprout quiesce
This caused sprout to crash with a signal 11.

```
07-11-2014 21:17:57.069 UTC Debug handlers.cpp:590: Tombstone record indicates Authentication Vector has been used successfully - ignoring timer pop
07-11-2014 21:17:57.069 UTC Verbose httpstack.cpp:68: Sending response 200 to request for URL /authentication-timeout, args (null)
07-11-2014 21:17:58.623 UTC Status main.cpp:843: Quiesce signal received
07-11-2014 21:17:58.623 UTC Status stack.cpp:670: Destroyed TCP transport for port 5054
07-11-2014 21:17:58.623 UTC Error pjsip: Assert failed: ../src/pj/os_core_unix.c:674 !"Calling pjlib from unknown/external thread. You must " "register external threads with pj_thread_register() " "before calling any pjlib functions."

Signal 11 caught

Basic stack dump:
/usr/share/clearwater/bin/sprout(_ZN6Logger9backtraceEPKc+0x6d)[0x49bf0d]
/usr/share/clearwater/bin/sprout(_ZN3Log9backtraceEPKcz+0x10d)[0x5520ad]
/usr/share/clearwater/bin/sprout(_Z17exception_handleri+0x4f)[0x59dbdf]
/lib/x86_64-linux-gnu/libc.so.6(+0x36150)[0x7fcb68f9c150]
/lib/x86_64-linux-gnu/libc.so.6(+0x96f37)[0x7fcb68ffcf37]
/usr/share/clearwater/bin/sprout(pj_mutex_lock+0x93)[0x5ffea0]
/usr/share/clearwater/bin/sprout(pj_lock_acquire+0x62)[0x6075ad]
/usr/share/clearwater/bin/sprout(pjsip_tpmgr_unregister_tpfactory+0x20)[0x5d2c9a]
/usr/share/clearwater/bin/sprout[0x5d755e]
/usr/share/clearwater/bin/sprout(_ZN19StackQuiesceHandler18close_trusted_portEv+0x32)[0x4a7952]
/usr/share/clearwater/bin/sprout(_ZN16QuiescingManager13process_inputEi+0xd9)[0x56ca59]
[...]

PR #947, opened 2015-02-19
[Reviewer: Ellie] Exception handling: health-checker code

rkd-msw:
This integrates https://github.com/Metaswitch/cpp-common/pull/241 with Sprout. As we agreed, I've just made any INVITE/200 pass health-checking for simplicity; I haven't done anything to ignore I-CSCF 200 OKs if we have an S-CSCF. I've updated my copy of the spec to say so - I'll send that to you for a quick review once I've implemented the Homestead/Chronos/Ralf code, in case implementing those leads to any spec changes.

I've tested:
* pretty extensively in UT (see below)
* live, by checking for the `19-02-2015 17:08:06.773 UTC Debug health_checker.cpp:68: Check for overall system health passed - not aborting` log in Sprout

### DIFF
--- a/sprout/main.cpp
+++ b/sprout/main.cpp
@@ -1718,9 +1718,10 @@ int main(int argc, char* argv[])
                            stack_data.stats_aggregator);
   HealthChecker* health_checker = new HealthChecker();
   pthread_t health_check_thread;
-  pthread_attr_t health_check_attr;
-  pthread_attr_init(&health_check_attr);
-  pthread_create(&health_check_thread, &health_check_attr, &HealthChecker::static_main_thread_function, (void*)health_checker);
+  pthread_create(&health_check_thread,
+                 NULL,
+                 &HealthChecker::static_main_thread_function,
+                 (void*)health_checker);
 
   
   init_common_sip_processing(load_monitor,
--- a/sprout/main.cpp
+++ b/sprout/main.cpp
@@ -1716,11 +1716,17 @@ int main(int argc, char* argv[])
   Counter* overload_counter =
       new StatisticCounter("rejected_overload",
                            stack_data.stats_aggregator);
+  HealthChecker* health_checker = new HealthChecker();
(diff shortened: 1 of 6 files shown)


rkd-msw:
I've switched to a condition variable, so this is now ready for review. I've tested the new code:
* with the same UTs
* by running live and checking that a log is generated once a minute
* by terminating under valgrind and checking that nothing was leaked - but I couldn't confirm this, because I hit https://github.com/Metaswitch/sprout/issues/853 rather than exiting cleanly.

sprout/main.cpp
```diff
@@ -1716,11 +1716,17 @@ int main(int argc, char* argv[])
   Counter* overload_counter =
       new StatisticCounter("rejected_overload",
                            stack_data.stats_aggregator);
+  HealthChecker* health_checker = new HealthChecker();
+  pthread_t health_check_thread;
+  pthread_attr_t health_check_attr;
+  pthread_attr_init(&health_check_attr);
```
eleanor-merry: Does this attr get used at all? Also can you change the line length below

sprout/ut/common_sip_processing_test.cpp
```diff
@@ -366,3 +442,108 @@ TEST_F(CommonProcessingTest, BadResponseDropped)
   // Expect it to just vanish.
   ASSERT_EQ(0, txdata_count());
 }
+
+// If:
+//  - an exception has already been hit
+//  - the health-checker runs a check
+//  - and all the requests seen recently have not had a response
+// then Sprout should be aborted.
+TEST_F(CommonProcessingTest, DeathTest_MissingResponseFailsHealthCheck)
```
eleanor-merry: As discussed, can you tidy up the warnings here?

eleanor-merry:
Reviewed

rkd-msw:
@eleanor-merry , I've applied your markups.

PR #952, opened 2015-02-24
Reviewer MIRW: Mangelwurzel leaves the Req URI alone on in dialog requests

graemerobertson:
Hi Matt,

Please can you review my change to mangelwurzel so that it doesn't mangle the Request URI on in-dialog requests? As discussed, mangling the Request URI on in-dialog requests because we've mangled it on the initial request doesn't make sense. We may wish to mangle it for other reasons on in-dialog requests, but right now we don't have a well defined use case for this.

I've tested by running the UTs and deploying on a system.

I've updated the design.

Thanks,
Graeme

### DIFF
--- a/sprout/mangelwurzel/mangelwurzel.cpp
+++ b/sprout/mangelwurzel/mangelwurzel.cpp
@@ -237,7 +237,6 @@ void MangelwurzelTsx::on_rx_response(pjsip_msg* rsp, int fork_id)
 /// headers and send the request on. It can also change the request in various
 /// ways depending on the configuration in its Route header.
 /// - It can mangle the dialog identifiers using its mangalgorithm.
-/// - It can mangle the Request URI using its mangalgorithm.
 /// - It can mangle the Contact URI using its mangalgorithm.
 /// - It can mangle the To URI using its mangalgorithm.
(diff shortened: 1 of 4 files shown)


sprout/mangelwurzel/mangelwurzel.cpp
```diff
@@ -264,11 +263,6 @@ void MangelwurzelTsx::on_rx_in_dialog_request(pjsip_msg* req)
     mangle_dialog_identifiers(req, pool);
   }
 
-  if (_config.req_uri)
-  {
-    mangle_req_uri(req, pool);
-  }
-
   if (_config.contact)
```
mirw: Why do we mangle the Contact URI if we're not mangling the request URI?  Is the key point here that we should be enabling Contact URI mangling whenever we're mangling the request URI?

mirw:
One comment about whether this is necessary.  Apart from that, it's fine.

graemerobertson:
Nice spot. I've completely changed this now, as discussed. Please can you re-review? If you're happy with this, I'll update the docs to reflect the removal of the contact option.

I've tested live and run the UTs.

mirw:
As discussed, you should remove the `contact` field from the `Config` class... and please check for other instances.

Other than that, looks fine.

PR #955, opened 2015-02-25
[Reviewer: Ellie] Allow a comma-separated list of DNS servers

rkd-msw:
Can you review this change to allow signaling_dns_server and enum_server to be a list like `1.2.3.4,2.3.4.5`? I've tested live.

Note that:
* the UTs don't build - I need to fix up enumservice_test.cpp to use vectors of strings
* I should update the help text, but haven't got round to that
* I haven't resolved the (existing) duplication between dnsresolver.cpp and dnscachedresolver.cpp - that's slightly more work than I thought, because dnsresolver.cpp supports IPv6 but it looks like dnscachedresolver.cpp doesn't. I'll raise an issue to track.


include/cfgoptions.h
```diff
@@ -56,6 +56,8 @@ enum struct MemcachedWriteFormat
   BINARY, JSON
 };
 
+typedef std::vector<std::string> str_vector;
```
eleanor-merry: I don't know that I like this..

sprout/dnsresolver.cpp
```diff
@@ -59,33 +59,67 @@ DNSResolver::DNSResolver(const struct IP46Address& server) :
   // Set options to ensure we always get a response as quickly as possible -
   // we are on the call path!
   struct ares_options options;
-  options.flags = ARES_FLAG_PRIMARY | ARES_FLAG_STAYOPEN;
+  options.flags = ARES_FLAG_STAYOPEN;
```
eleanor-merry: Can you put a comment for what this option does?

sprout/dnsresolver.cpp
```diff
@@ -59,33 +59,67 @@ DNSResolver::DNSResolver(const struct IP46Address& server) :
...
   options.timeout = 1000;
-  options.tries = 1;
+  options.tries = 2;
```
eleanor-merry: Why 2 tries? Shouldn't this be the number of DNS servers?

sprout/main.cpp
```diff
@@ -249,7 +249,8 @@ static void usage(void)
        "                            If 'pcscf,icscf,as', it also Record-Routes between every AS.\n"
        " -G, --ralf <server>        Name/IP address of Ralf (Rf) billing server.\n"
        " -X, --xdms <server>        Name/IP address of XDM server\n"
-       "     --dns-server <server>  IP address of the DNS server to use (defaults to 127.0.0.1)\n"
+       "     --dns-server <server>[,<server2>,<server3>]\n"
```
eleanor-merry: You're planning to update the ENUM help text too, yes?

eleanor-merry:
Looks good

sprout/main.cpp
rkd-msw: Yes, I had definitely not forgotten that. <_<    >_>

rkd-msw:
All markups applied, and UTs fixed.

Issue #918, opened 2015-01-20
Sprout generates sas http events during user registrations but not visible when clicking View


tcaterisano:
SAS HTTP events don't seem to be correlated with the Registration. They show up as a separate item to View but are labeled as a SIP REGISTER event in sas.

When clicking on the View link, sas displays "Tab still loading..." forever.  Clicking in the Detailed Timeline does show the HTTP information.

```
2015-01-20 15:16:25.005sprout@sprout-2-plf.os2.richlab.datcon.co.ukReceived HTTP POST for /authentication-timeout
2015-01-20 15:16:25.005sprout@sprout-2-plf.os2.richlab.datcon.co.ukSending get request to memcached for key av\\2145190001@os2.richlab.datcon.co.uk\5cf6583b6a941ef6
2015-01-20 15:16:25.006sprout@sprout-2-plf.os2.richlab.datcon.co.ukMemcached get request succeeded for key av\\2145190001@os2.richlab.datcon.co.uk\5cf6583b6a941ef6
2015-01-20 15:16:25.006sprout@sprout-2-plf.os2.richlab.datcon.co.ukAuthentication vector store operation GET succeeded for private ID: "2145190001@os2.richlab.datcon.co.uk"
```

mirw:
@AlexHockey, can you take a look?

AlexHockey:
Registration is a two-phase process: 

*  The client sends a REGISTER which sprout challenges with a 401. 
*  The client then sends another REGISTER (with authentication credentials) which sprout checks and assuming all is well responds with a 200. 

Between these two REGISTERs sprout runs a timer (via chronos). If the second register never comes the timer pops and sprout takes some action to clear up state in homestead and the HSS. The problem here is that if the second register does happen sprout doesn’t unset the timer, but instead ignores it when it pops. This leads to the rogue trace you see in SAS.

Two solutions spring to mind:

1.  When the second REGISTER is processed, cancel the chronos timer. 
2.  Change sprout to work out if it will ignore the timer pop before it logs anything to SAS. A bit nasty though, and not really in the spirit of how SAS logging works. 

It feels like option 1 would be best.

AlexHockey:
Can't get too excited about this given it's just diags "spam". Making a medium priority.

mirw:
Is there a reason that this isn't correlated to one of the original REGISTER transactions?  If not, should it be?

AlexHockey:
[...]

PR #959, opened 2015-03-02
[Reviewer: Alex] Correlate timer pop events in SAS even when we don't act on them

rkd-msw:
Should fix https://github.com/Metaswitch/sprout/issues/918. I propose to test by:
* running the live tests against a deployment without this change, and checking SAS to confirm that https://github.com/Metaswitch/sprout/issues/918 repros
* running the live tests against a deployment with this change, and checking SAS to confirm that I can't see any timer-pop-only SAS events


AlexHockey:
The change to correlation looks fine. But as I mention in #918 I think we should add a new log explaining that we are ignoring the timer pop.

Test plan looks good.

Issue #951, opened 2015-02-23
Bad ENUM rules crash BGCF sproutlet


bossmc:
If the ENUM translation of a number returns a regex to apply, but the application fails (possibly due to bad ENUM configuration), then the BGCF sproutlet crashes:

```
23-02-2015 17:25:02.347 UTC Debug pjutils.cpp:2077: Global user +447417894444
23-02-2015 17:25:02.347 UTC Debug bgcfsproutlet.cpp:137: Performing ENUM lookup for user +447417894444
23-02-2015 17:25:02.347 UTC Debug enumservice.cpp:186: Translating URI via JSON ENUM lookup
23-02-2015 17:25:02.347 UTC Debug enumservice.cpp:233: Comparing first 11 numbers of +447417894444 against prefix +4474178919
23-02-2015 17:25:02.347 UTC Debug enumservice.cpp:233: Comparing first 13 numbers of +447417894444 against prefix +447417894444
23-02-2015 17:25:02.348 UTC Debug enumservice.cpp:240: Match found
23-02-2015 17:25:02.348 UTC Info enumservice.cpp:216: Number +447417894444 found, translated URI = +447417894444
23-02-2015 17:25:02.348 UTC Debug bgcfsproutlet.cpp:208: Invalid ENUM response: +447417894444
23-02-2015 17:25:02.348 UTC Debug sproutletproxy.cpp:1500: Adding message 0x2c1d6a0 => txdata 0x2c1d138 mapping
23-02-2015 17:25:02.348 UTC Verbose sproutletproxy.cpp:1218: bgcf-0x2bfc270 sending Response msg 404/INVITE/cseq=102 (tdta0x2c1d090)
23-02-2015 17:25:02.348 UTC Debug sproutletproxy.cpp:1507: Removing message 0x2c1a730 => txdata 0x2c1a1c8 mapping
23-02-2015 17:25:02.348 UTC Debug sproutletproxy.cpp:1295: Free message tdta0x2c1a120
23-02-2015 17:25:02.349 UTC Debug pjsip:  tdta0x2c1a120 Destroying txdata Request msg INVITE/cseq=102 (tdta0x2c1a120)

Signal 11 caught

Basic stack dump:
/usr/share/clearwater/bin/sprout(_ZN6Logger9backtraceEPKc+0x6d)[0x4a24ad]
/usr/share/clearwater/bin/sprout(_ZN3Log9backtraceEPKcz+0x10d)[0x5622dd]
[...]

PR #962, opened 2015-03-05
Reviewer AMC: Don't crash in BGCF when ENUM returns invalid rule

graemerobertson:
Hi Andy,

Can you review this? It fixes https://github.com/Metaswitch/sprout/issues/951. The fix is as described in the issue.

The new UT crashes without the fix, and doesn't crash with the fix.

I did a code read of all the other places we called free_msg and they look fine.

Grab me if you've got any other questions.

Thanks!
Graeme

### DIFF
--- a/sprout/ut/scscf_test.cpp
+++ b/sprout/ut/scscf_test.cpp
@@ -1762,7 +1762,10 @@ TEST_F(SCSCFTest, TestEnumNPBGCFTel)
   doSuccessfulFlow(msg, testing::MatchesRegex(".*+15108580401;rn.*+151085804;npdi@homedomain.*"), hdrs, false);
 }
 
-// Test where the BGCF does an ENUM lookup which returns an invalid rule.
+// Test where the BGCF does an ENUM lookup which returns an invalid rule. This
+// test uses two ENUM rules. The first one is invoked by the S-CSCF before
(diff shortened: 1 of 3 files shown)


sprout/ut/scscf_test.cpp
```diff
@@ -1762,6 +1762,22 @@ TEST_F(SCSCFTest, TestEnumNPBGCFTel)
...
+  SCOPED_TRACE("");
+  register_uri(_store, _hss_connection, "6505551239", "homedomain", "sip:wuntootreefower@10.114.61.213:5061;transport=tcp;ob");
+  _hss_connection->set_impu_result("sip:6505551000@homedomain", "call", HSSConnection::STATE_REGISTERED, "");
+  Message msg;
+  msg._toscheme = "tel";
+  msg._to = "16505551239";
+  msg._route = "Route: <sip:homedomain;orig>";
+  msg._todomain = "";
+  list<HeaderMatcher> hdrs;
+  doSlowFailureFlow(msg, 404, "", "ENUM failure");
```
bossmc: Could do with a comment explaining how we use the two ENUM rules to get to the line of code (and possibly transit-based route that's more likely in the field)?

bossmc:
One comment about comments otherwise fine.

PR #967, opened 2015-03-09
[Reviewer: Rob] Iss966

### LINKED ISSUE #966: Deregistration requests that have an invalid URI cause Sprout to crash
This is because registrationutils asserts that PJSIP was able to create a message using the invalid SIP URI.

eleanor-merry:
Rob, can you review this change to stop Sprout from crashing if a RTR has an invalid SIP URI. As part of this I've updated the UTs to actually run the task rather than just the individual methods

Tested in UT and live. 

Fixes #966


sprout/registration_utils.cpp
```diff
@@ -132,28 +132,38 @@ void RegistrationUtils::register_with_application_servers(Ifcs& ifcs,
...
+                     true, 
+                     is_initial_registration, 
+                     tdata->msg, 
+                     as_list, 
+                     trail);
+      pjsip_tx_data_dec_ref(tdata);
+    }
+    else
+    {
+      LOG_DEBUG("Unable to create third party registration");
```
rkd-msw: Would a SAS log be better here? Should we maybe also give more information (e.g. the AS URI)?

sprout/ut/handlers_test.cpp
```diff
@@ -182,36 +183,91 @@ class DeregistrationTaskTest : public SipTest
...
+  {
+    MockHttpStack::Request req(_httpstack,
+                               "/registrations?send-notifications=" + notify,
+                               "",
+                               "send-notifications=" + notify,
+                               body, 
+                               method);
+    DeregistrationTask::Config cfg(_regstore, _remotestore, _hss, NULL);
+    DeregistrationTask* task = new DeregistrationTask(req, &cfg, 0);
+    EXPECT_CALL(*_httpstack, send_reply(_, expected_rc, _));
```
rkd-msw: I thought (from @AlexHockey 's writeup a while ago) that expecting things in helper methods was a test smell. I'm not convinced either way - I think `invalid_dereg_request` is small enough that it might be clearer inline, but for `valid_dereg_request` that would be a lot of extra copy-and-pasting. @AlexHockey , thoughts?

AlexHockey: The problem with testing things in helper methods is that these methods easily morph into more complex sequences. They become much more complicated, gain more parameters, and become harder to re-use and maintain. 

My advice here is to separate the parts that build the request and expect the result. The tests that do an invalid request would look like:

```c++
build_request(params);
EXPECT_CALL(*_httpstack, send_reply(_, 400, _));
task->run();
```

The valid request testcases might ,might look like:

```c++
build_request(params);
expect_regstore_updates(other_params);
EXPECT_CALL(*_httpstack, send_reply(_, 200, _));
task->run();
```

(where in the second example I've split the expectations in two). I think that makes it easier to see why the request is (in)valid, avoids the temptation to create epic sequences, and is ultimately more composable.

rkd-msw:
Looks good - one minor comment on logs and one bigger one on UT structure.

(Should we search all the codebases for `assert`s and remove them?)

PR #969, opened 2015-03-09
[Reviewer: Rob] 14.04

eleanor-merry:
Sprout changes for 14.04


rkd-msw:
Looks good. No comments.

sprout/ut/sproutletproxy_test.cpp
```diff
@@ -1650,7 +1650,7 @@ TEST_F(SproutletProxyTest, UASError)
 
   ASSERT_EQ(NUM_FORKS, txdata_count());
 
-  for (int ii = 0; ii < NUM_FORKS; ++ii)
+  for (int ii = NUM_FORKS - 1; ii >= 0; --ii)
```
rkd-msw: Why the reversal?

rkd-msw:
I lied. One comment.

sprout/ut/sproutletproxy_test.cpp
rkd-msw: (We discussed, and it's because the ordering isn't guaranteed here and happens to have reversed between versions. I'm happy with this.)

PR #978, opened 2015-03-12
[Reviewer: Andy] New sas events

AlexHockey:
Andy, this PR contains the client-side changes needed for the new SAS events you have reviewed. 

I have tested this by setting a up a few different scenarios and checking the trace looks sensible. I haven't done exhaustive testing of each event, as this doesn't seem like a good use of time (especially given that most of the events have no parameters, so there is little chance of them not displaying correctly). 

The scenarios I have tested are as follows (all used a PC deployment with Ralf but no CDF). Let me know if you think anything more would be useful (though I may push back ;-)

* Registration.
* Successfull call with orig and term MMTEL AS
* Failed call (terminating subscriber not registered). 
* Failed call (homestead stopped). 

No docs needed (other than for me to write up the SAS logging guidelines, which I'll do tomorrow).


AlexHockey:
Though I should test what a trace looks like when we there is a CDF. I'll test this tomorrow.

bossmc:
Fine.

PR #982, opened 2015-03-17
[Reviewer: Ellie] Sprout depends on libevent-2.0-5

mirw:
Same as https://github.com/Metaswitch/homestead/pull/194, except libgnutls isn't required here.


eleanor-merry:
Looks good

mirw:
Actually, it's not clear that sprout should depend on libevent.  I'll repurpose the issue and close this.

PR #986, opened 2015-03-19
[Reviewer: Alex] If an AS-processed SIP request is on a different trail from the original...

### LINKED ISSUE #127: SAS logging of AS chains is probably wrong
There is code in sprout which changes the trail identifier on an incoming INVITE if it correlates the INVITE to an existing AS chain.  This is probably wrong as the INVITE has already been logged to SAS with the original trail identifier, and changing it runs the risk of the right markers not being raised on this original trail.

SAS already has the ability to correlate multiple trails on SIP call identifier, so it's not necessary for Sprout to do anything in this case.

rkd-msw:
... request, correlate the trails

Fixes #127. I've opted not to log ICID markers - this didn't add much obvious benefit, it wasn't going to be easy to test, and would add more load to SAS.

Before merging I need to:
* wait for https://github.com/Metaswitch/sas-client/pull/51 to be merged
* merge sas-client/master into sas-client/clearwater
* update the sas-client reference in this PR (and - sad times - everywhere else)

### DIFF
--- a/include/scscfsproutlet.h
+++ b/include/scscfsproutlet.h
@@ -208,9 +208,9 @@ public:
   virtual void on_timer_expiry(void* context);
 
 private:
-  // Examines the top route header to determine the relevant AS chain
-  // (from the ODI token) and the session case (based on the presence of
-  // the 'orig' param).
+  /// Examines the top route header to determine the relevant AS chain
+  /// (from the ODI token) and the session case (based on the presence of
+  /// the 'orig' param), and sets those as member variables.
   void retrieve_odi_and_sesscase(pjsip_msg* req);
   
   /// Determines the served user for the request.
--- a/include/scscfsproutlet.h
+++ b/include/scscfsproutlet.h
@@ -208,8 +208,12 @@ public:
   virtual void on_timer_expiry(void* context);
 
 private:
-  /// Determines the session case and the served user for the request,
-  /// and links to the appropriate AS Chain.
+  // Examines the top route header to determine the relevant AS chain
+  // (from the ODI token) and the session case (based on the presence of
+  // the 'orig' param).
+  void retrieve_odi_and_sesscase(pjsip_msg* req);
+  
+  /// Determines the served user for the request.
   pjsip_status_code determine_served_user(pjsip_msg* req);
 
   /// Gets the served user indicated in the message.
--- a/sprout/aschain.cpp
+++ b/sprout/aschain.cpp
@@ -196,9 +196,17 @@ AsChainLink AsChainLink::create_as_chain(AsChainTable* as_chain_table,
 //
 // @Returns whether processing should stop, continue, or skip to the end.
 void AsChainLink::on_initial_request(pjsip_msg* msg,
-                                     std::string& server_name)
+                                     std::string& server_name,
+                                     SAS::TrailId msg_trail)
 {
   server_name = "";
+
+  if (_as_chain->trail() != msg_trail)
+  {
+    // Associate the two trails in SAS so B2BUA calls are displayed properly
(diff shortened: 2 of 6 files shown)


sprout/aschain.cpp
```diff
@@ -196,10 +196,20 @@ AsChainLink AsChainLink::create_as_chain(AsChainTable* as_chain_table,
...
 void AsChainLink::on_initial_request(pjsip_msg* msg,
-                                     std::string& server_name)
+                                     std::string& server_name,
+                                     SAS::TrailId msg_trail)
 {
   server_name = "";
 
+  if (_as_chain->trail() != msg_trail)
+  {
+    // LCOV_EXCL_START - SAS UI association not suitable for UT
```
AlexHockey: Surely you can hit this code in UT.

sprout/ut/aschain_test.cpp
```diff
@@ -171,12 +171,12 @@ TEST_F(AsChainTest, AsInvocation)
...
 
   // Nothing to invoke. Just proceed.
-  as_chain_link.on_initial_request(tdata->msg, server_name);
+  as_chain_link.on_initial_request(tdata->msg, server_name, 0u);
   EXPECT_EQ(server_name, "");
 
   // Invoke external AS on originating side.
   LOG_DEBUG("ODI %s", as_chain_link2.to_string().c_str());
-  as_chain_link2.on_initial_request(tdata->msg, server_name);
+  as_chain_link2.on_initial_request(tdata->msg, server_name, 0u);
```
AlexHockey: Pass a different trail ID in here and you execute the code in the as chain link :-)

AlexHockey:
The code generally looks fine - one minor comment inline. 

On your decision to not add ICID correlation - how confident are you that VoLTE TASs will always preserve the ODI token? If they don't then #127 isn't fixed. If we added ICID correlation then we can say we have done everything we can to try and correlate - this isn't the case at the moment.

sprout/ut/aschain_test.cpp
rkd-msw: Yeah, but I wouldn't actually be testing anything, surely?

bossmc: Testing that it doesn't crash :smile:

AlexHockey: Just what I was going to say.

AlexHockey:
When implementing ICIDs we should be careful not to be too spammy. For example logging the ICID on each S-CSCF is unnecessary if we have co-located app servers, or if the ODI token correlates.

Better implementations would be:

*  Log the ICID in the S-CSCF but only if the ODI is missing or does not correlate. This is the most efficient we can be I think. 
*  Log the ICID in the stack module on initial requests.  This has the advantages that it deals with B2BUAs *anywhere* in the network (not just ASs) but at the cost of a bit more load on SAS.

rkd-msw:
Can you review my ICID logging changes in sprout/scscfsproutlet.cpp? They pass UT with full coverage (I didn't need any changes - both branches were covered already), but as agreed, I haven't tested live - Bono doesn't add P-C-V and mangelwurzel doesn't strip ODI tokens.

include/scscfsproutlet.h
```diff
@@ -208,8 +208,12 @@ class SCSCFSproutletTsx : public SproutletTsx
   virtual void on_timer_expiry(void* context);
 
 private:
-  /// Determines the session case and the served user for the request,
-  /// and links to the appropriate AS Chain.
+  // Examines the top route header to determine the relevant AS chain
+  // (from the ODI token) and the session case (based on the presence of
+  // the 'orig' param).
```
AlexHockey: You should be using triple slashes I think. It's also worth mentioning that the function stores the variables in member methods rather than returning them.

AlexHockey:
Looks good - one cosmetic comment inline.

PR #979, opened 2015-03-16
[Reviewer: Ellie] Make the RegStore distinguish between data contention and genuine errors. 

AlexHockey:
Ellie, please could you review my fix for #977.

This changes `RegStore::set_aor_data` to return a `Store::Status` rather than simply success/failure. I have also updated the users of the store to fail the request in progress if the store returns `Store::ERROR` (the exception is the registration timeout handler which continues to return 200 OK to chronos). 

Existing UTs all run fine, and I have added new UTs to cover failures of the RegStore. I will also check the live tests run OK, but I'm having trouble setting up a working deployment at the moment.

PR #988, opened 2015-03-20
[Reviewer: Ellie] Fix non-incrementing CSeqs

AlexHockey:
This fixes a regression in https://github.com/Metaswitch/sprout/pull/979 where CSeqs were not updated after processing a SUBSCRIBE. The fix is to move the CSeq increment to before we write the AoR back to the store. 

Tested with a new UT.


eleanor-merry:
Looks good

Issue #950, opened 2015-02-23
sprout uses Perl-Compatible regular expressions for ENUM


mirw:
[RFC 2915](https://www.ietf.org/rfc/rfc2915.txt) says POSIX Extended Regular Expressions should be used.

This should be easy to fix - there's an option to boost::regex to set the format.

mirw:
There's no need for us to maintain backward compatibility with previous versions - the two types of regular expressions only differ in pretty esoteric cases and a quick poll suggests no-one is using those.

PR #989, opened 2015-03-22
[Reviewer: Matt] POSIX regexes for ENUM

rkd-msw:
Fixes https://github.com/Metaswitch/sprout/issues/950. Exising UTs pass, and I've added a specific UT to confirm that Perl regexes aren't used.


mirw:
Looks good.

PR #990, opened 2015-03-26
[Reviewer: Rob] Make memento AS plugin cope with updated cassandra store

AlexHockey:
Hi Rob. This is another co-req change needed under sto1773 (I forgot that the Memento sproutlet is built as part of sprout and not memento). Note that I haven't removed any error checking here - start never returns an error when the store is configured with no threads (as the call list store is). 

Tested by checking the memento live tests pass.


rkd-msw:
Looks good.

PR #994, opened 2015-03-30
[Reviewer: Andy] Support Min-SE header

### LINKED ISSUE #529: Not using the Min-SE header can cause INVITEs to loop
The session-expires value is always set to the default-session-expires value, and doesn't take into account the Min-SE header that a client can send back requesting a longer time. This can mean that an INVITE can get stuck in a loop, where:
- Bono/Sprout sets the value to 600 and sends the INVITE on to a client.
- The client returns with a Min-SE value of (say) 700.
- Bono/Sprout resets the value to 600
- Repeat

We should support the Min-SE header defined in RFC 4028. 

We should also add a new config option max-session-expires, to protect against the case where a client asks for a really long session-expires time (which would mean that we couldn't rely on session refreshes). Then if there's a case where our max-session-expires is smaller than the received session-expires/min-se header, reject the call immediately.

eleanor-merry:
Andy, can you please review this change to support the Min-SE header. This fixes #529 (but doesn't fully support RFC 4028). 

Tested in UTs.

### DIFF
--- a/sprout/pjutils.cpp
+++ b/sprout/pjutils.cpp
@@ -53,6 +53,7 @@ extern "C" {
 #include "constants.h"
 #include "custom_headers.h"
 #include "sasevent.h"
+#include "sproutsasevent.h"
 
 static const int DEFAULT_RETRIES = 5;
 static const int DEFAULT_BLACKLIST_DURATION = 30;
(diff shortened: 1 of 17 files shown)


sprout/pjutils.cpp
```diff
@@ -2171,3 +2172,80 @@ bool PJUtils::does_uri_represent_number(pjsip_uri* uri,
...
+// Adds a Session-Expires header to the request to force the UEs to
+// exchange periodic session refresh messages. We use the value of the 
+// Min-SE header if set, or the default session expiry value.
+// 
+// Returns success if we could add a valid session expires header
+bool PJUtils::add_session_expires(pjsip_msg* req,
+                                  pj_pool_t* pool, 
+                                  SAS::TrailId trail)
+{
+  bool added_se = true;
```
bossmc: This isn't technically right, this function doesn't always add an SE, and sometimes it adds one and then returns `false`.  I think this is more "valid_se" (and the function comment needs updating too).

bossmc:
One dull comment, otherwise fine.

bossmc:
Also, you'll obviously need to update `cpp-common` with the new SAS version number.

PR #1000, opened 2015-04-14
[Reviewer: Andy] Iss888

### LINKED ISSUE #888: Sprout uses jsoncpp for parsing json - it should use rapidjson instead
Splitting this issue out from #559, as it's a lower priority to change these cases

eleanor-merry:
Andy, can you please review these changes to move to rapidjson (this finishes off the work started in #889)

Tested by checking the UTs/Live tests still pass, and that SAS logs of a reg/call look sensible.
Fixes #888


sprout/authentication.cpp
```diff
@@ -114,48 +113,54 @@ pjsip_auth_srv auth_srv;
...
-bool verify_auth_vector(Json::Value* av, const std::string& impi, SAS::TrailId trail)
+bool verify_auth_vector(rapidjson::Document* av, const std::string& impi, SAS::TrailId trail)
 {
   bool rc = true;
 
+  rapidjson::StringBuffer buffer;
+  rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+  av->Accept(writer);
+  std::string av_str = buffer.GetString();
+  LOG_DEBUG("Verifying AV: %s", av_str.c_str());
```
bossmc: This is only used for logging and serializing JSON is expensive.  Wrap in a log level check?

eleanor-merry: I use it later on in the SAS logs

sprout/authentication.cpp
```diff
@@ -215,12 +220,12 @@ pj_status_t user_lookup(pj_pool_t *pool,
...
     pj_cstr(&cred_info->scheme, "digest");
     pj_strdup(pool, &cred_info->username, acc_name);
-    if (av->isMember("aka"))
+    if (av->HasMember("aka"))
     {
       // AKA authentication.  The response in the AV must be used as a
       // plain-text password for the MD5 Digest computation.  Convert the text
       // into binary as this is what PJSIP is expecting.
-      std::string response = (*av)["aka"]["response"].asString();
+      std::string response = (*av)["aka"]["response"].GetString();
```
bossmc: You don't check for presence of `response` in the `aka` object...

sprout/authentication.cpp
```diff
@@ -235,13 +240,13 @@ pj_status_t user_lookup(pj_pool_t *pool,
       pj_strdup(pool, &cred_info->realm, realm);
       status = PJ_SUCCESS;
     }
-    else if (av->isMember("digest"))
+    else if (av->HasMember("digest"))
     {
-      if (pj_strcmp2(realm, (*av)["digest"]["realm"].asCString()) == 0)
+      if (pj_strcmp2(realm, (*av)["digest"]["realm"].GetString()) == 0)
```
bossmc: ditto (and probably thoughout).

sprout/bgcfservice.cpp
```diff
@@ -76,94 +70,96 @@ void BgcfService::update_routes()
...
+  }
+ 
+  // Now parse the document
+  rapidjson::Document doc;
+  doc.Parse<0>(bgcf_str.c_str());
 
-    file.close();
+  if (doc.HasParseError())
+  {
+    LOG_WARNING("Failed to read BGCF configuration data");
```
bossmc: Can we get the error message (http://miloyip.github.io/rapidjson/md_doc_dom.html#ParseError)?  Also, isn't this `LOG_ERROR` level?

sprout/enumservice.cpp
```diff
@@ -107,63 +102,76 @@ JSONEnumService::JSONEnumService(std::string configuration)
...
-    if (!reader.parse(file, root))
-    {
-      LOG_WARNING("Failed to read ENUM configuration data\n%s",
-                  reader.getFormattedErrorMessages().c_str());
-      return;
-    }
-    file.close();
+    // LCOV_EXCL_START
+    LOG_WARNING("Failed to read ENUM configuration data from %s",
+                configuration.c_str());
```
bossmc: `LOG_ERROR`?

sprout/enumservice.cpp
```diff
@@ -107,63 +102,76 @@ JSONEnumService::JSONEnumService(std::string configuration)
...
-    if (root["number_blocks"].isArray())
-    {
-      Json::Value number_blocks = root["number_blocks"];
+  // Now parse the document
+  rapidjson::Document doc;
+  doc.Parse<0>(enum_str.c_str());
+
+  if (doc.HasParseError())
+  {
+    LOG_WARNING("Failed to read ENUM configuration data");
```
bossmc: As above (both points).

sprout/enumservice.cpp
```diff
@@ -107,63 +102,76 @@ JSONEnumService::JSONEnumService(std::string configuration)
...
+    return;
+  }
+
+  try
+  {
+    JSON_ASSERT_CONTAINS(doc, "number_blocks");
+    JSON_ASSERT_ARRAY(doc["number_blocks"]);
```
bossmc: :open_mouth: This made it out of Chronos?  Why aren't we using this elsewhere?

sprout/handlers.cpp
```diff
@@ -321,37 +322,25 @@ RegStore::AoR* RegistrationTimeoutTask::set_aor_data(RegStore* current_store,
...
-  if ((json_body.isMember("binding_id")) &&
-      ((json_body)["binding_id"].isString()))
-  {
-    _binding_id = json_body.get("binding_id", "").asString();
-  }
-  else
+  catch (JsonFormatError err)
   {
-    LOG_WARNING("Binding ID not available in JSON");
+    LOG_INFO("Badly formed opaque data (missing aor_id or binding_id");
```
bossmc: ")"  <- There it is!

sprout/enumservice.cpp
eleanor-merry: This is actually the cpp-common version. Chronos uses its own set (and should be moved across to use cpp-common instead). 
I've used this where I thought it simplified the code (e.g. not in nested ifs)

sprout/icscfrouter.cpp
```diff
@@ -214,10 +214,11 @@ int ICSCFRouter::parse_hss_response(Json::Value& rsp, bool queried_caps)
...
 
-  if ((!rsp.isMember("result-code")) ||
-      ((rsp["result-code"].asString() != "2001") &&
-       (rsp["result-code"].asString() != "2002") &&
-       (rsp["result-code"].asString() != "2003")))
+  if ((!rsp->HasMember("result-code")) ||
+      (!(*rsp)["result-code"].IsInt()) ||
+      (((*rsp)["result-code"].GetInt() != 2001) &&
+       ((*rsp)["result-code"].GetInt() != 2002) &&
+       ((*rsp)["result-code"].GetInt() != 2003)))
```
bossmc: `GetInt` isn't free (it asserts on the type of the `Value` before extracting the actual value), pull it out first?

sprout/ut/hssconnection_test.cpp
```diff
@@ -469,26 +468,26 @@ TEST_F(HssConnectionTest, ServerFailure)
 
 TEST_F(HssConnectionTest, SimpleUserAuth)
 {
-  Json::Value* actual;
+  rapidjson::Document* actual;
   _hss.get_user_auth_status("privid69", "pubid44", "", "", actual, 0);
   ASSERT_TRUE(actual != NULL);
-  EXPECT_EQ("server-name", actual->get("scscf", "").asString());
+  EXPECT_TRUE(strcmp("server-name", (*actual)["scscf"].GetString()) == 0);
```
bossmc: I think `EXPECT_EQ((std::string("server-name"), (*actual)...);` will compile (and be more intuitive for future readers).

bossmc:
In general looks good, some all-permeating comments:

* For loading BGCF/S-CSCF/ENUM JSON files, I'd use `LOG_ERROR` to ensure users see the errors.
* When a parse fails, you can ask RapidJSON for the error string (in English or Chinese!) which seems like a good idea.
* There's various places where you check for the presence of a parent object but then dereference the children without checking them first.  Please self-review carefully for more of these (I think the only examples I saw were in the file I mentioned it in, but there may be more).

PR #1004, opened 2015-04-15
Reviewer AMC: Inter sproutlet request trace upstream

steelyvoid:
Adds logging at VERBOSE level of requests about to be sent downstream to a sproutlet, allowing inter-sproutlet requests to be traced whether or not they cross machine boundaries.

Tested by doing a live test of a BGCF call at VERBOSE level and by running the sprout UTs with NOISY=T:4 to generate inter-sproutlet logs for all UTs (checking each time that inter-sproutlet log output was sensible)


sprout/sproutletproxy.cpp
```diff
@@ -774,6 +775,38 @@ void SproutletProxy::UASTsx::schedule_requests()
...
+        //
+        // Serialise the message in a separate buffer using the function
+        // exposed by PJSIP.  In principle we could use req.req's own
+        // serialisation buffer structure for this, but then we'd need to
+        // explicitly invalidate it afterwards to avoid accidentally sending
+        // the wrong data over SIP at some future point.  Safer to use a local
+        // buffer.
+        if (Log::enabled(Log::VERBOSE_LEVEL))
+        {
+          char buf[MAX_SIP_MSG_SIZE];
```
bossmc: This is a 65Kb stack variable in a golden path function, as a premature (but confident) optimization can you pull this into its own function (so the massive stack is only created if VERBOSE logging is enabled?

sprout/sproutletproxy.cpp
```diff
@@ -774,6 +775,38 @@ void SproutletProxy::UASTsx::schedule_requests()
...
+          pj_ssize_t size;
+
+          // Ensure that buf contains a valid string before we attempt to
+          // serialise the message in it (so that we aren't left with an
+          // unterminated string if pjsip_msg_print fails).
+          buf[0] = 0;
+          size   = pjsip_msg_print(req.req->msg, buf, sizeof(buf));
+          size   = std::max(0L, size);
```
bossmc: If `size < 0` then the print failed and the next log won't be very useful, probably better to have a special log for this case ("Sending large (> 65535 byte) message to downstream sproutlet %s"?).  We shouldn't hit this, but if we do, a blank log won't help us :smile:

sprout/sproutletproxy.cpp
bossmc: I think there's `PJSIP_MAX_PKT_LEN` already (It's set to 65535 in https://github.com/Metaswitch/pjsip-upstream/blob/master/pjsip/include/pjsip/sip_config.h#L287)

sprout/sproutletproxy.cpp
```diff
@@ -774,6 +775,38 @@ void SproutletProxy::UASTsx::schedule_requests()
...
+
+          LOG_VERBOSE("Sending %d bytes %s to downstream sproutlet %s:\n"
+                      "--start msg--\n\n"
+                      "%.*s\n"
+                      "--end msg--",
+                      size,
+                      pjsip_tx_data_get_info(req.req),
```
bossmc: I'm not sure what this function returns... what does an example log look like?

bossmc:
Reviewed.

sprout/sproutletproxy.cpp
steelyvoid: Won’t the compiler notice the scope of this parameter and only allocate stack space if the branch is entered?

sprout/sproutletproxy.cpp
steelyvoid: Is it really worth adding an untestable branch (that might contain a bug either now or at some point in the future)?  Surely anyone seeing an empty log will look at the code and see that 0 => “there was some issue with generating the log – possibly size”.

sprout/sproutletproxy.cpp
steelyvoid: 15-04-2015 08:13:32.933 UTC Verbose sproutletproxy.cpp:807: Sending 1248 bytes Request msg SUBSCRIBE/cseq=1 (tdta0x7fda4c409640) to downstream sproutlet icscf:
--start msg--

SUBSCRIBE sip:8088260000@l3cc.metaswitch.test SIP/2.0^M
Record-Route: <sip:172.21.20.2:5054;transport=TCP;lr;service=scscf;billing-role=charge-orig>^M
Via: SIP/2.0/TCP 172.21.4.209:5080;received=172.21.4.209;branch=z9hG4bK+9fbdccd0a3f34469d4775f49780aec191+sip+6+a64e2212^M
From: <sip:8088260000@l3cc.metaswitch.test>;tag=172.21.4.209+6+ab4a14b8+410126a5^M
To: <sip:8088260000@l3cc.metaswitch.test>^M
CSeq: 1 SUBSCRIBE^M
…
--end msg--

bossmc: Ok, that's sensible.

sprout/sproutletproxy.cpp
bossmc: Quick prototyping suggests that `gcc` won't do this (apparently it could, but even with `-O3` and a huge buffer, I still see it allocate a massive stack frame).  It's not a bad idea to pull this out into it's own function anyway (in case we decide we want to use it elsewhere).

steelyvoid: OK

sprout/sproutletproxy.cpp
```diff
@@ -1771,3 +1787,32 @@ int SproutletWrapper::compare_sip_sc(int sc1, int sc2)
...
+  // Serialise the message in a separate buffer using the function
+  // exposed by PJSIP.  In principle we could use tdata's own
+  // serialisation buffer structure for this, but then we'd need to
+  // explicitly invalidate it afterwards to avoid accidentally sending
+  // the wrong data over SIP at some future point.  Safer to use a local
+  // buffer.
+  size = pjsip_msg_print(tdata->msg, buf, sizeof(buf));
+
+  // Defensively set size to zero if pjsip_msg_print failed
+  size   = std::max(0L, size);
```
bossmc: Rogue spaces.

bossmc:
Looks good.  One mark-up then it's good to merge.

PR #1005, opened 2015-04-16
Reviewer: Rob] Don't remove restund user when upgrading

AlexHockey:
A really boring PR to not delete the restund user when upgrading restund. This prevents an error message when upgrading. This follows the pattern we use for the bono/sprout users.

Tested by upgrading restund and checking that no error message was produced. Restund was running happily after the upgrade.


rkd-msw:
Looks good.

Issue #1007, opened 2015-04-17
Sprout encodes the Message-Body field incorrectly in JSON ACRs


mirw:
Sprout encodes the Message-Body field as a mixed array of strings and integers, when it should be encoded as an array of objects (containing strings).

PR #1009, opened 2015-04-17
[Reviewer: Ellie] Encode Message-Body correctly

mirw:
Ellie,

Please can you review this fix to encode Message-Body correctly?  It's encoded as an array of strings when it should be an array of objects containing strings.  Reviewing with `?w=1` might be clearer.

Live-tested only as there's no UT for ACRs.

Matt

### DIFF
--- a/sprout/acr.cpp
+++ b/sprout/acr.cpp
@@ -1067,19 +1067,23 @@ std::string RalfACR::get_message(pj_time_val timestamp)
            i != _msg_bodies.end();
            ++i)
       {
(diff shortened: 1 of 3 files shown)


eleanor-merry:
Looks good. There are UTs for ACRs, but acr.cpp isn't fully covered yet

mirw:
I've extended the UT.  Please can you confirm you're happy with this?

mirw:
(fixes #1007)

eleanor-merry:
Looks good

PR #1012, opened 2015-04-19
[Reviewer: Andy] Clean up and rename stateful_proxy.cpp

rkd-msw:
This pull request:
* removes all the "StatefulProxyTest" UTs except those explicitly testing P-CSCF or IBCF function
* removes all the code from stateful_proxy.cpp that was only hittable when edge_proxy was false
* removes callservices.cpp which was only invoked from that dead code
* renames all the stateful_proxy* files to bono* files

This is not all the cleanup I could potentially do (I haven't made Bono a Sproutlet, the UTs still have some fake Homer/Homestead connections, and init_stateful_proxy still has a lot of unused arguments), but it's a good start which I think can be merged as-is.

The main bit that isn't a straight deletion is https://github.com/Metaswitch/sprout/commit/266f0ca732bd0ad0947e42d0bfc5b47d0140b7d2#diff-f9464c1ac834aca5054b8b60eca6115dR1343 - I added some code to do SIP routing to the Request-URI even if we don't have Route headers and the Request-URI isn't the home domain. This is necessary to make the IBCF UTs pass once proxy_calculate_targets (which is otherwise mostly S-CSCF/memcached code) is deleted.

I'll live-test the P-CSCF function before merging, but don't think it's worth testing IBCF function now - though we should watch out for regressions when we cut the next release.


mirw:
I haven't reviewed the code, but can I just check that https://github.com/Metaswitch/sprout/commit/266f0ca732bd0ad0947e42d0bfc5b47d0140b7d2#diff-f9464c1ac834aca5054b8b60eca6115dR1343 isn't going to allow untrusted peers to route other than to the home domain.  (I appreciate it's necessary for IBCF, but it should not be allowed for clients using Bono as a P-CSCF.)

rkd-msw:
I think we should be fine - we'll either reject with 403 or add a Route header reflecting the Service-Route, and route based on that - but I'll test that case explicitly.

sprout/bono.cpp
```diff
@@ -421,170 +409,37 @@ void process_tsx_request(pjsip_rx_data* rdata)
...
+  status = proxy_process_routing(tdata);
 
-    if (status != PJ_SUCCESS)
-    {
-      LOG_ERROR("Error processing route, %s",
-                PJUtils::pj_status_to_string(status).c_str());
+  if (status != PJ_SUCCESS)
+  {
+    LOG_ERROR("Error processing route, %s",
+        PJUtils::pj_status_to_string(status).c_str());
```
bossmc: Weird indent.

sprout/bono.cpp
```diff
@@ -421,170 +409,37 @@ void process_tsx_request(pjsip_rx_data* rdata)
...
-      // or it did not indicate that willing was required).  Create a dummy
-      // ACR.
-      if (acr == NULL)
-      {
-        acr = new ACR();
-      }
-    }
+  acr = cscf_acr_factory->get_acr(get_trail(rdata),
+      CALLING_PARTY,
+      acr_node_role(rdata->msg_info.msg));
```
bossmc: Indent.

sprout/bono.cpp
```diff
@@ -4686,7 +3143,7 @@ void UACTransaction::exit_context()
 
 pj_status_t init_stateful_proxy(RegStore* registrar_store,
                                 RegStore* remote_reg_store,
-                                CallServices* call_services,
+                                void* call_services,
```
bossmc: Unused?

sprout/ut/coverage-not-yet
```diff
@@ -37,7 +37,7 @@ modules/rapidjson/include/rapidjson/allocators.h
 modules/rapidjson/include/rapidjson/internal/stack.h
 modules/rapidjson/include/rapidjson/writer.h
 
-stateful_proxy.cpp
+bono.cpp
```
bossmc: Is this still true?  Have you improved coverage with this change?

bossmc:
Fine.

PR #1013, opened 2015-04-19
[Reviewer: Ellie] ICSCF should send 480 when HSS sends 5003

rkd-msw:
Fixes https://github.com/Metaswitch/homestead/issues/210.


eleanor-merry:
Looks good

Issue #964, opened 2015-03-06
is_user_numeric doesn't correctly identify phone numbers


graemerobertson:
We use is_user_numeric to determine whether we actually have a telephone number despite not having a tel URI. Currently is_user_numeric determines a phone number simply by checking that the first character is either + or 0-9 and everything else is 0-9. tel URIs are actually a lot more complicated than that - see http://www.in2eps.com/fo-abnf/tk-fo-abnf-sipuri.html#teluri. We should conform to that, especially in the I-CSCF sproutlet when we use the function to determine whether to call translate_sip_uri_to_tel_uri.

PR #1014, opened 2015-04-20
[Reviewer: Ellie] Ignore separators in is_user_numeric

rkd-msw:
Fixes https://github.com/Metaswitch/sprout/issues/964. I think this accepts all valid instances of `global_number_digits` from http://www.in2eps.com/fo-abnf/tk-fo-abnf-sipuri.html#teluri - I haven't worried about the additional complexity of local_number_digits, because we use this function to decide:
* whether to do an ENUM lookup
* whether to convert a SIP URI to a TEL URI in the I-CSCF

and both operations are only valid for global E.164 numbers (even if we have `enforce_user_phone=N`, it just doesn't make sense to do a ENUM or LIR lookup for `tel:*1234A#`).

### DIFF
--- a/sprout/bgcfservice.cpp
+++ b/sprout/bgcfservice.cpp
@@ -45,8 +45,7 @@
 #include "log.h"
 #include "sas.h"
 #include "sproutsasevent.h"
-
-const boost::regex BgcfService::CHARS_TO_STRIP = boost::regex("[.)(-]");
+#include "pjutils.h"
 
 BgcfService::BgcfService(std::string configuration) :
   _configuration(configuration),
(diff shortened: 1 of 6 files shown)


sprout/sprout_test.mk
```diff
@@ -146,7 +146,8 @@ TARGET_SOURCES_TEST := test_main.cpp \
                        mangelwurzel_test.cpp \
                        alarm_test.cpp \
                        communicationmonitor_test.cpp \
-                       common_sip_processing_test.cpp
+                       common_sip_processing_test.cpp \
+		       pjutils_test.cpp
```
bossmc: Boo!

sprout/bgcfservice.cpp
```diff
@@ -242,7 +241,7 @@ std::vector<std::string> BgcfService::get_route_from_number(
   {
     int len = std::min(number.size(), (*it).first.size());
 
-    if (remove_visual_separators(number).compare(0, 
+    if (PJUtils::remove_visual_separators(number).compare(0,
```
eleanor-merry: Alignment

eleanor-merry:
Looks good

PR #1017, opened 2015-04-24
[Reviewer: Alex] UTs for new rate-limited load monitor function

rkd-msw:
UTs for https://github.com/Metaswitch/cpp-common/pull/281. I've basically reworked LoadMonitorTest.RequestComplete to check the allowed request rate, rather than the smoothed latency.


AlexHockey:
Looks good. 

Arguably getting the refill rate directly out of a load monitor member variable is breaking encapsulation a bit, but the alternative (calling `admit_request` at a certain rate) seems like overkill for such a small, standalone test suite.

PR #1020, opened 2015-04-29
[Reviewer: Rob] Tell sipp to use a dumb terminal

mirw:
Rob,

Please can you review this fix to sprout to use a dumb terminal with clearwater-sip-stress.  On Juju, the terminal isn't always set correctly, which can break this function.

Matt


rkd-msw:
Looks good.

PR #1023, opened 2015-04-30
[Reviewer: Andy] Make DNS blacklist duration configurable

mirw:
Andy,

Please can you review this fix to make the DNS blacklist duration configurable? There's no UT for this, but I've tested it live (and it's actually just refactoring).

This is the fix for sprout.  Once you've reviewed, I'll apply the same fix to homestead, ralf, memento, etc.

Matt


bossmc:
Fine.

PR #1025, opened 2015-05-01
[Reviewer: Rob] Add in Chronos plugin to Sprout

eleanor-merry:
Rob, can you review this change to add a chronos plugin to sprout

### DIFF
--- a/sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_chronos_plugin.py
+++ b/sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_chronos_plugin.py
@@ -1,0 +1,83 @@
+# @file sprout_chronos_plugin.py
+#
+# Project Clearwater - IMS in the Cloud
+# Copyright (C) 2015  Metaswitch Networks Ltd
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or (at your
+# option) any later version, along with the "Special Exception" for use of
+# the program along with SSL, set forth below. This program is distributed
+# in the hope that it will be useful, but WITHOUT ANY WARRANTY;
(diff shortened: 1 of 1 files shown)


sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_chronos_plugin.py
```diff
@@ -0,0 +1,83 @@
...
+        _log.debug("Reloading Chronos")
+        os.system("service chronos reload")
+
+    def on_joining_cluster(self, cluster_view):
+        _log.debug("Sprout's Chronos cluster is changing")
+        self.on_cluster_changing(cluster_view)
+
+    def on_new_cluster_config_ready(self, cluster_view):
+        _log.debug("Started running Chronos resynchronization")
+        os.system("service chronos resync")
```
rkd-msw: We should probably use https://docs.python.org/2/library/subprocess.html instead of `os.system`, but I think the benefits are fairly minimal unless we wrap it in more stuff to handle error logging.

sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_chronos_plugin.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def on_cluster_changing(self, cluster_view):
+        write_chronos_cluster_settings("/etc/chronos/chronos_cluster.conf",
+                               cluster_view,
+                               self.local_server)
+        _log.debug("Reloading Chronos")
+        os.system("service chronos reload")
+
+    def on_joining_cluster(self, cluster_view):
+        _log.debug("Sprout's Chronos cluster is changing")
```
rkd-msw: Should this log be in `on_cluster_changing`, and a different log be in `on_joining_cluster`?

rkd-msw:
Looks good.

PR #1026, opened 2015-05-02
[Reviewer: Andy] Retry AV tombstones writes that fail due to CAS

### LINKED ISSUE #1010: Deregistration leaves subscriber registered
While performing some non-clearwater tests, I've run into a flaky situation where I register a subscriber and subsequently de-register that subscriber, yet it remains registered in both homestead and the HSS. I say "flaky" because it's not consistently reproducible and it seems that the longer the clearwater system is up, the more likely this problem is to occur.

Diags have been captured.

mirw:
Andy,

Please can you review this fix to make Sprout retry AV tombstones writes that fail due to CAS.  We don't normally expect them to fail due to genuine CAS, but they can fail if we read from a backup and then try to write back to a (previously-unavailable) primary.  This fixes #1010.

There's no UT infrastructure for failing memcached writes, but I have tested live.  I ran up a system with 2 Sprouts and ran light (10kBHCA) stress through it.  Without the patch, shortly after I killed Memcached, I saw logs of the following form:

```
02-05-2015 00:58:22.963 UTC Error avstore.cpp:78: Failed to write Authentication Vector for private_id 2010000421@mirw.cw-ngv.com
02-05-2015 00:58:22.963 UTC Error authentication.cpp:619: Tried to tombstone AV for 2010000421@mirw.cw-ngv.com/0f3ea1dc23b199f5 after processing an authentication, but failed
```

With the patch, I just saw the first log, not the second (as it was retried and succeeded).

Matt

### DIFF
--- a/sprout/authentication.cpp
+++ b/sprout/authentication.cpp
@@ -443,8 +443,8 @@ void create_challenge(pjsip_authorization_hdr* auth_hdr,
     // Write the authentication vector (as a JSON string) into the AV store.
     LOG_DEBUG("Write AV to store");
     uint64_t cas = 0;
-    bool success = av_store->set_av(impi, nonce, av, cas, get_trail(rdata));
-    if (success)
+    Store::Status status = av_store->set_av(impi, nonce, av, cas, get_trail(rdata));
+    if (status == Store::OK)
     {
       // We've written the AV into the store, so need to set a Chronos
       // timer so that an AUTHENTICATION_TIMEOUT SAR is sent to the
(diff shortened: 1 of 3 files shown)


sprout/authentication.cpp
```diff
@@ -605,13 +605,38 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
...
+        store_status = av_store->set_av(impi, nonce, av, cas, trail);
+        if (store_status == Store::DATA_CONTENTION)
+        {
+          // LCOV_EXCL_START - No support for contention in UT
+          LOG_DEBUG("Data contention writing tombstone - retry");
+          delete av;
+          av = av_store->get_av(impi, nonce, cas, trail);
+          if (av == NULL)
+          {
+            store_status = Store::ERROR;
```
bossmc: Surely this isn't an error?  Either we still want to write in the tombstone, on which case we should loop back round, or we don't care in which case this is a success state?

bossmc:
One comment inline which I'll leave for when you get back from holiday to thrash out.

sprout/authentication.cpp
mirw: `get_av` returns NULL if

*   there is no AV in the store
*   the store contains a corrupt AV
*   a communication error occurs, e.g. memcached is not responding or it returns a memcached error.

The last two situations are genuine errors - we definitely shouldn't retry.  The first is arguably a success condition but shouldn't really happen - we successfully read this a fraction of a second earlier and we never delete AVs, so the only way that one should disappear is through expiry (where we should always pop a Chronos timer first).

I think it's arguable that we should add a parameter to `get_av` to return the failure reason but this feels like a lot of plumbing/API pollution for not much real benefit.  Thoughts?

bossmc: Yeah, that sounds reasonable, I'm happy with this change.

bossmc:
Happy to merge.

PR #1027, opened 2015-05-05
[Reviewer: Andy] Use new SAS event

eleanor-merry:
Using new SAS event from https://github.com/Metaswitch/clearwater-sas-resources/pull/67


bossmc:
Cool

PR #1030, opened 2015-05-06
[Reviewer: Ellie] Memcached clustering plugin

rkd-msw:
Can you review `sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_memcached_plugin.py`?

Note that:
* I'll rename `write_cluster_settings` to `write_memcached_cluster_settings`
* I've gone for fairly minimal logs, because the logs generated before calling into the plugin and the logs generated by run_command should be enough to tell what's going on

### DIFF
--- a/debian/sprout-base.postinst
+++ b/debian/sprout-base.postinst
@@ -94,8 +94,7 @@ case "$1" in
         add_section /etc/security/limits.conf sprout /etc/security/limits.conf.sprout
         /usr/share/clearwater/infrastructure/scripts/sprout.monit
 
-        # Stop the cluster manager, so that it is restarted by Monit and picks up the new memcached
-        # and Chronos scaling plugins.
+        # Stop the cluster manager, so that it is restarted by Monit and picks up the new plugins.
         service clearwater-cluster-manager stop || /bin/true
 
         # Restart sprout.  Always do this by terminating sprout so monit will
--- a/debian/sprout-base.postinst
+++ b/debian/sprout-base.postinst
@@ -94,6 +94,10 @@ case "$1" in
         add_section /etc/security/limits.conf sprout /etc/security/limits.conf.sprout
         /usr/share/clearwater/infrastructure/scripts/sprout.monit
 
+        # Stop the cluster manager, so that it is restarted by Monit and picks up the new memcached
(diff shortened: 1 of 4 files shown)


sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_memcached_plugin.py
```diff
@@ -0,0 +1,74 @@
...
+from metaswitch.clearwater.cluster_manager.plugin_base import \
+    SynchroniserPluginBase
+from metaswitch.clearwater.cluster_manager.plugin_utils import \
+    run_command, write_cluster_settings
+from metaswitch.clearwater.cluster_manager.alarms import issue_alarm
+from metaswitch.clearwater.cluster_manager import constants
+import logging
+import os
+
+_log = logging.getLogger("memcached_plugin")
```
eleanor-merry: Worth including sprout in the name? This would be useful for a colocated sprout/memento

debian/sprout-base.postinst
```diff
@@ -94,6 +94,10 @@ case "$1" in
         add_section /etc/security/limits.conf sprout /etc/security/limits.conf.sprout
         /usr/share/clearwater/infrastructure/scripts/sprout.monit
 
+        # Stop the cluster manager, so that it is restarted by Monit and picks up the new memcached
```
eleanor-merry: I wouldn't include the list of plugins here as it will just go out of date

eleanor-merry:
Looks good

sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_memcached_plugin.py
rkd-msw: I'd forgotten about colocated Sprout/Memento - that doesn't work very well with these plugins. As discussed, I'm going to:
* make memento use /etc/clearwater/memento_cluster_settings to avoid conflicts
* as a more general solution, make the plugins define what config files they want to own, and only load the first plugin that (say) wants to modify /etc/cassandra/cassandra.yaml.

PR #1032, opened 2015-05-12
Reviewer AMC: Pull in pjmedia, requires rename of mod_auth

graemerobertson:
We've removed all references to pjmedia from sprout_base, but some app servers require it. So we need to explicitly link it. Unfortunately, pulling this in also pulls in a reference to a symbol called mod_auth, so we rename our mod_auth. mmtel and call_diversion_as now load agian.

### DIFF
--- a/include/authentication.h
+++ b/include/authentication.h
@@ -44,7 +44,7 @@
 #include "acr.h"
 #include "analyticslogger.h"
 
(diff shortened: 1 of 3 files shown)


bossmc:
Aces, Charles, aces.

Issue #871, opened 2014-11-19
Can’t add the /etc/clearwater/remote_cluster_settings file without restarting sprout


dtheobald:
You can’t add the /etc/clearwater/remote_cluster_settings file without restarting sprout (i.e. reload won’t work because there was no /etc/clearwater/remote_cluster_settings file when I first started sprout. Granted the use case for this is quite rare, that being that this would only be a problem when converting a non-GR system to a GR system. However, for completeness, perhaps an empty ‘remote_cluster_settings’ file should be installed by default.

rkd-msw:
Fixed by the above PR.

PR #1035, opened 2015-05-13
[Reviewer: Ellie] Create an empty /etc/clearwater/remote_cluster_settings if we have a …

rkd-msw:
…remote site

Pretty standard - this just gets a blank remote_cluster_settings in place, ready to be updated by etcd without requiring a software restart.

I've tested by:
* upgrading Sprout to this version
* checking that an empty remote_cluster_settings was created
* checking that Sprout started up, used this file and didn't crash
* Registering, making a call and unregistering, checking we didn't crash or have any odd logs
* Adding a server to remote_cluster_settings, reloading, and checking that it was used

I had debug logging turned on throughout, in case any debug logs caused a crash (one did).

I initially made the remote_cluster_settings stuff conditional on remote_site_name, but then removed this so it's unconditional - this fixes https://github.com/Metaswitch/sprout/issues/871, and means that non-etcd GR deployments still work, without the need to add remote_site_name. (In theory it would allow you to add a remote site without impacting service, although I don't think we need to test/prove this.)

### DIFF
--- a/sprout/main.cpp
+++ b/sprout/main.cpp
@@ -1658,9 +1658,8 @@ int main(int argc, char* argv[])
 
         if (!(((MemcachedStore*)remote_data_store)->has_servers()))
         {
-          LOG_ERROR("Remote cluster settings file '%s' does not contain a valid set of servers",
-                    opt.remote_store_servers.c_str());
-          return 1;
+          LOG_WARNING("Remote cluster settings file '%s' does not contain a valid set of servers",
+                      opt.remote_store_servers.c_str());
         };
       }
     }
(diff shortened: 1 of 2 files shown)


rkd-msw:
Just to confirm, by 'empty' I actually mean 'contains `servers=` and nothing else'.

sprout/main.cpp
```diff
@@ -1658,9 +1658,8 @@ int main(int argc, char* argv[])
 
         if (!(((MemcachedStore*)remote_data_store)->has_servers()))
         {
-          LOG_ERROR("Remote cluster settings file '%s' does not contain a valid set of servers",
-                    opt.remote_store_servers.c_str());
-          return 1;
+          LOG_WARNING("Remote cluster settings file '%s' does not contain a valid set of servers",
```
eleanor-merry: Why have you reduced the severity here?

eleanor-merry:
Looks good, one question

sprout/main.cpp
rkd-msw: This isn't an error any more - this is a mainline case (when you start up with just 'servers=', before you get reloaded).

PR #1036, opened 2015-05-15
Reviewer MIRW: Remove unnecessary group key locks from SAS logging routine to avoid …

steelyvoid:
…IO/group lock deadly embrace.

This change takes out the group key locks taken by sas_log_rx_msg when querying the transaction object for the trail ID corresponding to a message.  As the calling code is running in the context of the transport thread (and already holds the IO queue lock), acquisition of a group lock here breaks the hierarchy for these locks (which requires that the group lock be taken first) and can lead to deadly embraces with worker threads acquiring the locks in the correct order.

Note that the transaction can be safely queried from this routine without the lock, as

-   there is only one transport thread, and it is this thread (only) which is responsible for the destruction of transaction objects.  Therefore the transaction object itself cannot be invalidated under the feet of sas_log_rx_msg
-   while in principle the trailID in the transaction might be changed by another thread, thereby corrupting the value read in sas_log_rx_msg, the chances of this happening are extremely remote and the impact of such a corruption minor (SAS will accept the record as normal - it just won't appear in SAS traces for the call)

### DIFF
--- a/sprout/common_sip_processing.cpp
+++ b/sprout/common_sip_processing.cpp
@@ -141,6 +141,28 @@ static void sas_log_rx_msg(pjsip_rx_data* rdata)
 {
   SAS::TrailId trail = 0;
 
+  // Look for the SAS Trail ID for the corresponding transaction object.
+  //
+  // Note that we are NOT locking the transaction object before we fetch the
+  // trail ID from it.  This is deliberate - we cannot get a group lock from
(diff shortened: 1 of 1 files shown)


bossmc:
@mirw - As you have the background on this, can I ask you to review please?

mirw:
Fine - merged.

PR #1038, opened 2015-05-17
Reviewer Alex - Config plugins

bossmc:
Three plugins installed by Sprout to manage shared files.  Tested live.


sprout-base.root/usr/share/clearwater/clearwater-config-manager/plugins/sprout_scscf_json_plugin.py
```diff
@@ -0,0 +1,57 @@
...
+    def on_config_changed(self, value):
+        _log.info("Updating S-CSCF selection information")
+
+        with open("/etc/clearwater/s-cscf.json", "w") as ofile:
+            ofile.write(value);
+
+        run_command("service sprout reload");
+
+def load_as_plugin(ip):
+    return SproutSCSCFJsonPlugin()
```
AlexHockey: Would it be better to have a single `SproutJsonPlugin` class and create one instance of it per plugin? That would cut down on code copying.

sprout-base.root/usr/share/clearwater/clearwater-config-manager/tools/upload_scscf_json
```diff
@@ -0,0 +1,59 @@
...
+  cat > $FILENAME << EOF
+{ "s-cscfs": [] }
+EOF
+fi
+
+# Upload the file to etcd
+curl -X PUT \
+     http://$local_ip:4000/v2/keys/configuration/scscf_json \
+     --data-urlencode \
+     value@$FILENAME > /dev/null
```
AlexHockey: Same comment here: wouldn't it be better to have a single script that is called in 3 different ways?

AlexHockey:
Looks good apart from all the code copying ;-)

bossmc:
Markups applied

PR #1040, opened 2015-05-18
[Reviewer: Ellie] Don't load the remote memcached plugin if there's no remote site

rkd-msw:


rkd-msw:
@mirw, you're likely to want this fix.

rkd-msw:
(It's a plugin-only fix so 'make deb-only' will pick it up.)

eleanor-merry:
Looks good

Issue #1029, opened 2015-05-05
IPv6 is generally broken for the S-CSCF sproutlet


eleanor-merry:
If the sprout_hostname is an IPv6 address then the S-CSCF plugin fails to load as the BGCF URI is invalid 

Debug scscfsproutlet.cpp:99:   BGCF URI           = sip:bgcf.[::1]:5054;transport=TCP
Error scscfsproutlet.cpp:125: Invalid BGCF URI sip:bgcf.[::1]:5054;transport=TCP

eleanor-merry:
Similar to #929

eleanor-merry:
A workaround to the above is to make the sprout_hostname a hostname. This then crashes on the S-CSCF node URI:

Debug scscfsproutlet.cpp:95: Creating S-CSCF Sproutlet
Debug scscfsproutlet.cpp:96:   S-CSCF cluster URI = sip:sprout.em4.cw-ngv.com:5054;transport=TCP
Debug scscfsproutlet.cpp:97:   S-CSCF node URI    = sip:::1:5054
Debug scscfsproutlet.cpp:98:   I-CSCF URI         = sip:sprout.em4.cw-ngv.com:5052;transport=TCP
Debug scscfsproutlet.cpp:99:   BGCF URI           = sip:bgcf.sprout.em4.cw-ngv.com:5054;transport=TCP
Error scscfsproutlet.cpp:117: Invalid S-CSCF node URI sip:::1:5054

This is built from the local_ip value in /etc/clearwater/config. We can't just wrap this in square brackets, as pj_getaddrinfo fails to parse this, e.g.:

Error stack.cpp:220: Failed to decode IP address [::1] (gethostbyname() has returned error (PJ_ERESOLVE))

Raising the priority as there's not a simple workaround

bossmc:
Raising further to improve IPv6 support.

rkd-msw:
Reopening - https://github.com/Metaswitch/sprout/commit/29f41eb451f86c12078cfb5e78ae66db87bef86a#diff-a3f32660e29268991e5b2bc959f00cfb partly undoes #1173, so you now have URIs like `sip:bgcf.[::1]` again.

rkd-msw:
(This is particularly noticeable on an all-in-one node.)

rkd-msw:
[...]

PR #1041, opened 2015-05-18
[Reviewer: Rob] Add in bracket call

### LINKED ISSUE #1028: scscf_uri should be created using bracket_ipv6_address.py
The scscf_uri is created in the debian file at https://github.com/Metaswitch/sprout/blob/dev/sprout-base.root/etc/init.d/sprout#L257. This assumes that the sprout_hostname is IPv4, and just appends the scscf_port. 
This should use the bracket_ipv6_address script instead (like at https://github.com/Metaswitch/sprout/blob/dev/sprout-base.root/etc/init.d/sprout#L110)

eleanor-merry:
Calls bracket_ipv6_address when creating the scscf_uri. 
Fixes #1028  (but not the wider issue covered by #1029 )

### DIFF
--- a/sprout-base.root/etc/init.d/sprout
+++ b/sprout-base.root/etc/init.d/sprout
@@ -260,9 +260,12 @@ get_daemon_args()
           then
             DAEMON_ARGS="$DAEMON_ARGS --scscf-uri=$scscf_uri"
           else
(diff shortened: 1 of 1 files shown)


rkd-msw:
Looks good.

PR #1042, opened 2015-05-19
[Reviewer: Ellie] Use public addresses as aliases

rkd-msw:
Tested live, with a version of poll-sip that uses $public_hostname in the request URI. OPTIONS polls without this change fail with 483 Too Many Hops, and succeed with this change.

(I've verified that having a trailing `,` when alias_list is undefined is harmless.)


rkd-msw:
Fixes #1043.

eleanor-merry:
Looks good. Can you also verify the live tests pass?

rkd-msw:
Tests passed.

PR #1045, opened 2015-05-20
[Reviewer: Ellie] Honour upstream_port=0, which should use SRV records for upstream_hos…

### LINKED ISSUE #1021: Bono doesn't use RFC 3263 DNS resolution to locate Sprout
Since Bono always constructs a Sprout SIP URI that includes a port number, it always just uses A records to resolve Sprout's IP address - it never uses NAPTR or SRV.  It should, as these give much better control over load-balancing.

rkd-msw:
…tname

This lets Bono use SRV if the upstream_port setting is explicitly set to 0. (It has to be explicitly set to 0 - init.d defaults it to 5054 otherwise, which is consistent with us generally not requiring SRV records by default.

`upstream_port=0` isn't very intuitive for SRV, but I think it's good enough. Fixes #1021.

### DIFF
--- a/sprout/main.cpp
+++ b/sprout/main.cpp
@@ -893,14 +893,6 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
     }
   }
 
-  // If the upstream proxy port is not set, default it to the trusted port.
-  // We couldn't do this earlier because the trusted port might be set after
-  // the upstream proxy.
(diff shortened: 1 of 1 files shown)


rkd-msw:
I've tested live.

eleanor-merry:
Looks good. Can you update the docs for this as well?

rkd-msw:
Done in the big GR docs pull request.

PR #1046, opened 2015-05-20
[Reviewer: Ellie] Remote store log spam

rkd-msw:
This stops us using the remote store if no servers are configured, preventing some log spam you saw.

### DIFF
--- a/include/regstore.h
+++ b/include/regstore.h
@@ -304,6 +304,8 @@ public:
     std::string serialize_aor(AoR* aor_data);
     AoR* deserialize_aor(const std::string& aor_id, const std::string& s);
 
+    bool underlying_store_has_servers() { return _data_store->has_servers(); }
+
     Store* _data_store;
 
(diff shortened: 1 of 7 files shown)


include/regstore.h
```diff
@@ -304,6 +304,8 @@ class RegStore
     std::string serialize_aor(AoR* aor_data);
     AoR* deserialize_aor(const std::string& aor_id, const std::string& s);
 
+    bool underlying_store_has_servers() { return (_data_store != NULL) && _data_store->has_servers(); }
```
eleanor-merry: Line length

eleanor-merry: Also, can the data_store be NULL here?

eleanor-merry:
Looks good

include/regstore.h
rkd-msw: Apparently yes, in UT - it crashed on me.

PR #1049, opened 2015-05-21
[Reviewer: Andy] Iss1048

eleanor-merry:
Changes to use scripts instead of tools in the upload scripts. This also pulls in the latest dev.


bossmc:
The changes to the upload scripts look fine, I've not reviewed anything else.

PR #1051, opened 2015-05-22
[Reviewer: Ellie] Disable plugin loading for I-CSCF

rkd-msw:
As discussed, only Sprout nodes which are S-CSCFs should join the Memcached or Chronos clusters. This implements that, by making the plugins run a Bash command that checks if $scscf has been explicitly set to 0.

This is pretty unpleasant code, and our S-CSCF/I-CSCF config model is hacky in other ways (e.g. it breaks shared config) - can you raise an issue (or ask for a user story) to track fixing this up?

I haven't tested, but I'll spin up a Sprout node and:
* not mention scscf in the config file and verify that these plugins load
* explicitly set scscf=5054 in the config file and verify that these plugins load
* explicitly set scscf=0 in the config file and verify that these plugins don't load

### DIFF
--- a/sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_chronos_plugin.py
+++ b/sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_chronos_plugin.py
@@ -83,6 +83,6 @@ class SproutChronosPlugin(SynchroniserPluginBase):
 
 
 def load_as_plugin(local_server, local_site, remote_site):
-    is_icscf = (run_command('. /etc/clearwater/config; [[ "$scscf" == 0 ]]') == 0)
-    if not is_icscf:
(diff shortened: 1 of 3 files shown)


sprout-base.root/usr/share/clearwater/clearwater-cluster-manager/plugins/sprout_chronos_plugin.py
```diff
@@ -83,4 +83,6 @@ def on_leaving_cluster(self, cluster_view):
 
 
 def load_as_plugin(local_server, local_site, remote_site):
-    return SproutChronosPlugin(local_server, local_site, remote_site)
+    is_icscf = (run_command('. /etc/clearwater/config; [[ "$scscf" == 0 ]]') == 0)
```
eleanor-merry: Can this be is_icscf_only to make it clearer when the plugins shouldn't run?

eleanor-merry:
OK

rkd-msw:
Tested.

PR #1052, opened 2015-05-22
[Reviewer: Andy] Make sure the HTTP stack handlers are always using valid memory. 

AlexHockey:
This PR moves the various HTTP handlers we use into the same scope as the `sem_wait` that main spends most of it's time in (so that the memory the handlers use won't be reused). 

This manifested as an HTTP flow from chronos to sprout appearing in the SAS call flow. I have confirmed that this change fixes that issue.


bossmc:
Looks good.

Issue #1018, opened 2015-04-24
Sprout doesn't time out SESSION_TERMINATED ASs fast enough


mirw:
> I have Project Clearwater and a 2-site AS, fronted by a pool of 3 proxies.
> I have a SIPp script running call load – 400 callers and 400 callees.
> All callers have originating services on site 1 with DefaultHandling SESSION_TERMINATED.
> All callees have terminating services on site 2 with DefaultHandling SESSION_TERMINATED.
> If I take down the first proxy in the pool, Sprout tries to use it and declares it dead after 4 seconds.
> **It then takes ~30s for Sprout to try the second proxy in the pool.**
> This proxy forwards the call on to the correct AS, which sends back a 100 Trying.
> Sprout rejects the 100 Trying, sending a Cancel, presumably because it has already torn down the call in those 30s.

After 4s, our "liveness timer" pops and we generate an internal 408 response but this doesn't trigger a retry:

```
24-04-2015 11:31:17.475 UTC Debug sproutletproxy.cpp:89: Sproutlet timer popped, id = 140447578697408
24-04-2015 11:31:17.475 UTC Debug sproutletproxy.cpp:1492: Timer has popped
24-04-2015 11:31:17.475 UTC Verbose sproutletproxy.cpp:1259: scscf-0x7fbc801b8b10 cancelling fork 0, reason = 0
24-04-2015 11:31:17.475 UTC Debug pjutils.cpp:697: Cloned tdta0x7fbc804cbf20 to tdta0x7fbc5c559000
24-04-2015 11:31:17.475 UTC Debug sproutletproxy.cpp:1067: Remove top Route header Route: <sip:odi_phufA66FgU@10.248.0.50:5054;lr;orig>
24-04-2015 11:31:17.475 UTC Debug sproutletproxy.cpp:1500: Adding message 0x7fbc5c559610 => txdata 0x7fbc5c5590a8 mapping
24-04-2015 11:31:17.475 UTC Debug sproutletproxy.cpp:1500: Adding message 0x7fbc5d19c850 => txdata 0x7fbc5d19c2e8 mapping
24-04-2015 11:31:17.475 UTC Debug sproutletproxy.cpp:1507: Removing message 0x7fbc5c559610 => txdata 0x7fbc5c5590a8 mapping
24-04-2015 11:31:17.475 UTC Debug sproutletproxy.cpp:1295: Free message tdta0x7fbc5c559000
[...]

PR #1054, opened 2015-05-22
[Reviewer: Steve] Make SIP TCP connect timeout configurable

mirw:
Steve,

Please can you review the second half of my fix to the TCP connection timeout aspect of #1018?

This simply plumbs the connection timeout through to PJSIP.

Note that even after this fix there are still issues surrounding AS timeouts.  In particular,

*   if you don't specify TCP transport on the AS URI, sprout falls back to UDP after the TCP connection attempt times out, and so doesn't try other TCP targets - obviously, the work-around to this is simply to specify TCP transport
*   this fix handles the case where an AS takes too long to connect, but it doesn't handle the case where an AS accepts the connection but doesn't respond to the SIP request - maybe that's OK, though, or at least much less mainline (I think we're getting into undefined parts of the specs here).

Cheers,

Matt

### DIFF
--- a/include/cfgoptions.h
+++ b/include/cfgoptions.h
@@ -131,6 +131,7 @@ struct options
   int                                  exception_max_ttl;
   int                                  sip_blacklist_duration;
   int                                  http_blacklist_duration;
+  int                                  sip_tcp_connect_timeout;
 };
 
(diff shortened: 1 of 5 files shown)


steelyvoid:
Looks good

PR #1056, opened 2015-05-25
[Reviewer: Andy] Fix path to plugin_utils

mirw:
I picked up the latest code out of master and it didn't work. It looks like plugin_utils.py has been moved into etcd_shared, but the references here and in clearwater-etcd haven't been updated.


bossmc:
Yep, looks fine.

Issue #1055, opened 2015-05-24
All deployments expect remote Memcached cluster - even if not GR


mirw:
This increases load on the deployment and causes log spam as follows.

    24-05-2015 20:42:07.205 UTC Error memcachedstore.cpp:514: Failed to read data for reg\\sip:2010006520@mirw.cw-ngv.com from 0 replicas

It might also cause incorrect behavior (although I can't immediately see how).

Commit https://github.com/Metaswitch/sprout/commit/a4221db307465aa1b2cfb6a84dcd7258a1b5ae7f seems to have removed the checks.

Further, clearwater-etcd seems to assume that all clusters are GR - see https://github.com/Metaswitch/clearwater-etcd/blob/master/clearwater-cluster-manager.root/usr/share/clearwater/clearwater-cluster-manager/scripts/load_from_memcached_cluster, which (due to bash behavior) always receives a `remote_site_name` parameter (possibly actually what was supposed to be the `node_type` parameter).

rkd-msw:
This is intentional - since Sprout now has to accept an empty remote memcached cluster (so that the first Sprout node that comes up has valid configuration), it makes sense for a non-GR Sprout to just be "one with no remote memcached servers configured" (like how a non-GR Cassandra is just one with no servers in another datacenter configured).

https://github.com/Metaswitch/sprout/commit/743cfd1c00c5c37789265ab9cea2ce294ddb829f should have changed all the checks of "do I have a remote store" to "do I have a remote store and does it have servers" (so there should be no change to load, behaviour or logging) - though I missed one check at https://github.com/Metaswitch/sprout/blob/743cfd1c00c5c37789265ab9cea2ce294ddb829f/sprout/registrar.cpp#L259 which is probably causing the log you're still seeing.

PR #1058, opened 2015-05-26
[Reviewer: Matt] Add has servers checks

eleanor-merry:
This adds the missing has_servers checks. 

Part of the fix for #1055


mirw:
I'm not sure I like the general idea of passing this useless object around everywhere, but otherwise looks sensible.  You can go ahead and merge.

PR #1061, opened 2015-05-26
[Reviewer: Matt] Add in check for I-CSCF

### LINKED ISSUE #1060: An I-CSCF without S-CSCF doesn't support ENUM
Because the enum_service is only enabled for an S-CSCF - https://github.com/Metaswitch/sprout/blob/dev/sprout/main.cpp#L1527. 

Fix is to add a check for whether I-CSCF is enabled.

eleanor-merry:
Check whether I-CSCF or S-CSCF function is enabled when creating the enum_service

Fixes #1060

### DIFF
--- a/sprout/main.cpp
+++ b/sprout/main.cpp
@@ -1524,9 +1524,9 @@ int main(int argc, char* argv[])
                                        hss_comm_monitor);
   }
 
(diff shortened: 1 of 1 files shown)


mirw:
Looks good.

PR #1063, opened 2015-05-27
[Reviewer: Andy] Change how S-CSCF value is checked to avoid error logs

eleanor-merry:
The cluster-manager logs contain the following in the mainline case:
```
27-05-2015 17:55:34.406 UTC ERROR plugin_utils.py:58: Command . /etc/clearwater/config; [ "x$scscf" = "x0" ] failed with return code 1 and printed output ''
```
This PR changes how we check the SCSCF value to avoid the misleading error log. Tested live


bossmc:
Looks ok, even if this configuration flag is terribly designed....

BTW: `echo -n` will stop `echo` from adding the `\n` to the end of the output.

PR #1065, opened 2015-05-28
[Reviewer: Andy] Etcd plugins now receive parameter objects

AlexHockey:
Co-req change to accompany Metaswitch/clearwater-etcd#70

### DIFF
--- a/sprout-base.root/usr/share/clearwater/clearwater-config-manager/plugins/sprout_scscf_json_plugin.py
+++ b/sprout-base.root/usr/share/clearwater/clearwater-config-manager/plugins/sprout_scscf_json_plugin.py
@@ -44,8 +44,8 @@ from sprout_json_plugin import SproutJsonPlugin
 _log = logging.getLogger("sprout_scscf_json_plugin")
 
 class SproutSCSCFJsonPlugin(SproutJsonPlugin):
-    def __init__(self):
+    def __init__(self, _params):
         super(SproutSCSCFJsonPlugin, self).__init__("/etc/clearwater/s-cscf.json", "scscf_json")
 
-def load_as_plugin():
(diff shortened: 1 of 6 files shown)


sprout-base.root/usr/share/clearwater/clearwater-config-manager/plugins/sprout_scscf_json_plugin.py
```diff
@@ -44,8 +44,8 @@
 _log = logging.getLogger("sprout_scscf_json_plugin")
 
 class SproutSCSCFJsonPlugin(SproutJsonPlugin):
-    def __init__(self):
+    def __init__(self, _params):
```
bossmc: Is this a convention I'm not aware of?  Does the underscore indicate "unused"?

bossmc:
One comment (that applies throughout and probably applies to the other PRs).

sprout-base.root/usr/share/clearwater/clearwater-config-manager/plugins/sprout_scscf_json_plugin.py
AlexHockey: Other repos were already doing this so I did the same thing here. It doesn't match a convention I know about but I quite like it :-)

Generally in python you normally denote an unused variable with an underscore. This is  fine for tuple unpacking (e.g. `host, _, _ = get_connection_info()` ignores the protocol and port that the hypothetical function returns). It doesn't work well for function arguments - you can't use two underscores as you will be using duplicate argument names.

bossmc: Ok, happy with that.

PR #1072, opened 2015-06-04
Reviewer AMC: Suppress Tel URI conversion in I-CSCF if number not global

steelyvoid:
A fix to a technical spec misalignment. TS 24.229 5.3.2.1 says that "user=phone" => Tel URI conversion in I-CSCF on Initial Requests should only be done for global numbers ("...with the user part starting with a + ...").

Tested with a UT that verifies that an INVITE for a "user=phone" SIP URI with a non global number does not match a location record for the corresponding Tel URI and is rejected


sprout/icscfsproutlet.cpp
```diff
@@ -493,14 +493,16 @@ void ICSCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
...
+    // If the Req URI is a SIP URI with the user=phone parameter set, is not a
+    // GRUU and the user part starts with '+' (i.e. is a global phone number),
+    // we should replace it with a tel URI, as per TS24.229 5.3.2.1.
     if (PJSIP_URI_SCHEME_IS_SIP(uri))
     {
       pjsip_sip_uri* sip_uri = (pjsip_sip_uri*)uri;
 
       if ((!pj_strcmp(&sip_uri->user_param, &STR_USER_PHONE)) &&
           (PJUtils::is_user_numeric(sip_uri->user)) &&
+          (PJUtils::is_user_global(sip_uri->user)) &&
```
bossmc: Some discussion with @eleanor-merry and @rkd-msw suggests that this change should be controlled by the `enforce_global_only_lookups` flag (expanding the meaning of this flag to be "be strict about what counts as global").  Thoughts?

bossmc:
One question inline.

bossmc:
Superseded by #1079

PR #1076, opened 2015-06-05
[Reviewer: Matt] Changes to help with AS interop

AlexHockey:
Please can you review this PR that makes the following changes:

* The default handling timeouts are now configurable. Setting a timeout to 0 disables the timeout completely. 
* Sprout can now be configured to identify domain names as stateless proxies. If a request to a stateless proxy times out the target is not blacklisted. 

I have tested this is UT as far as possible (note that session terminated default handling is a bit broken, but not by this change). 

I plan to do the following live testing:

* Check the live tests work. 
* Check blacklisting works by:
    * Setting up a 2-node Manglewurzel cluster
    * Setting up SRV to route to manglewurzel-1 preferentially. 
    * SIGSTOP-ing manglewurzel-1 and checking it is/isn't blacklisted (Depending on whether its a stateless proxy). 
* Check timeout configuration works, again using a mangelwurzel cluster. 

I will document the new config values in the [config options reference](http://clearwater.readthedocs.org/en/latest/Clearwater_Configuration_Options_Reference/index.html). The timers will be "Advanced Options" and the stateless proxy option an "Experimental Option".


include/scscfsproutlet.h
```diff
@@ -83,7 +86,9 @@ class SCSCFSproutlet : public Sproutlet
                  ACRFactory* acr_factory,
                  bool user_phone,
                  bool global_only_lookups,
-                 bool override_npdi);
+                 bool override_npdi,
+                 int session_continue_timeout,
```
mirw: Apply default timeouts here?  (Probably never used but feels more elegant.)

sprout-base.root/etc/init.d/sprout
```diff
@@ -178,6 +178,9 @@ get_daemon_args()
         [ -z "$min_token_rate" ] || min_token_rate_arg="--min-token-rate=$min_token_rate"
         [ -z "$exception_max_ttl" ] || exception_max_ttl_arg="--exception-max-ttl=$exception_max_ttl"
         [ -z "$signaling_namespace" ] || namespace_prefix="ip netns exec $signaling_namespace"
+        [ -z "$session_continue_timeout_ms" ] || session_continue_timeout_arg="--session-continue-timeout=$session_continue_timeout_ms"
+        [ -z "$session_terminated_timeout_ms" ] || session_terminated_timeout_arg="--session-terminated-timeout=$session_continue_timeout_ms"
+        [ -z "$stateless_proxies" ] || stateless_proxies_arg="--stateless-proxies=$stateless_proxies"
```
mirw: I'm not really sure why we still use this approach - the `DAEMON_ARGS="$DAEMON_ARGS ..."` approach we use below is more concise.  However, I don't feel strongly.

sprout/basicproxy.cpp
```diff
@@ -1825,17 +1838,34 @@ void BasicProxy::UACTsx::on_tsx_state(pjsip_event* event)
...
+        LOG_DEBUG("Request to server timed-out");
+
+        if (!_stateless_proxy)
+        {
+          // The next hop is NOT a stateful proxy so if it hasn't responded then
+          // it should be blacklisted.  We don't blacklist stateless proxies to
+          // avoid blacklisting them due to an unresponsive server further
+          // downstream.
+          LOG_DEBUG("Next hop is NOT a stateless-proxy - blacklist");
+          PJUtils::blacklist_server(_servers[_current_server]);
```
mirw: Should we be SAS-logging?  I'm not just thinking about here, but how would we debug why/whether a host is being blacklisted (or not).

sprout/main.cpp
```diff
@@ -328,6 +335,20 @@ static void usage(void)
        "                            The amount of time to blacklist an HTTP peer when it is unresponsive.\n"
        "     --sip-tcp-connect-timeout <milliseconds>\n"
        "                            The amount of time to wait for a SIP TCP connection to establish.\n"
+       "     --session-continue-timeout\n"
+       "                            If an Application Server with default handling of 'continue session'\n"
+       "                            and is unresponsive, this is the time that sprout will wait (in ms)\n"
```
mirw: Typo: "If an X *and* is unresponsive"

sprout/main.cpp
```diff
@@ -328,6 +335,20 @@ static void usage(void)
...
+       "                            before bypassing the AS and moving onto the next AS in the chain.\n"
+       "     --session-terminated-timeout\n"
+       "                            If an Application Server with default handling of 'terminate session'\n"
+       "                            and is unresponsive, this is the time that sprout will wait (in ms)\n"
```
mirw: And here.

sprout/main.cpp
```diff
@@ -328,6 +335,20 @@ static void usage(void)
...
+       "                            before terminating the session.\n"
+       "     --stateless-proxies\n"
+       "                            A comma separated list of domain names that are considered to be SIP\n"
```
mirw: Are they considered or are they actually SIP proxies?  ...ah, I think you mean that Sprout will consider to be stateless proxies.  Maybe just "treated as"?

sprout/main.cpp
```diff
@@ -397,6 +418,7 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
   int opt_ind;
   int reg_max_expires;
   int sub_max_expires;
+  std::vector<std::string> stateless_proxies;
```
mirw: Should we move this declaration nearer the use?

mirw:
Looks good.  Just minor mark-ups apart from question of whether we should SAS log.

sprout/main.cpp
AlexHockey: I tried that originally but if you put the variable declaration in the case block gcc complains about the switch statement crossing the initialization. But it turns out you can work around that by declaring a new scope within the case block.

sprout/basicproxy.cpp
AlexHockey: I don't *think* so. Blacklisting is a node-scope action so is not something we would normally SAS log. I think a better solution would be to add `LOG_STATUS` lines to track when servers are (un)blacklisted. I don't have time to do this under this PR but will raise an issue to cover it.

mirw: Not `LOG_STATUS`.  Either SAS or ENT logs.

PR #1077, opened 2015-06-05
[Reviewer: Alex] Add configurable SIP send timeout

mirw:
Alex,

Please can you review this enhancement to add a `sip_tcp_send_timeout` option that controls the amount of time we'll wait for acknowledgement of a TCP PSH.

I've not UT-ed, because there isn't anything to UT, but I have tested live (from Bono to Sprout) and saw a transaction fail quickly with a transport failure if Sprout drops off the network.

Matt


AlexHockey:
Looks good to me.

PR #1079, opened 2015-06-05
Review AMC:  Only translate user=phone SIP URIs to tel URIs in ICSCF if number is global (or enforce_global_lookups not set)

steelyvoid:
A fix to a technical spec misalignment. TS 24.229 5.3.2.1 says that "user=phone" => Tel URI conversion in I-CSCF on Initial Requests should only be done for global numbers ("...with the user part starting with a + ...") if the enforce_global_lookups feature is turned on.

Tested with a UT that verifies that an INVITE for a "user=phone" SIP URI with a non global number does not match a location record for the corresponding Tel URI and is rejected, and another that verifies that rejection doesn't happen if enforce_global_lookups is off

### DIFF
--- a/include/icscfsproutlet.h
+++ b/include/icscfsproutlet.h
@@ -75,7 +75,7 @@ public:
                  SCSCFSelector* scscf_selector,
                  EnumService* enum_service,
                  bool enforce_global_only_lookups,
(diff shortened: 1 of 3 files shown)


bossmc:
Looks good.

PR #1081, opened 2015-06-05
[Reviewer: Matt] Fix up terminating tel: URI handling, including when User-Data is malformed (ready to merge)

### LINKED ISSUE #1070: Sprout fails if the subscriber it requested isn't available in the IFCs
It should probably return 400 Bad Request in this case.

rkd-msw:
Fixes #1070. It looks like most of the problem was that terminating tel: URI handling was just a bit bad - if an AS sent a user=phone URI to the I-CSCF (unsolicited outdial), the I-CSCF converted it to a tel: URI, found it in the HSS and forwarded it to the S-CSCF as a tel: URI, things would go terribly wrong because it wasn't a sip: URI.  I've fixed that up and UTed it (with and without malformed User-Data).


sprout/scscfsproutlet.cpp
```diff
@@ -1224,7 +1224,9 @@ void SCSCFSproutletTsx::route_to_target(pjsip_msg* req)
              ((pjsip_sip_uri*)req_uri)->maddr_param.ptr);
     send_request(req);
   }
-  else if ((!PJUtils::is_home_domain(req_uri)) &&
+  else if (PJSIP_URI_SCHEME_IS_SIP(req_uri) &&
+           (!PJUtils::is_home_domain(req_uri)) &&
+           (!PJUtils::is_home_domain(req_uri)) &&
```
mirw: Why the duplicate test?

sprout/scscfsproutlet.cpp
```diff
@@ -1233,24 +1235,13 @@ void SCSCFSproutletTsx::route_to_target(pjsip_msg* req)
              PJUtils::uri_to_string(PJSIP_URI_IN_REQ_URI, req_uri).c_str());
     send_request(req);
   }
-  else if (PJSIP_URI_SCHEME_IS_SIP(req_uri))
+  else
   {
     // The Request-URI contains a home domain, so route to any UE bindings
```
mirw: This comment seems a bit misleading - it looks like we handle TEL URIs here.

sprout/scscfsproutlet.cpp
```diff
@@ -1259,7 +1250,7 @@ void SCSCFSproutletTsx::route_to_ue_bindings(pjsip_msg* req)
 {
   // Get the public user identity corresponding to the RequestURI.
   pjsip_uri* req_uri = req->line.req.uri;
-  std::string public_id = PJUtils::aor_from_uri((pjsip_sip_uri*)req_uri);
+  std::string public_id = PJUtils::public_id_from_uri(req_uri);
```
mirw: Is there still value in keeping the `aor_from_uri` function around?

mirw:
Looks good.  Have you live-tested at all (and should we)?

rkd-msw:
I can't really see what live testing would add beyond my UTs - I'd really be setting up fairly artificial call scenarios (e.g. misconfigured IFCs, tel:-URI only users) and making test calls, which the UTs do just as effectively and quicker.

mirw:
If you're confident, fine.  I think my concern was that there might be interactions wider than the scope of a single I- or S-CSCF (although it's hard to remember from almost 2 months ago).

PR #1082, opened 2015-06-08
[Reviewer: RKD] Restart clearwater-infrastructure before starting SIPp

graemerobertson:
***Not to be merged whilst the code is frozen***

Hi Rob,

Can you have a look at my proposed change to restart clearwater-infrastructure before starting SIPp? Any reason why this isn't a good idea? Saves having to remember to do it manually, and there doesn't seem to be much reason not to just do it automatically.

Tested by adding the line to an existing clearwater-sip-stress node and starting clearwater-sip-stress.

I can't find any associated doc changes so that's good...

Thanks,
Graeme


rkd-msw:
Yes, looks good.

Issue #1084, opened 2015-06-09
We should check that clearwater-cluster/config-manager is installed before reloading it


eleanor-merry:
As otherwise we get warnings during install/upgrade if we're not using etcd. This should be fixed in Sprout/Memento/Homer/Homestead/Ralf

eleanor-merry:
Fixed in #1085, https://github.com/Metaswitch/memento/pull/81/files, https://github.com/Metaswitch/ralf/pull/159, https://github.com/Metaswitch/homestead/pull/230, https://github.com/Metaswitch/crest/pull/217

PR #1085, opened 2015-06-09
[Reviewer: Andy] Check for install

eleanor-merry:
Checks whether cluster/config manager is installed before stopping them. If you're happy with this, I'll make the same fix to Memento/Ralf/Homer/Homestead

Part of the fix for #1084

### DIFF
--- a/debian/sprout-base.postinst
+++ b/debian/sprout-base.postinst
@@ -97,10 +97,10 @@ case "$1" in
         # Stop the cluster/config managers, so that it is restarted by Monit
         # and picks up the new plugins. We check whether the process is 
         # installed to avoid warning logs when etcd isn't being used
-        if [ -e "/etc/init.d/clearwater-cluster-manager" ]; then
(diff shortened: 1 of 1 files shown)


debian/sprout-base.postinst
```diff
@@ -95,9 +95,14 @@ case "$1" in
...
-        service clearwater-cluster-manager stop || /bin/true
-        service clearwater-config-manager stop || /bin/true
+        # and picks up the new plugins. We check whether the process is 
+        # installed to avoid warning logs when etcd isn't being used
+        if [ -e "/etc/init.d/clearwater-cluster-manager" ]; then
+          service clearwater-cluster-manager stop || /bin/true
+        fi  
+        if [ -e "/etc/init.d/clearwater-config-manager" ]; then
+          service clearwater-config-manager stop || /bin/true
+        fi
```
bossmc: I'd prefer to test with `-f` (true if the path is a file, not a directory) and, for bonus points `-x` (true if path is executable).

bossmc:
One comment and one optional improvement.

PR #1093, opened 2015-06-12
[Reviewer: Matt] Refactor SNMP function to make Sprout an SNMP agent

rkd-msw:
Matt,

Can you review this new SNMP code? It's generally as discussed in the design.

Some notes:
* snmp_latency_table.cpp isn't quite up to code review standard, but I've included it so that you can see what's involved in actually creating an SNMP table, given the infrastructure
* You'll see some CW_LOG_DEBUG references - because the SNMP header files include syslog, which defines a conflicting LOG_DEBUG macro, I've had to change all the LOG_X instances in cpp-common to CW_LOG_X. (I'll talk to you and see if there are any alternatives.) I've excluded most of those from this pull request because it's just clutter.
* There are some valgrind warnings which I've suppressed - these only happen on table creation (so the leaks will be fixed in size and not grow over time), and I don't think it's worth tracking them down now, since we're about to move to 14.04 (and a different net-snmp).
* I'm probably going to actually put these files in cpp-common, but reviewing it like this means you can see the new code alongside the integration.


rkd-msw:
Reclaiming to apply design feedback before review.

rkd-msw:
Superseded - I moved this code into cpp-common and it was reviewed there.

Issue #1091, opened 2015-06-11
Sprout loops ACKs to itself after responding with 503 to an INVITE


plfrancis:
After Sprout returns a 503 for an INVITE due to overload controls, it begins looping ACKs to itself until Max-Forwards has "expired". This additional processing triggers a LIR over the Cx for each ACK which adds to system latency, thereby "reinforcing" the overload condition (the overload was initially caused by Cx interface latency). A SAS trace has been collected.

JimBunch:
Raising the priority to High since the system does not recover from the overload condition.

matt-williams:
The set-up here is vm1.richlab.datcon.co.uk resolves to I-CSCF, sprout.vm1.richlab.datcon.co.uk resolves to S-CSCF.

The flow here is

*   Perimeta sends an INVITE (with request-URI vm1... and route sprout...).
*   Clearwater (S-CSCF) rejects it with a 503.
*   Perimeta sends an ACK (with request-URI vm1... and route sprout...).
*   Clearwater (S-CSCF) recognizes it as out-of-dialog and forwards it to vm1 (I-CSCF).
*   Clearwater (I-CSCF) does an LIR, finds the S-CSCF and forwards it (with request-uri vm1... and route sprout1...).
*   Clearwater (S-CSCF) recognizes it as out-of-dialog and forwards it to vm1 (I-CSCF).
*   ...

I think the key issue here is that we perform LIRs at all for ACKs - ACKs should either be in-dialog (in which case we follow the route-set) or hop-by-hop (to a negative INVITE).  If the I-CSCF receives an out-of-dialog ACK, I think it should reject it.  I need to check the specs to check that's correct.

There's an argument that we should apply overload processing to ACKs as well as other requests, but this is difficult - if we don't forward them, they'll just be resent, the call will eventually fail and they'll probably be retried - probably better to allow them to carry on.

mirw:
I've coded a fix (https://github.com/Metaswitch/sprout/tree/iss1091), but not yet tested - I've been pulled to investigate another issue.

rkd-msw:
I'll UT and possibly live test.

rkd-msw:
I've had a go in UT and I'm not sure that your fix solves the issue. I'm concerned that:
* The ACK comes in, with the request line as `ACK sip:3196914123@homedomain:5060;transport=UDP SIP/2.0` and no Route headers
* No local Sproutlet owns the request, so we apply standard SIP proxying rules
[...]

PR #1097, opened 2015-06-16
[Reviewer: Matt] Drop ACKs which the I-CSCF shouldn't be handling

rkd-msw:
Matt,

Can you review this fix for https://github.com/Metaswitch/sprout/issues/1091? I've:
* taken your initial fix
* added code to not forward requests on if they'd loop - I've made the checking quite strict to avoid false positives
* added a UT for this case
* fixed up one of the S-CSCF tests, which sent in a Request-URI of `sip:+15108580301;npdi@homedomain` and expected it to not be changed, which now falls foul of the "don't forward unchanged requests" logic. I've made this a `user=phone` URI (so that SIP routing rules aren't applied) and checked that there are no new coverage warnings. I'm going to talk to Ellie to see if she has any ideas on how to test this more rigorously.

### DIFF
--- a/sprout/sproutletproxy.cpp
+++ b/sprout/sproutletproxy.cpp
@@ -787,28 +787,52 @@ void SproutletProxy::UASTsx::schedule_requests()
       }
       else
       {
-        // No local Sproutlet, proxy the request.
         LOG_DEBUG("No local sproutlet matches request");
-        size_t index;
-
-        pj_status_t status = allocate_uac(req.req, index);
-
-        if (status == PJ_SUCCESS)
+        
+        // Check whether:
+        // - there are no Route headers
+        // - the Request-URI points to us
+        // - no Sproutlet has changed the Request-URI during processing
+        //
+        // If all these things are true, then forwarding this on would cause a loop, so we should
+        // drop it.
+        pjsip_uri* original_uri = _req->msg->line.req.uri;
+        pjsip_uri* next_hop = PJUtils::next_hop(req.req->msg);
+        if (next_hop == req.req->msg->line.req.uri &&
+            (PJUtils::is_uri_local(next_hop) ||
+             PJUtils::is_home_domain(next_hop)) &&
+            (PJUtils::uri_to_string(PJSIP_URI_IN_REQ_URI,
+                                    original_uri) ==
+             PJUtils::uri_to_string(PJSIP_URI_IN_REQ_URI,
+                                    req.req->msg->line.req.uri)))
         {
-          // Successfully set up UAC transaction, so set up the mappings and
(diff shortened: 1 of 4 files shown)


sprout/sproutletproxy.cpp
```diff
@@ -787,28 +787,52 @@ void SproutletProxy::UASTsx::schedule_requests()
...
-          if (req.req->msg->line.req.method.id != PJSIP_ACK_METHOD)
-          {
-            _dmap_uac[req.upstream] = _uac_tsx[index];
-            _umap[(void*)_uac_tsx[index]] = req.upstream;
-          }
-
-          // Send the request.
-          _uac_tsx[index]->send_request();
+          LOG_WARNING("No local sproutlet accepted request, the URI has not been changed, and the next route is local"
+                      " - dropping message to avoid loops");
```
mirw: Should we be dropping the request, or should we be responding to it?  ...or should we only be dropping it if it's an ACK (if you want a quick fix)?

sprout/sproutletproxy.cpp
```diff
@@ -787,28 +787,52 @@ void SproutletProxy::UASTsx::schedule_requests()
...
-
-        if (status == PJ_SUCCESS)
+        
+        // Check whether:
+        // - there are no Route headers
+        // - the Request-URI points to us
+        // - no Sproutlet has changed the Request-URI during processing
+        //
+        // If all these things are true, then forwarding this on would cause a loop, so we should
+        // drop it.
```
mirw: I think we should be clear that this is a temporary fix (and maybe even link to this issue).

sprout/sproutletproxy.cpp
```diff
@@ -787,28 +787,52 @@ void SproutletProxy::UASTsx::schedule_requests()
...
-        // No local Sproutlet, proxy the request.
         LOG_DEBUG("No local sproutlet matches request");
-        size_t index;
-
-        pj_status_t status = allocate_uac(req.req, index);
-
-        if (status == PJ_SUCCESS)
+        
+        // Check whether:
+        // - there are no Route headers
```
mirw: To be precise, there are no Route headers *on the new request*.  There might have been Route headers on the original request.  Is that OK?

sprout/sproutletproxy.cpp
mirw: I don't think this should be a `LOG_WARNING` - it should be a SAS event.

mirw:
Looks good.  A few comments inline.

rkd-msw:
I've made a safer fix in https://github.com/Metaswitch/sprout/pull/1099.

Issue #1095, opened 2015-06-15
Sprout crashing at low call rate with VoLTE TAS


plfrancis:
Running 10 - 15 CPS traffic with 15,000 subscribers doing 5 minute registration causes Sprout to crash. The back-trace is shown below. A diagnostics bundle was collected.

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f08871ce688 in ?? () from /lib/x86_64-linux-gnu/libgcc_s.so.1
(gdb) bt
#0  0x00007f08871ce688 in ?? () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#1  0x00007f08871cf6f8 in _Unwind_Backtrace () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#2  0x00007f0886f04e26 in __GI___backtrace (array=<optimized out>, size=32) at ../sysdeps/x86_64/backtrace.c:109
#3  0x00000000004bbc7d in Logger::backtrace (this=0x270a010, data=<optimized out>) at /var/build/CC_7.0.0-WIP/cc/sprout/modules/cpp-common/src/logger.cpp:277
#4  0x0000000000571a4d in Log::backtrace (fmt=fmt@entry=0x6f28e0 "Signal %d caught") at /var/build/CC_7.0.0-WIP/cc/sprout/modules/cpp-common/src/log.cpp:185
#5  0x00000000005bda3c in signal_handler (sig=11) at main.cpp:1008
#6  <signal handler called>
#7  0x0000000000000038 in ?? ()
#8  0x00007f087c15552d in SCSCFSproutletTsx::on_rx_response (this=0x7f0824747170, rsp=0x7f085c71c680, fork_id=0) at scscfsproutlet.cpp:532
#9  0x00000000005b3825 in SproutletWrapper::rx_response (this=0x7f082459be50, rsp=0x7f085c71c118, fork_id=0) at sproutletproxy.cpp:1438
#10 0x00000000005b39e0 in SproutletProxy::UASTsx::tx_response (this=0x7f0824298220, downstream=downstream@entry=0x7f082499d0d0, rsp=0x7f085c71c118)
    at sproutletproxy.cpp:893
#11 0x00000000005b3c3f in SproutletWrapper::tx_response (this=this@entry=0x7f082499d0d0, rsp=<optimized out>) at sproutletproxy.cpp:1734
[...]

PR #1098, opened 2015-06-16
[Reviewer: Andy] Don't retry SIP transactions on SIP-level timeout - it's too slow

mirw:
Andy,

Please can you review this partial fix to issue #1095?  It simply disables retry on transaction-level timeouts - they're too slow to be any use and cause extra work (and possible race conditions/crashes) when they're hit.

Thanks,

Matt


mirw:
I've UT-ed but not live-tested - I was going to ask Phil to do this.

bossmc:
Looks fine.

PR #1099, opened 2015-06-16
[Reviewer: Matt] Drop rogue ACKs within the I-CSCF

rkd-msw:
This is a less-wide-ranging (and therefore safer) fix for issue #1091.

I haven't defined a ICSCFSproutletDropItTsx - this seemed like overkill, so I've just dropped it in the existing ICSCFSproutletTsx.

UTed but not tested live.

### DIFF
--- a/sprout/ut/icscfsproutlet_test.cpp
+++ b/sprout/ut/icscfsproutlet_test.cpp
@@ -3117,3 +3117,26 @@ TEST_F(ICSCFSproutletTest, RouteOrigInviteBadServerName)
 
   delete tp;
 }
+
+// Test the case where the I-CSCF receives an ACK. This is not valid and should be dropped.
+TEST_F(ICSCFSproutletTest, RouteOutOfDialogAck)
+{
+  // Create a TCP connection to the I-CSCF listening port.
+  TransportFlow* tp = new TransportFlow(TransportFlow::Protocol::TCP,
+                                        stack_data.icscf_port,
+                                        "1.2.3.4",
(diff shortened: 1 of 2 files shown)


rkd-msw:
Note that the checks (ACK, no Route headers, we're in the Request-URI) are chosen to match the ACKs in https://github.com/Metaswitch/sprout/issues/1091 as precisely as possible, to avoid us dropping something we don't want to.

sprout/ut/icscfsproutlet_test.cpp
```diff
@@ -3117,3 +3117,26 @@ TEST_F(ICSCFSproutletTest, RouteOrigInviteBadServerName)
...
+  // Inject an ACK request to a local URI
+  Message msg1;
+  msg1._method = "ACK";
+  msg1._requri = "sip:3196914123@homedomain;transport=UDP";
+  inject_msg(msg1.get_request(), tp);
+
+  // Expect it to just be dropped
+  ASSERT_EQ(0, txdata_count());
+  free_txdata(); 
+  cwtest_advance_time_ms(33000L);
```
mirw: Worth commenting why this is necessary.

mirw:
Looks good.  We should prioritize fixing this properly after the short-term urgency is over.

rkd-msw:
I've raised #1100 and #1101 - I think between them, these cover the proper fix.

rkd-msw:
I'll merge once @plfrancis confirms this fixes #1091.

PR #1111, opened 2015-06-23
Reviewer Matt - Store the ACR in the AsChain

bossmc:
Matt, this makes the `AsChain` (which is reference-counted) be the owner of the `ACR` for a single S-CSCF phase (`orig`, `orig-cdiv` or `term`).  We discussed the change (and @mike-evans also proposed a similar approach).  I've tested through the existing UTs which ensure that we're not leaking any ACRs and that we're producing ACRs in the correct places, I've not tested live.

One thing that's a bit of a shame is I can't see a way to stop someone just reading `_acr` when they should be calling `get_acr()` to check the `_as_chain` first, any ideas?

### DIFF
--- a/include/scscfsproutlet.h
+++ b/include/scscfsproutlet.h
@@ -234,7 +234,7 @@ private:
 
   /// Creates an AS chain for this service role and links this service hop to
   /// it.
-  AsChainLink create_as_chain(Ifcs ifcs, std::string served_user);
+  AsChainLink create_as_chain(Ifcs ifcs, std::string served_user, ACR*& acr);
 
   /// Apply originating services for this request.
   void apply_originating_services(pjsip_msg* req);
@@ -303,6 +303,11 @@ private:
   void sas_log_start_of_sesion_case(pjsip_msg* req,
                                     const SessionCase* session_case,
                                     const std::string& served_user);
+
+  /// Fetch the ACR for the current transaction, ACRs should always be retrived
+  /// through this API, not by inspecting _acr directly.  May return NULL
+  /// in some cases.
+  ACR* get_acr();
 
   /// Pointer to the parent SCSCFSproutlet object - used for various operations
   /// that require access to global configuration or services.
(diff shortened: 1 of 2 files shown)


include/scscfsproutlet.h
```diff
@@ -304,6 +304,11 @@ class SCSCFSproutletTsx : public SproutletTsx
                                     const SessionCase* session_case,
                                     const std::string& served_user);
 
+  /// Fetch the ACR for the current transaction, ACRs should always be retrived
+  /// through this API, not by inspecting _acr directly.  May return NULL
```
mirw: Why should ACRs always be retrieved through this API?  (Maybe not an interface comment, but should be covered somewhere.)

mirw:
Looks good.  On preventing access, you could create an inner class with a private `ACR*` member.  You can call `get_acr()` on that, but can't see its `_acr` field.  A bit heavyweight, but should be robust.

Alternatively, is it worth renaming `_acr` (e.g. `_in_dialog_acr`) so that people see that it's unusual?

(Should we actually have 2 `SproutletTsxs` - one for dialog-initiating and one for in-dialog?  That's definitely more work than we want to have here, though!)

bossmc:
Yeah, an inner class would work (I considered making a SingleTsxAsChainLink that exists only to hold the ACR which is pretty much equivalent), seems like a lot of work for not much gain.

Renaming `_acr` is a little tricky, we use the ACR for in-dialog requests but also for locally-rejected requests (e.g. if we can't find the IFC for the subscriber, we'll build an ACR and use it to process the 404.  Since there's no AsChain in this case, we store that ACR on `_acr`.

Same argument argues away from having 2 `SproutletTsx`s, it isn't a clean split sadly.

mirw:
As just discussed, I think we should have an `_in_dialog_acr` and a `_local_rejection_acr`.  It's a bit verbose, but people shouldn't be using these variables (so verbosity doesn't matter) and it means people aren't going to mess it up.

bossmc:
Done.

Issue #587, opened 2014-06-03
sprout crashes should dump out SAS trail ID (if known)


mirw:
If sprout dumped out the SAS trail ID, this would make debugging stress problems much easier, as we could pull the SAS trial and so see what was going on at the time.

Marking as high-priority as this feels really useful.

bossmc:
@mirw - What's the latest on this?

mirw:
We can probably close this.  It would be nice to move SAS trail into the thread context, but it's not crucial.

PR #1113, opened 2015-06-26
[Reviewer: Alex] Dump SAS trail (and related details) out when we hit an exception

mirw:
Alex,

Please can you review this partial fix to #587?  It doesn't go the whole way, but it should help in most cases.

Cheers,

Matt


AlexHockey:
Code looks good. What testing have you done?

mirw:
Tested live.

PR #1115, opened 2015-06-26
[Reviewer: Matt] Use SNMP agent function in Sprout

rkd-msw:
Matt,

This pull request:
* switches Sprout and Bono over to the new SNMP stats in place of 0MQ ones
* fixes up the UTs to match, including stubbing out some net-snmp API calls

It doesn't completely purge the 0MQ stuff because a CDiv AS table still uses it - but I might try and address that as a bonus.

I've verified that 'make test' passes cleanly (including coverage and valgrind).


debian/control
```diff
@@ -25,7 +25,7 @@ Description: Debugging symbols for sprout-libs
 Package: sprout
 Architecture: any
 Depends: sprout-base, sprout-scscf, sprout-icscf, sprout-bgcf, sprout-mmtel-as
-Suggests: sprout-dbg, clearwater-logging, clearwater-snmp-handler-sprout, clearwater-snmp-handler-alarm
+Suggests: sprout-dbg, clearwater-logging, clearwater-snmp-alarm-agent
```
mirw: Why do we need this here when we also have it on sprout-base?

debian/control
```diff
@@ -39,7 +39,7 @@ Description: Debugging symbols for sprout, the SIP Router
 Package: sprout-base
 Architecture: any
 Depends: clearwater-infrastructure, clearwater-tcp-scalability, clearwater-log-cleanup, sprout-libs, clearwater-memcached, libboost-regex1.54.0, libboost-system1.54.0, libboost-thread1.54.0, libzmq3, libevent-pthreads-2.0-5, chronos, clearwater-socket-factory, libboost-filesystem1.54.0, astaire
```
mirw: As discussed this morning, we also need libsnmp30.

debian/control
```diff
@@ -158,7 +158,7 @@ Description: Debugging symbols for mangelwurzel, the B2BUA and SCC-AS emulator
 Package: bono
 Architecture: any
 Depends: clearwater-infrastructure, clearwater-tcp-scalability, clearwater-log-cleanup, sprout-libs, libboost-regex1.54.0, libboost-system1.54.0, libboost-thread1.54.0, libzmq3, libevent-pthreads-2.0-5, clearwater-socket-factory, libboost-filesystem1.54.0
```
mirw: We also need libsnmp30.

include/hssconnection.h
```diff
@@ -61,7 +61,12 @@ class HSSConnection
   HSSConnection(const std::string& server,
                 HttpResolver* resolver,
                 LoadMonitor *load_monitor,
-                LastValueCache *stats_aggregator,
+                SNMP::IPCountTable* homestead_count_tbl,
```
mirw: Should we bundle these into an object?  It's a bit unwieldy passing 5 arguments around just for statistics.

sprout/bono.cpp
```diff
@@ -152,6 +152,13 @@ static ACRFactory* icscf_acr_factory;
 static bool edge_proxy;
 static pjsip_uri* upstream_proxy;
 static ConnectionPool* upstream_conn_pool = NULL;
+
+
+static oid BONO_CONNECTED_CLIENTS_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 2, 1 };
+static oid BONO_CONNECTED_SPROUTS_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 2, 3, 1 };
+static SNMP::IPCountTable* sprout_ip_tbl = NULL;
+static SNMP::U32Scalar* flow_count = NULL;
```
mirw: Surely we should pull this out into a separate module - it feels like clutter in the main Bono file.

sprout/connection_pool.cpp
```diff
@@ -471,53 +468,18 @@ int ConnectionPool::recycle_thread(void* p)
...
 void ConnectionPool::decrement_connection_count(pjsip_transport *trans)
 {
   std::string host = PJUtils::pj_str_to_string(&trans->remote_name.host);
-  assert(_host_conn_count.find(host) != _host_conn_count.end());
-
-  if (_host_conn_count[host] == 1)
+  if (_sprout_count_tbl->get(host)->decrement() == 0)
   {
-    _host_conn_count.erase(host);
+    _sprout_count_tbl->remove(host);
```
mirw: Is there a race condition here where the connection count has been incremented by someone else in the meantime?

sprout/connection_pool.cpp
```diff
@@ -471,53 +468,18 @@ int ConnectionPool::recycle_thread(void* p)
...
-}
-
-
 void ConnectionPool::decrement_connection_count(pjsip_transport *trans)
 {
   std::string host = PJUtils::pj_str_to_string(&trans->remote_name.host);
-  assert(_host_conn_count.find(host) != _host_conn_count.end());
-
-  if (_host_conn_count[host] == 1)
+  if (_sprout_count_tbl->get(host)->decrement() == 0)
```
mirw: What happens if you decrement a counter that's already 0?  (I know this shouldn't happen, but might if you hit the race condition below.)

sprout/main.cpp
```diff
@@ -211,6 +213,22 @@ QuiescingManager* quiescing_mgr;
...
+static oid SPROUT_QUEUE_SIZE_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 3, 8 };
+static oid SPROUT_REQUEST_COUNT_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 3, 6 };
+static oid SPROUT_OVERLOAD_COUNT_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 3, 7 };
+
+static oid SPROUT_HOMESTEAD_COUNT_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 3, 3, 1 };
+static oid SPROUT_HOMESTEAD_LATENCY_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 3, 3, 2 };
+static oid SPROUT_HOMESTEAD_MAR_LATENCY_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 3, 3, 3 };
+static oid SPROUT_HOMESTEAD_SAR_LATENCY_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 3, 3, 4 };
+static oid SPROUT_HOMESTEAD_UAR_LATENCY_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 3, 3, 5 };
+static oid SPROUT_HOMESTEAD_LIR_LATENCY_OID[] = { 1, 2, 826, 0, 1, 1578918, 9, 3, 3, 6 };
```
mirw: As above, it feels a bit clunky to put all this lot in main.cpp.

sprout/main.cpp
```diff
@@ -1571,9 +1659,8 @@ int main(int argc, char* argv[])
     ralf_connection = new HttpConnection(opt.ralf_server,
                                          false,
                                          http_resolver,
-                                         "connected_ralfs",
+                                         NULL, // No SNMP table for connected Ralfs
```
mirw: How are we tracking this deficiency?

sprout/ut/dialog_tracker_test.cpp
```diff
@@ -44,9 +44,12 @@
 #include "utils.h"
 #include "siptest.hpp"
 #include "dialog_tracker.hpp"
+#include "snmp_scalar.h"
 
 using namespace std;
 
+static SNMP::U32Scalar fake_connection_count("", NULL, 0);
```
mirw: Does this work?  I thought we couldn't declare these statically because they'd try to write into netsnmp data structures before it had been initialized?

mirw: Worth a comment that defining this statically can only be done in UT because we've stubbed out the real implementation of netsnmp?  If you did this in production code, it would probably crash because netsnmp hadn't been initialized.

mirw:
Looks good.  Some comments inline - most notably about not putting all the statistics definitions in the top-level module.

One other comment (which you're already thinking about) is that I think we need to figure out how to avoid netsnmp polluting our namespace - it does `#define READ`, which is a very unfriendly thing to do - we already this with the websockets code that used this for an enum value (and which netsnmp's behavior broke).

sprout/connection_pool.cpp
rkd-msw: Yes, I think so - but I think this race condition existed before, and only happens in Bono (which isn't going to be used in production, and has relatively few threads).

I am nervous that this could be hit in other IPCountTables (like in HttpConnection), but can't see a good way to avoid it without locking.

sprout/main.cpp
rkd-msw: I'm not convinced we need to - it's never been a stat that we've publicised. I suggest we just leave it as is, and if there's demand for it in future (or we feel it would be useful) we raise a user story (or just track it out of bugfix).

debian/control
rkd-msw: Added in a separate pull request.

include/hssconnection.h
rkd-msw: I don't feel that strongly - having a HSSConnectionStats class that just exists to wrap these five objects feels a bit clunky. I'd be happy to have HSSConnection construct these tables and not have them passed in - that means you can't have two at once, but I don't think we ever would want to. Thoughts?

sprout/connection_pool.cpp
rkd-msw: Given that HttpConnection used a lock previously (https://github.com/Metaswitch/cpp-common/pull/315/files#diff-79f2570d8cea76ddbb903d352c21713fL919), I'm just going to use a similar lock (certainly in HttpConnection, probably in here too, where Bono only has one or two worker threads and so lock contention isn't a big concern).

rkd-msw: In fact, we only call `decrement_connection_count` or `increment_connection_count` while holding `_tp_hash_lock`, so we're fine here.

sprout/connection_pool.cpp
rkd-msw: It would wrap to UINT_MAX, but we should have sufficient locking in place to avoid hitting this race condition.

debian/control
rkd-msw: I've removed it from sprout-base.

sprout/main.cpp
mirw: Why do we need this for Homestead and not for Ralf?

include/hssconnection.h
mirw: Don't feel too strongly about HSSConnectionStats.  I don't think we should mandate HSSConnection is a singleton without making it a proper singelton (i.e. `getInstance()`).

mirw:
A few further comments inline.

sprout/main.cpp
rkd-msw: I've added a comment to sto857 (the Ralf stats story) suggesting that it should cover Ralf-related stats in other components.

PR #1126, opened 2015-07-06
Allows server to be a host name [reviewer Rob]

alango:


sprout/enumservice.cpp
```diff
@@ -288,9 +292,22 @@ DNSEnumService::DNSEnumService(const std::vector<std::string>& dns_servers,
     {
       dns_server_addr.af = AF_INET6;
     }
+    else if ((ret = getaddrinfo(server->c_str(), NULL, NULL, &res)) == 0)
```
rkd-msw: It doesn't look like you're using `ret` anywhere else, so you could just remove it and check the return value of getaddrinfo directly(`else if (getaddrinfo(server->c_str(), NULL, NULL, &res) == 0)`) - I think that's slightly more readable.

rkd-msw:
Looks good to me - only one comment (about `ret` being unnecessary) and then you're good to merge.

PR #1127, opened 2015-07-10
Registration stats added [reviewer: Rob]

alango:


include/registration_utils.h
```diff
@@ -65,12 +66,14 @@ void register_with_application_servers(Ifcs& ifcs,
                                        int expires,
                                        bool is_initial_registration,
                                        const std::string& served_user,
-                                       SAS::TrailId trail);
+                                       SAS::TrailId trail,
```
rkd-msw: We usually keep the SAS::TrailId as the last parameter - can you switch these two parameters around?

sprout/authentication.cpp
```diff
@@ -752,6 +767,20 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
 
     TRC_ERROR("Authentication failed, %s", error_msg.c_str());
 
+    if (av != NULL)
+    {
+      if (av->HasMember("digest"))
+      {
+        auth_stats_tables->sip_digest_auth_tbl->increment_attempts();
+        auth_stats_tables->sip_digest_auth_tbl->increment_successes();
```
rkd-msw: It looks like you're calling `increment_successes` here - but isn't this the failure case?

sprout/authentication.cpp
```diff
@@ -605,6 +610,17 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
       SAS::Event event(trail, SASEvent::AUTHENTICATION_SUCCESS, 0);
       SAS::report_event(event);
 
+      if (av->HasMember("digest"))
+      {
+        auth_stats_tables->sip_digest_auth_tbl->increment_attempts();
```
rkd-msw: It would be better to call `increment_attempts` as early as possible, rather than alongside the call to `increment_successes`. This means that:
* if there are bugs which mean that we just don't handle a REGISTER (e.g. we return from this function too early) it will be possible to spot that.
* more generally, it makes the "Attempts" column a bit more meaningful - if it's just defined as the sum of the "Successes" and "Failure" column, then it's not really adding anything.

sprout/authentication.cpp
```diff
@@ -752,6 +767,20 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
 
     TRC_ERROR("Authentication failed, %s", error_msg.c_str());
 
+    if (av != NULL)
```
rkd-msw: I'm a bit nervous about this - it means we don't update a statistic if we failed because `av` was NULL.

Fortunately, I think there's a better way to check this - you have the `auth_hdr` variable, which is a [pjsip_authorization_hdr](http://www.pjsip.org/pjsip/docs/html/structpjsip__authorization__hdr.htm) and contains a [pjsip_digest_credential](http://www.pjsip.org/pjsip/docs/html/structpjsip__digest__credential.htm).

That has an "algorithm" field, which can be "AKAv1-MD5" for AKA auth, or just "MD5" for SIP Digest auth. If you use [pj_strcmp2](http://www.pjsip.org/pjlib/docs/html/group__PJ__PSTR.htm#ga3321b5cee71785a7b450ad29afe0895b) to check that field, you can tell whether an authentication request uses AKA without relying on the AV - which lets you increment the right 'attempts' column before retrieving the AV, and to increment the right 'failures' column even if you fail to retrieve the AV.

sprout/registration_utils.cpp
```diff
@@ -197,6 +211,45 @@ static void send_register_cb(void* token, pjsip_event *event)
 
     third_party_register_failed(tsxdata->public_id, tsxdata->trail);
   }
+  
+  if (tsx->status_code == 200)
+  {  
+    if (tsxdata->is_initial_registration)
+    {
+      third_party_reg_stats_tables->init_reg_tbl->increment_attempts();
```
rkd-msw: Again, it would be good to increment the attempts as early as possible - it might make sense to increment the attempts when we send a REGISTER to the AS, rather than when we receive a response.

sprout/sprout_base.mk
```diff
@@ -81,6 +81,7 @@ TARGET_SOURCES := logger.cpp \
                   snmp_accumulator_table.cpp \
                   snmp_counter_table.cpp \
                   snmp_ip_count_table.cpp \
+				  snmp_success_fail_count_table.cpp \
```
rkd-msw: It looks like you've (or rather, vim has) used tabs here rather than spaces.

sprout/ut/registrar_test.cpp
```diff
@@ -1108,6 +1131,208 @@ TEST_F(RegistrarTest, AppServersInitialRegistrationFailure)
...
+  ASSERT_NO_FATAL_FAILURE(r1.matches(out));
+  EXPECT_EQ(NULL, out->body);
+
+  tpAS.expect_target(current_txdata(), false);
+  inject_msg(respond_to_current_txdata(200));
+
+  free_txdata();
+}
+
+//TEST_F(RegistrarTest, AppServersReRegistrationFailure)
```
rkd-msw: Better to delete this code rather than commenting it out. (If you think you might want it later, make an interim git commit before deleting it.)

sprout/ut/registrar_test.cpp
```diff
@@ -1108,6 +1131,208 @@ TEST_F(RegistrarTest, AppServersInitialRegistrationFailure)
...
+  // We add two identical IP addresses so that we hit the retry behaviour, 
+  // but we don't have to worry about which IP address is selected first. 
+  add_host_mapping("app-server", "1.2.3.4, 1.2.3.4");
+
+  TransportFlow tpAS(TransportFlow::Protocol::UDP, stack_data.scscf_port, "1.2.3.4", 56789);
+
+  SCOPED_TRACE("REGISTER (1)");
+  Message msg;
+  msg._expires = "Expires: 0";
+  msg._contact_params = ";+sip.ice;reg-id=1";
```
rkd-msw: You don't need this instance of `msg._contact_params` - you're setting it again a couple of lines later.

sprout/ut/registrar_test.cpp
```diff
@@ -1127,6 +1352,76 @@ TEST_F(RegistrarTest, AppServersInitialRegistrationFailure)
...
+  SCOPED_TRACE("REGISTER (injected)");
+  ASSERT_EQ(1, txdata_count());
+  SCOPED_TRACE("REGISTER (200 OK)");
+  pjsip_msg* out = current_txdata()->msg;
+  EXPECT_EQ(200, out->line.status.code);
+  EXPECT_EQ("OK", str_pj(out->line.status.reason));
+  free_txdata();
+
+  SCOPED_TRACE("REGISTER (forwarded)");
+  // REGISTER passed on to AS
```
rkd-msw: I think the "REGISTER passed on to AS" and "REGISTER not passed on to AS" comments are the wrong way round.

sprout/ut/fakesnmp.hpp
```diff
@@ -84,9 +85,26 @@ class FakeIPCountTable : public IPCountTable
   void remove(std::string key) {};
 };
 
+class FakeSuccessFailCountTable : public SuccessFailCountTable
```
rkd-msw: I wonder if it's worth fleshing this class out a bit more:
* adding attempts/successes/failure member variables
* having the increment methods increment those members
* checking those member variables during the tests (e.g. `ASSERT_EQ(1u, FAKE_SIP_DIGEST_AUTH_TABLE._failure_count)`)
* adding a `reset()` method to reset all those counts to 0, which can be called after every test

This would catch some of the potential bugs here (e.g. where you've called `increment_successes` instead of `increment_failures` above), and save time on the live test phase, so I think it's worth doing.

sprout/registrar.cpp
```diff
@@ -495,9 +501,51 @@ void process_register_request(pjsip_rx_data* rdata)
   int st_code = PJSIP_SC_OK;
   SAS::TrailId trail = get_trail(rdata);
 
+  // Moved this block of code further up so that the correct stats tables can be
```
rkd-msw: It's better to structure comments for the future reader - in six months someone won't need to know that this block of code has been *moved*, but they will want to know why it is where it is. Maybe something like "We loop through the Contact headers as soon as possible, so that we can get the expiry time and determine whether this is a deregistration."

sprout/registrar.cpp
```diff
@@ -514,6 +562,16 @@ void process_register_request(pjsip_rx_data* rdata)
                                NULL,
                                NULL,
                                NULL);
+    if (expiry == 0)
+    {
+      reg_stats_tables->de_reg_tbl->increment_attempts();
+      reg_stats_tables->de_reg_tbl->increment_failures();
+    }
+    else // If not a de-register request, then count as an initial register attempt.
```
rkd-msw: It's worth explaining why this is valid in the comments - it's reasonably clear from the code that you're treating anything that isn't a de-register as an initial registration, but a comment can add value by explaining why (because with a invalid URI, this can't be a re-registration, as this user can't have successfully made an initial registration).

sprout/registrar.cpp
```diff
@@ -764,6 +801,23 @@ void process_register_request(pjsip_rx_data* rdata)
                                NULL);
     delete acr;
     delete aor_data;
+
+    if (is_initial_registration)
+    {
+      reg_stats_tables->init_reg_tbl->increment_attempts();
```
rkd-msw: Again, if you increment the attempts as soon as you know whether this is an initial registration/re-registration/de-registration, then this column is more useful, and you don't need to repeat `increment_attempts` in multiple places.

rkd-msw:
Generally looks pretty good - I had a few comments, but I don't think there's anything that should be much work to change.

rkd-msw:
Looks good - only a couple of comments.

PR #1131, opened 2015-07-12
[Reviewer: Alex] Challenge INVITEs in Sprout

rkd-msw:
This:
* refactors the authentication module so it can apply to non-REGISTER messages
* makes the authentication module transaction-stateful (so 407 responses to an INVITE can be ACKed)

I need to make a corresponding sas-resources change.


AlexHockey:
As discussed, this PR is out of date with our current thinking so passing back to you to update or close.

rkd-msw:
Our current thinking changed again - we're now keeping the code in the authentication module and doing an external network hop to ensure it's triggered.

sprout/authentication.cpp
```diff
@@ -110,6 +110,34 @@ static AnalyticsLogger* analytics;
...
+  {
+    credentials = &((pjsip_authorization_hdr*)auth_hdr)->credential.digest;
+  }
+  else
+  {
+    credentials = &((pjsip_proxy_authorization_hdr*)auth_hdr)->credential.digest;
+  }
+
+  return credentials;
+}
```
AlexHockey: I think you've written this function the way it is to minimize code duplication but you've made it less type safe. I think it would be better and easier to follow as:

```
if (method == REGISTER) {
    header = find(PJSIP_H_AUTHORIZATION);
    credentials = header ? header->credential.digest : NULL; 
} else {
    header = find(PJSIP_H_PROXY_AUTHORIZATION);
    credentials = header ? header->credential.digest : NULL; 
}
```

rkd-msw: I'm actually now pretty confident that `pjsip_proxy_authorization_hdr` and `pjsip_authorization_hdr` are typedefs, so I might rework this based on that.

sprout/authentication.cpp
```diff
@@ -499,95 +522,141 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
...
+               (auth_hdr->credential.digest.response.slen == 0))
+      {
+        // The integrity protected indicator is include and set to yes.  This
+        // indicates that AKA authentication is in use and the REGISTER was
+        // received on an integrity protected channel, so we will let the
+        // request through if there is no challenge response, but must check
+        // the challenge response if included.
+        TRC_INFO("AKA authenticated request integrity protected by edge proxy");
+
+        SAS::Event event(trail, SASEvent::AUTHENTICATION_NOT_NEEDED_INTEGRITY_PROTECTED, 0);
```
AlexHockey: Use a different SAS unique ID

sprout/authentication.cpp
AlexHockey: What if they stop being typedefs in the future?

rkd-msw: That feels very unlikely - their ABNF is the same ("Proxy-Authorization" HCOLON credentials, "Authorization" HCOLON credentials).

sprout/authentication.cpp
```diff
@@ -791,7 +857,24 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
...
+    pjsip_endpt_send_response2(stack_data.endpt, rdata, tdata, NULL, NULL);
+    // LCOV_EXCL_STOP
+  }
+  else
+  {
+    // Let the tsx know about the original message
+    pjsip_tsx_recv_msg(tsx, rdata);
+    // Send our response in this transaction
+    pjsip_tsx_send_msg(tsx, tdata);
+  }
```
AlexHockey: Why is this change needed?

sprout/pjutils.cpp
```diff
@@ -449,25 +449,53 @@ void PJUtils::add_integrity_protected_indication(pjsip_tx_data* tdata, Integrity
...
-  if ((auth_hdr != NULL) &&
-      (auth_hdr->credential.digest.username.slen != 0))
+  if (rdata->msg_info.msg->line.req.method.id == PJSIP_REGISTER_METHOD)
   {
-    // private user identity is supplied in the Authorization header so use it.
-    impi_out = PJUtils::pj_str_to_string(&auth_hdr->credential.digest.username);
-    TRC_DEBUG("Private identity from authorization header = %s", impi_out.c_str());
+    // Check to see if the request has an explicit IMPI in the Authorization header
+    pjsip_authorization_hdr* auth_hdr = (pjsip_authorization_hdr*)
+      pjsip_msg_find_hdr(rdata->msg_info.msg, PJSIP_H_AUTHORIZATION, NULL);
```
AlexHockey: Clearer to get this header just before you look at it (rather than looking at the To header in between).

sprout/pjutils.cpp
```diff
@@ -449,25 +449,53 @@ void PJUtils::add_integrity_protected_indication(pjsip_tx_data* tdata, Integrity
...
-    // public user identity by stripping the sip: prefix.
-    impi_out = PJUtils::default_private_id_from_uri(to_uri);
-    TRC_DEBUG("Private identity defaulted from public identity = %s", impi_out.c_str());
+    // Retrieve the IMPU for a non-REGISTER request by determining the originating served user, and
+    // the IMPI from the Proxy-Authorization header.
+
+    pjsip_uri* impu_uri = orig_served_user(rdata->msg_info.msg);
+    impu_out = PJUtils::public_id_from_uri(impu_uri);
+
+    // Check to see if the request has an explicit IMPI in the Authorization header
```
AlexHockey: Proxy-Authorization header, not Authorization header.

sprout/pjutils.cpp
```diff
@@ -449,25 +449,53 @@ void PJUtils::add_integrity_protected_indication(pjsip_tx_data* tdata, Integrity
...
+    {
+      impi_out = PJUtils::pj_str_to_string(&auth_hdr->credential.digest.username);
+      TRC_DEBUG("Private identity from authorization header = %s", impi_out.c_str());
+    }
+    else
+    {
+      // IMPI not supplied, so construct a default from the IMPU by stripping the sip: prefix.
+      impi_out = PJUtils::default_private_id_from_uri(impu_uri);
+      TRC_DEBUG("Private identity defaulted from public identity = %s", impi_out.c_str());
+    }
```
AlexHockey: Worth putting this if/else block in a subfunction?

sprout/authentication.cpp
AlexHockey: Ok. I would probably choose to rely on the compiler rather than that assumption.

Note that C++ will automatically cast between types that are defined to the same thing, but does not allow this if they don't. You can probably leverage that if you like.

sprout/ut/authentication_test.cpp
```diff
@@ -380,6 +418,44 @@ TEST_F(AuthenticationTest, NoAuthorizationNonReg)
...
+  RespMatcher(407).matches(tdata->msg);
+  free_txdata();
+
+  // ACK that response
+  AuthenticationMessage ack("ACK");
+  ack._cseq = 1;
+  inject_msg_direct(ack.get());
+ 
+  _hss_connection->delete_result("/impi/6505550001%40homedomain/av?impu=sip%3A6505550001%40homedomain");
+}
```
AlexHockey: It's worth testing that the authentication module can also accept a properly authenticated non-REGISTER, and rejects an non-REGISTER with bad credentials.

include/sproutsasevent.h
```diff
@@ -83,17 +83,18 @@ namespace SASEvent
...
+  const int AUTHENTICATION_NOT_NEEDED_PROXY_AUTHORIZATION = SPROUT_BASE + 0x000043;
+  const int AUTHENTICATION_NEEDED_PROXY_AUTHORIZATION = SPROUT_BASE + 0x000044;
+  const int AUTHENTICATION_CHALLENGE_AKA = SPROUT_BASE + 0x000045;
+  const int AUTHENTICATION_CHALLENGE_DIGEST = SPROUT_BASE + 0x000046;
+  const int AUTHENTICATION_FAILED_OVERLOAD = SPROUT_BASE + 0x000047;
+  const int AUTHENTICATION_FAILED_MALFORMED = SPROUT_BASE + 0x000048;
+  const int AUTHENTICATION_FAILED_NO_AV = SPROUT_BASE + 0x000049;
+  const int AUTHENTICATION_FAILED = SPROUT_BASE + 0x00004A;
+  const int AUTHENTICATION_SUCCESS = SPROUT_BASE + 0x00004B;
+  const int AUTHENTICATION_TIMER_POP_IGNORED = SPROUT_BASE + 0x00004C;
```
AlexHockey: Steve has asked us to not renumber SAS events for cosmetic reasons as it makes his job harder. I suggest we don't change these events unless you think the maintainability gain is really significant.

AlexHockey:
Code looks good. You have live tested as well, correct?

sprout/authentication.cpp
rkd-msw: If we're challenging an INVITE with a 407, we need to be transaction-stateful to absorb the ACK. (Right?)

AlexHockey: Ah, OK. Definitely worth a comment!

rkd-msw:
I have live-tested, and I'm going to check that test script in.

PR #1136, opened 2015-07-17
[Reviewer: Rob] Log the to/from URI as SIP URI markers

### LINKED ISSUE #924: Cannot find traces in SAS by doing a "SIP URI" search.
SAS allows you to search for calls and other traces by entering a SIP URI that is involved in the trace. If you set up a Project Clearwater deployment (with Bono as the P-CSCF) this does not work. 

I think this is simply because we never log any of the markers specified here: https://github.com/Metaswitch/sas-client/blob/master/include/sas.h#L89

AlexHockey:
Rob, please could you review this PR that fixes #924. Although we are logging some new markers, the perf impact should be negligible (as they are still much smaller than the SIP message itself).

Tested live, by checking a call could be found by searching on both the calling and called URIs.

### DIFF
--- a/include/pjutils.h
+++ b/include/pjutils.h
@@ -271,6 +271,7 @@ pj_bool_t is_user_global(const std::string& user);
 pj_bool_t is_user_global(const pj_str_t& user);
 
 std::string remove_visual_separators(const std::string& user);
+std::string remove_visual_separators(const pj_str_t& number);
 
 pj_bool_t is_user_numeric(const std::string& user);
 pj_bool_t is_user_numeric(const pj_str_t& user);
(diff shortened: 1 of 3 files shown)


rkd-msw:
Looks good. Have you tested with non-numeric URIs (e.g. by making a call to Alex.Hockey@domain)?

AlexHockey:
I have now. This highlighted that we are logging all usernames as DNs which makes the UI look a bit rubbish. I'm going to update the PR to fix this.

AlexHockey:
I have updated the PR to make URIs with non-numeric usernames display properly. I've done this by putting the whole URI in the DN parameter if the username is not numeric. Please can you re-review?

I've tested calls and registrations involving URIs containing DNs and URIs not containing DNs. All displayed sensibly in SAS.

PR #1137, opened 2015-07-18
Reviewer Matt - Don't have specific rules for .depends files

bossmc:
This is the same Makefile change that I made to cedar that you reviewed earlier this week.  Basically, we build the `.d` file alongside the `.o` file rather than up-front.

Tested by running make various times, touching random files between some runs.  If you're happy with this, I'll apply the same change to other users of `platform.mk`.


mirw:
Looks plausible.  Presumably it works?

Issue #1121, opened 2015-07-01
SIP URIs with just npdi (and no other routing information) can loop pointlessly


rkd-msw:
This is the case where:
* a SIP URI comes in (to example.com) as `sip:1234@example.com`
* ENUM rewrites it to `sip:1234;npdi@example.com` at the I-CSCF
* It immediately gets routed to the BGCF, which routes it back to example.com (I-CSCF)
* ENUM still rewrites it to `sip:1234;npdi@example.com`, and routes it to the BGCF
* etc. etc. until Max-Forwards decrements enough

I'm not sure if rewriting `sip:1234@example.com` to `sip:1234;npdi@example.com` is ever sensible, but we should handle this configuration better, not loop, and have the subscriber handled by the S-CSCF at some point.

rkd-msw:
Discussed with @eleanor-merry  and @bossmc - fix is to not treat `npdi` as "routing information" in the TS 24.229 sense.

bossmc:
Marking high priority since it blocks a high-priority issue.

rkd-msw:
Fixed by #1156.

Issue #1120, opened 2015-07-01
SAS logging around ENUM is confusing (particularly when 'npdi' is involved)


rkd-msw:
While investigating https://github.com/Metaswitch/sprout/issues/1096, I found the SAS logs pretty useless - in particular, there was always a high-level log made indicating we were changing the Request-URI, even if we subsequently checked the `npdi` flag and decided not to use the result.

Having discussed with @eleanor-merry , we should:
* have a high-level log indicating whether we're going to do ENUM and why (e.g. enforce_user_phone is set)
* move most existing ENUM logs to be protocol-flow-level
* have a high-level log at the end saying whether we changed the Request-URI, what to, and why (e.g. override_npdi was disabled)

rkd-msw:
Fixed by https://github.com/Metaswitch/clearwater-sas-resources/commit/9ce5c167bbb721de7f0f2037e89b0be4ad85bc9d#diff-d27b7cf4f49cf128e646ccab9a8dd245.

PR #1138, opened 2015-07-19
[Reviewer: Ellie] Rework number portability checking

rkd-msw:
Can you give this a very quick first pass? This is meant to fix https://github.com/Metaswitch/sprout/issues/1121 and makes a start on https://github.com/Metaswitch/sprout/issues/1120.

It:
* stops treating just `npdi` as "routing information"
* reworks the logic to be a bit clearer
* adds a specific SAS log at each outcome (though I haven't added the corresponding SAS logs - this is just illustrative)

If this is OK, I'll add the SAS logs, make similar I-CSCF and BGCF changes, and test.

### DIFF
--- a/sprout/scscfsproutlet.cpp
+++ b/sprout/scscfsproutlet.cpp
@@ -1443,48 +1443,59 @@ bool SCSCFSproutletTsx::uri_translation_and_route(pjsip_msg* req)
       // URI and substitute it in to the request.
       pjsip_uri* new_uri = (pjsip_uri*)PJUtils::uri_from_string(new_uri_str,
                                                                 get_pool(req));
-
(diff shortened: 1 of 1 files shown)


sprout/scscfsproutlet.cpp
```diff
@@ -1502,6 +1513,9 @@ bool SCSCFSproutletTsx::uri_translation_and_route(pjsip_msg* req)
     }
     else if (PJUtils::is_uri_phone_number(uri))
```
eleanor-merry: As discussed, this else if branch isn't catching all the right cases

eleanor-merry:
Code looks code. My main comment is that I think this needs a bunch more commenting

PR #1141, opened 2015-07-22
[Reviewer: Rob] Be consistent with other devices when logging SIP URIs and DNs

AlexHockey:
This updates #1136 so that Clearwater is consistent with other the way other products log URIs and DNs on calls, registrations and subscriptions. Please can you review.

I have tested by doing the following. In each case I checked that the trace could be found by URI and DN (if appropriate) and that the results look sensible in SAS. 

* Register with a DN-like URI.
* Register with a non-DN-like URI.
* Call a DN-like URI
* Call a non-DN-like URI.

I will run through these tests on the staging system as well once the PR is merged.


rkd-msw:
Looks good - no comments.

PR #1144, opened 2015-07-24
[reviewer: Rob] Non register stats

alango:


sprout/authentication.cpp
```diff
@@ -669,17 +673,22 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
   if ((credentials != NULL) &&
       (credentials->response.slen != 0))
   {
-    if (is_register)
+    if (!is_register)
+    {
+      auth_stats_table = non_register_auth_table;
```
rkd-msw: Worth a comment here to say that if a non-REGISTER request is challenged, it must be SIP Digest, so we only need the one table?

sprout/authentication.cpp
```diff
@@ -955,7 +937,8 @@ pj_status_t init_authentication(const std::string& realm_name,
                                 ChronosConnection* chronos_connection,
                                 ACRFactory* rfacr_factory,
                                 AnalyticsLogger* analytics_logger,
-                                SNMP::AuthenticationStatsTables* auth_stats_tbls)
+                                SNMP::AuthenticationStatsTables* auth_stats_tbls,
+                                SNMP::SuccessFailCountTable* non_reg_auth_table)
```
rkd-msw: Should this be a member of AuthenticationStatsTables for consistency? (I don't mind that much, though.)

rkd-msw:
Looks good, and I think simplifying the `if` statements by just choosing a table to update at the start has worked well - this change actually removes more lines of code than it adds.

PR #1145, opened 2015-07-27
[Reviewer: Rob] Finish pbx sip signaling

AlexHockey:
Rob, please can you review this PR to finish of SIP signalling for PBXs. This contains two changes:

* A new config option to control when we challenge requests. 
* The S-CSCF will now honor routes preloaded by the AS, and treat this as the "end of terminating processing" for the purposes of record-routing. 

The first change has been tested in Uts and live (with your SIPp script). The second change has only been UTed so far. I will be writing a script to live test it tomorrow. I will also check that the ACRs look sensible when an AS preloads a route (this is hard to do in UT).


sprout/authentication.cpp
```diff
@@ -618,26 +620,41 @@ static pj_bool_t needs_authentication(pjsip_rx_data* rdata, SAS::TrailId trail)
...
+      else
+      {
+        // No Proxy-Authorization header - this indicates the P-CSCF trusts this
+        // message so we don't need to perform further authentication.
+        SAS::Event event(trail, SASEvent::AUTHENTICATION_NOT_NEEDED_PROXY_AUTHORIZATION, 0);
+        SAS::report_event(event);
+        return PJ_FALSE;
+      }
     }
     else
```
rkd-msw: Do we really need this else branch? Isn't the compiler smart enough to see that we're covering both values of the enum?

sprout/scscfsproutlet.cpp
```diff
@@ -49,6 +49,9 @@
 #include "registration_utils.h"
 #include "scscfsproutlet.h"
 
+// Constant indicating there is no served user fro a request.
```
rkd-msw: fro -> for

sprout/ut/scscf_test.cpp
```diff
@@ -4179,7 +4164,7 @@ TEST_F(SCSCFTest, Cdiv)
   msg._via = "10.99.88.11:12345;transport=TCP";
   msg._to = "6505551234@homedomain";
   msg._todomain = "";
-  msg._route = "Route: <sip:homedomain;orig>";
```
rkd-msw: Why's this changed?

rkd-msw:
Looks good. Only trivial comments.

sprout/authentication.cpp
AlexHockey: The reason I coded it like this was so that if we extend the enum we will fall into the else branch. Given that is a straight up logic bug, I think it would be better to assert in the else branch. 

Ideally g++ would spot if we hadn't consumed all the possible values of an enum, but it's not smart enough to do this.

sprout/ut/scscf_test.cpp
AlexHockey: This tests orig-cdiv processing. It does this by first doing a completely unnecessary set of orig processing. This cost me time figuring our what the test was doing, so I tidied it up.

PR #1149, opened 2015-07-29
[reviewer: Rob] SIP transaction stats

alango:


sprout/bgcfsproutlet.cpp
```diff
@@ -61,12 +61,18 @@ BGCFSproutlet::BGCFSproutlet(int port,
   _user_phone(user_phone),
   _override_npdi(override_npdi)
 {
+  _incoming_sip_transactions_tbl = SNMP::SuccessFailCountByRequestTypeTable::create("bgscf_incoming_sip_transactions",
```
rkd-msw: Typo: bgscf -> bgcf

sprout/ut/sproutletproxy_test.cpp
```diff
@@ -485,17 +485,21 @@ class SproutletProxyTest : public SipTest
 
   static void TearDownTestCase()
   {
+    //printf("Starting tear down\n");
```
rkd-msw: Debugging printfs.

sprout/sproutletproxy.cpp
```diff
@@ -999,6 +1002,16 @@ SproutletWrapper::SproutletWrapper(SproutletProxy* proxy,
     _sproutlet_tsx = new SproutletTsx(this);
   }
 
+  if ((_sproutlet != NULL) &&
```
rkd-msw: Worth a comment (here, and in similar blocks you've added) to explain what this code is for in general - e.g. that we're updating our SNMP statistics?

modules/pjsip
```diff
@@ -1 +1 @@
-Subproject commit 32e0970a15abea432265c370073e783c3914f64d
```
rkd-msw: You should undo this change before merging. More generally, it's probably a good idea to only check in certain directories (`git add -u include; git add -u sprout`).

rkd-msw:
Looks good - as a general rule, it's probably worth sanity-checking these (with `git diff --stat` and/or `git diff`) before sending for review - that can pick up any files inadvertently changed (modules/pjsip) or obvious things (like commented-out printf statements).

PR #1159, opened 2015-07-31
[reviewer:Rob] changed order of snmp commands

alango:


rkd-msw:
Looks good - maybe worth a comment to say that this must happen after all the tables are registered, so that nobody reintroduces the same bug in future?

PR #1164, opened 2015-08-06
[Reviewer: Rob] Get output from upload_json

eleanor-merry:
Matching fix to https://github.com/Metaswitch/clearwater-etcd/pull/153


rkd-msw:
+1

PR #1166, opened 2015-08-07
Add create_request method

mbt-msw:
This allows creation of blank REQUESTs to be used by sproutlets when making downstream requests, when re-using existing requests is not appropriate.

The caller is responsible for filling in all the mandatory SIP headers before sending the message.


bossmc:
Nice, LGTM.

PR #1171, opened 2015-08-10
[Reviewer: Matt] Creation of overload statistics

TygerBurning:
Creating SNMP statistics and passing them to load monitor

Also adds some load monitor tests. Created OID's seem out of synchronisation because the MIB file says that 18-26 are already taken


sprout/ut/load_monitor_test.cpp
```diff
@@ -41,26 +41,47 @@
...
 };
 
+// Pushes the time until the next period
+// The time is within the first second of the next period
+void LoadMonitorTest::jump_to_next_periodstart()
+{
+  struct timespec now;
+  clock_gettime(CLOCK_REALTIME_COARSE, &now);
+  uint64_t ms_since_epoch = (now.tv_sec * 1000) + (now.tv_nsec / 1000000);
+  cwtest_advance_time_ms(300000 - (ms_since_epoch % 300000));
```
mirw: Why 300000?  Is there a constant we should be using?

mirw: Worth a comment as to how this works?

sprout/ut/load_monitor_test.cpp
```diff
@@ -181,3 +202,25 @@ TEST_F(TokenBucketTest, GetToken)
...
 
 }
+
+
+TEST_F(LoadMonitorTest, CorrectStatistics)
+{
+  // Scalars should report values from last update, not current values.
+  // Initialisation should count as an update though
+  EXPECT_EQ(target_latency.value, 100000);
+  EXPECT_EQ(smoothed_latency.value, 100000);
```
mirw: Explain where these values come from.

mirw:
Generally looks good - just a few commenting-related mark-ups.

mirw:
Looks good.

TygerBurning:
After Chronos comment also deleted tables after use

mirw:
Fine.

PR #1172, opened 2015-08-12
Reviewer AMC: Edit Chronos connection alarm

graemerobertson:
This better? If you're happy with this, I'll make the equivalent change in Ralf too.


bossmc:
Much better.

PR #1181, opened 2015-08-24
[Reviewer: Andy] Pull in max_session_expires

### LINKED ISSUE #1180: max_session_expires isn't plumbed into Sprout's init.d file
It should be.

eleanor-merry:
Passes the max_session_expires value from /etc/clearwater/config to Sprout. Tested live. 

Fixes #1180

### DIFF
--- a/sprout-base.root/etc/init.d/sprout
+++ b/sprout-base.root/etc/init.d/sprout
@@ -179,6 +179,7 @@ get_daemon_args()
         [ -z "$min_token_rate" ] || min_token_rate_arg="--min-token-rate=$min_token_rate"
         [ -z "$exception_max_ttl" ] || exception_max_ttl_arg="--exception-max-ttl=$exception_max_ttl"
         [ -z "$signaling_namespace" ] || namespace_prefix="ip netns exec $signaling_namespace"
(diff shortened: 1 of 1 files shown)


sprout-base.root/etc/init.d/sprout
```diff
@@ -179,6 +179,7 @@ get_daemon_args()
         [ -z "$min_token_rate" ] || min_token_rate_arg="--min-token-rate=$min_token_rate"
         [ -z "$exception_max_ttl" ] || exception_max_ttl_arg="--exception-max-ttl=$exception_max_ttl"
         [ -z "$signaling_namespace" ] || namespace_prefix="ip netns exec $signaling_namespace"
+        [ -z "$max_session_expires" ] || max_session_expires_arg="--max-session-expires=$max_session_expires"
```
bossmc: It's weird that `namespace_prefix` is now in the middle of this list.  Also, shouldn't that have been replaced with `/.../run-in-signaling-namespace`?

bossmc:
One comment, assuming you've tested that this actually works (e.g. that the command line param has the right name and is wired into the right bits of the code).

PR #1186, opened 2015-09-08
[Reviewer: Rob] Make handling of 3rd party registration stats more defensive

AlexHockey:
I hit a crash in UT caused by the registration_utils.cpp::third_party_reg_stats_tables pointer being NULL. This can happen if you call `remove_bindings` before `(de)register_with_application_servers`. I have fixed this by making `registration_utils.cpp` more defensive and checking its pointer. 

I think there is still a more minor issue where 3rd party reg stats will be wrong if we call `remove_bindings` first. I will raise an issue for this (it's really low urgency). 

Tested in UT.


rkd-msw:
Looks good.

PR #1187, opened 2015-09-11
[reviewer: Rob] Invite cancellation stats

alango:


include/scscfsproutlet.h
```diff
@@ -361,6 +363,11 @@ class SCSCFSproutletTsx : public SproutletTsx
   /// us accidentally record routing twice.
   bool _record_routed;
 
+  /// Track request type and whether a 1xx response has been seen so that the
+  /// correct stats can be updated.
+  SNMP::SIPRequestTypes _req_type;
```
rkd-msw: Would `pjsip_method_e` be a better type for this (http://www.pjsip.org/pjsip/docs/html/group__PJSIP__MSG__METHOD.htm#ga66b57e1b5645d2ee843141a0e657b0d1)? That's just an int you can retrieve from a pjsip_msg, so requires less processing (e.g. no string comparisons).

sprout/scscfsproutlet.cpp
```diff
@@ -665,6 +681,13 @@ void SCSCFSproutletTsx::on_tx_response(pjsip_msg* rsp)
 void SCSCFSproutletTsx::on_rx_cancel(int status_code, pjsip_msg* cancel_req)
 {
   TRC_INFO("S-CSCF received CANCEL");
+   
+  if (_req_type == SNMP::SIPRequestTypes::INVITE)
+  // If an INVITE is being cancelled, then update INVITE cancellation stats.
+  {
+    if (_seen_1xx) { _scscf->_invites_cancelled_after_1xx_tbl->increment(); }
```
rkd-msw: Braces etc. should be on a new line here - I don't think this is short enough that departing from our usual style makes it more readable.

rkd-msw:
Looks good.

PR #1188, opened 2015-09-11
[Reviewer: Ellie] Add script for creating s-cscf.json

rkd-msw:
I'd like to streamline our manual install docs a bit. One thing that isn't currently very nice is the s-cscf.json stuff, where we basically tell users what file to put into place.

This script generates a sensible default, so I can basically make that section of http://clearwater.readthedocs.org/en/stable/Manual_Install/index.html#provide-shared-configuration be:

```
sudo /usr/share/clearwater/bin/create-default-icscf-config
sudo /usr/share/clearwater/clearwater-config-manager/scripts/upload_scscf_json
```

+1, -1?


eleanor-merry:
Well, there's no copyright statement so I instantly hate it. 
Should we create this anyway (like we have default versions of the bgcf and enum files)?

rkd-msw:
We discussed - `upload_scscf_json` should create this (as `upload_bgcf_json` does).

rkd-msw:
Also, then I don't have to create a new file or add copyright statements, nailed it.

rkd-msw:
Fixed it to create a usable /etc/clearwater/sample/s-cscf.json which https://github.com/Metaswitch/sprout/blob/dev/sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json#L60 will put into place.

debian/sprout-icscf.postinst
```diff
@@ -53,6 +53,18 @@ set -e
 
 case "$1" in
     configure)
+        cat > /etc/clearwater/sample/s-cscf.json << EOF
+{
+   "s-cscfs" : [
+       {   "server" : "sip:${sprout_hostname}:5054;transport=TCP",
```
eleanor-merry: What if sprout_hostname hasn't been set up yet?

rkd-msw:
We discussed, and I think we concluded:
- this has to be a clearwater-infrastructure script to allow for the fact that shared_config won't exist at install time
- upload_scscf_json should fail immediately if neither /etc/clearwater/s-cscf.json nor /etc/clearwater/sample/s-cscf.json exist

debian/sprout-icscf.postinst
eleanor-merry: This needs to have scscf. in the SIP URI now as well

rkd-msw:
@eleanor-merry - I think you're implementing the behaviour from https://github.com/Metaswitch/sprout/pull/1188#issuecomment-140219420 (or something similar) as part of some other work you're doing, so I'm going to close this PR.

PR #1189, opened 2015-09-14
Delay Tsx deletion when timers are pending

mbt-msw:
Currently, if a `SproutletTsx` leaves a timer running after completing (final response sent and no more outstanding downstream transactions), the `SproutletTsx` (and associated `SproutletProxy::UASTsx`) get deleted before the timer pops.  When the timer does pop, sprout segfaults.

The fix for this is to track pending timers and only delete the `SproutletWrapper` and `SproutletProxy::UASTsx` if all relevant pending timers have either popped or been cancelled.  We need two sets to track this: one for each `SproutletWrapper`, and one for the `SproutletProxy::UASTsx` as a whole, which should contain the union of all its `SproutletWrappers'` pending timer sets.

One important detail is that during cancellation, we need to make sure that we only remove timers from the sets of pending timers if the cancellation was successful.  If the cancellation fails, that means the timer has already popped on another thread, and is most likely waiting for the `UASTsx` context.


bossmc:
@AlexHockey - I think it makes sense for you to review this.

sprout/ut/sproutletproxy_test.cpp
```diff
@@ -463,6 +463,62 @@ class FakeSproutletTsxB2BUA : public SproutletTsx
   pjsip_method_e _method;
 };
 
+template <int T>
```
AlexHockey: Why is this class templated?

sprout/ut/sproutletproxy_test.cpp
```diff
@@ -1952,3 +2009,91 @@ TEST_F(SproutletProxyTest, LoopDetection)
   delete tp;
 }
 
+TEST_F(SproutletProxyTest, DelayAfterReponse)
```
AlexHockey: Could you trim this testcase down a bit? There's a few things you test that you don't really care about (Req URI, routing headers) and they distract from the timer stuff, which is the interesting bit.

sprout/sproutletproxy.cpp
```diff
@@ -399,10 +399,19 @@ bool SproutletProxy::schedule_timer(pj_timer_entry* tentry, int duration)
...
-  TRC_DEBUG("Cancelled Sproutlet timer, id = %ld", (TimerID)tentry);
+  pj_timer_heap_t* timer_heap = pjsip_endpt_get_timer_heap(_endpt);
+  if (pj_timer_heap_cancel(timer_heap, tentry) > 0)
+  {
+    TRC_DEBUG("Cancelled Sproutlet timer, id = %ld", (TimerID)tentry);
+    return true;
+  }
+  else
+  {
+    TRC_DEBUG("Unable to cancel Sproutlet timer, id = %ld", (TimerID)tentry);
```
AlexHockey: Please comment (or extend the log text) to make clear that this is not an error, and may mean the timer has already been freed.

include/sproutletproxy.h
```diff
@@ -341,6 +346,8 @@ class SproutletWrapper : public SproutletTsxHelper
   } ForkStatus;
   std::vector<ForkStatus> _forks;
   
+  std::set<TimerID> _pending_timers;
```
AlexHockey: Worth commenting that the sproutlet wrapper won't suicide unless this set is empty.

AlexHockey:
Hi Mark. This looks good, a few comments inline. 

In an ideal world I'd like to see another simpler UT - one where the sproutlet does not send the first request so there are no UAC transactions at all. As it stands I can imagine bugs where the presence of the UACTsx is helping to keep the sproutlet tsx alive. However I don't think its worth spending a lot of time adding this extra test, so only do it if it's trivially easy.

sprout/ut/sproutletproxy_test.cpp
mbt-msw: Symmetry with `FakeSproutletTsxDelayRedirect`, which is also templated with the number of seconds to delay.

AlexHockey: The entire file is template happy...

Ah I think I see why - its to avoid you having to create a `Sproutlet` as well as a `SproutletTsx`. Fine.

mbt-msw:
I've pushed updated UTs.  The old UT is renamed to "DelayAfterForward" and is trimmed down.  The new UT (DelayAfterReponse) no longer forwards the request, so is testing liveness purely by the timer.  In order to track it (since there are no SIP messages to look at any more), I added a tracking set to the class for the test case to query directly.

AlexHockey:
Lovely. We're currently in code freeze so I'll merge this when the code has unfrozen.

PR #1203, opened 2015-09-30
[Reviewer: Ellie] Don't treat the conversion of a SIP URI to the equivalent TEL URI as a retarget

rkd-msw:
TS 24.229 section 5.4.3.3 basically says "if an AS turns sip:+1234 into tel:+1234 or vice-versa, **don't** start orig-cdiv processing". We didn't implement that, which put us into an endless loop in some cases.

This change should fix that while still being efficient - in particular, I don't convert the std::string into a pjsip_uri in the mainline case, I only do so if the comparison fails but we have a tel: and non-tel: URI.

Fully UTed, and I've checked that my new UT fails without this fix.


bossmc:
BLUFF

rkd-msw:
Andy has pointed out that there's a broader fix needed (not treating a URI change as a retarget if the new URI is in the HSS-provided list of aliases), but I think we'd still want this fix (some devices treat sip:+1234@example.com;user=phone and tel:+1234 as equivalent even if tel:+1234 isn't listed as an alias in the HSS).

sprout/scscfsproutlet.cpp
```diff
@@ -764,6 +764,64 @@ void SCSCFSproutletTsx::retrieve_odi_and_sesscase(pjsip_msg* req)
 
 }
 
+pj_str_t get_userpart(pjsip_uri* uri)
```
eleanor-merry: Don't we already have a function that does this? user_from_uri?

sprout/scscfsproutlet.cpp
```diff
@@ -764,6 +764,64 @@ void SCSCFSproutletTsx::retrieve_odi_and_sesscase(pjsip_msg* req)
...
+bool is_retarget(std::string a, std::string b, pj_pool_t* pool)
+{
+  if (a == b)
+  {
+    // URIs match exactly - this is not a retarget
+    return false;
+  }
+  else
+  {
+    // Handle the case where an AS has converted sip:+1234@example.com and tel:+1234 - this is not considered a retarget that should trigger orig-cdiv
```
eleanor-merry: Comment lines too long

sprout/scscfsproutlet.cpp
```diff
@@ -764,6 +764,64 @@ void SCSCFSproutletTsx::retrieve_odi_and_sesscase(pjsip_msg* req)
...
+      TRC_DEBUG("One URI is a tel: URI and one is not - checking user-parts");
+
+      pjsip_uri* a_uri = PJUtils::uri_from_string(a, pool);
+      pjsip_uri* b_uri = PJUtils::uri_from_string(b, pool);
+      pj_str_t a_user = get_userpart(a_uri);
+      pj_str_t b_user = get_userpart(b_uri);
+      URIClass a_uri_class = URIClassifier::classify_uri(a_uri, false);
+      URIClass b_uri_class = URIClassifier::classify_uri(b_uri, false);
+ 
+      // Check that both URIs are ones we'd treat as phone numbers - for example, if user=phone is
```
eleanor-merry: Comment too long, userr and blank line

eleanor-merry:
LGTM, but I'd like to see a spec reference in the comments

rkd-msw:
I ate humble pie and did what Andy told me to. Worth you re-reviewing - the UT is identical, only the mainline code has changed, so there's only ~50 lines.

sprout/scscfsproutlet.cpp
```diff
@@ -745,6 +745,35 @@ void SCSCFSproutletTsx::retrieve_odi_and_sesscase(pjsip_msg* req)
...
+  {
+    // URIs match exactly - this is not a retarget
+    return false;
+  }
+  else if (std::find(aliases.begin(), aliases.end(), new_served_user) != aliases.end())
+  {
+    TRC_DEBUG("Application server has changed URI %s to the aliased URI %s - "
+              "not treating as a retarget, not invoking originating-cdiv processing",
+              old_served_user.c_str(),
+              new_served_user.c_str());
```
bossmc: SAS log?

sprout/scscfsproutlet.cpp
```diff
@@ -1514,6 +1543,22 @@ bool SCSCFSproutletTsx::get_associated_uris(std::string public_id,
...
 }
 
+/// Look up the aliases for the given public ID, using the cache if
+/// possible (and caching them and the iFC otherwise).
+/// The uris parameter is only filled in correctly if this function
+/// returns true,
+bool SCSCFSproutletTsx::get_aliases(std::string public_id,
+                                    std::vector<std::string>& aliases)
+{
+  bool success = get_data_from_hss(public_id);
```
bossmc: Is `get_data_from_hss()` cached?  We're going to look up the IFCs later and that probably also calls `get_data_from_hss()`.

Alternative solution, remember the alias's from the previous S-CSCF invocation (as we do when we remember the `_as_chain_link.served_user()`) maybe this means we'll do aliasing using slightly out of date information... does that matter?

rkd-msw: Yes, it's cached.

We potentially want to store all this stuff (iFCs, CCFs, associated URis and aliases) on the AsChain, both for efficiency (fewer Homestead hits) and for consistency (keeping the state at the start of the AS chain, so you don't get weird behaviour if a PPR happens partway through).

rkd-msw: ...but that's a separate pull request.

sprout/scscfsproutlet.cpp
```diff
@@ -1514,6 +1543,22 @@ bool SCSCFSproutletTsx::get_associated_uris(std::string public_id,
   return success;
 }
 
+/// Look up the aliases for the given public ID, using the cache if
+/// possible (and caching them and the iFC otherwise).
+/// The uris parameter is only filled in correctly if this function
```
eleanor-merry: Not uris

eleanor-merry:
LGTM

PR #1205, opened 2015-10-02
[Reviewer: Andy] Session expires fixes

AlexHockey:
Andy, please can you review this PR to implement session timers as per RFC 4028 (rather than our previous approximation to it). 

As proper handling of the session-timer header requires us to maintain state between the request and response I have refactored the logic out into a new class, `SessionExpiresHelper`, that is called to update requests and responses correctly. 

Previously we rejected requests that contained a min-SE higher than our target value, on the basis that it would mean refreshes weren't sent frequently enough. As discussed this logic is actually really incomplete. For one thing there is nothing to prevent establishment of a call where neither UA supports timer. Also downstream devices can just raise the min-SE without rejecting the request and the first time we hear about this is on the 200 OK, by which point it's too late to reject the request. As such I have removed the logic to reject requests with a too-large min-SE. now we set the Session-Expires to the min-SE and proxy the request on. 

Tested in UT.


include/session_expires_helper.h
```diff
@@ -0,0 +1,107 @@
...
+  /// @param pool  - The pool associated with the request.
+  /// @param trail - SAS trail ID.
+  void process_request(pjsip_msg* req,
+                       pj_pool_t* pool,
+                       SAS::TrailId trail);
+
+  /// Do session expiry processing on a response. This carries out the
+  /// processing required by the RFC.
+  ///
+  ///
```
bossmc: Extra empty line?

include/session_expires_helper.h
```diff
@@ -0,0 +1,107 @@
...
+  bool timer_supported(pjsip_msg* msg);
+
+  // The session expiry we would like to use for this dialog.
+  SessionInterval _target_se;
+
+  // Whether this transaction forms an initial request for the dialog. This
+  // affects what SAS logging we produce.
+  bool _initial_request;
+
+  // Whether the UAC support session timers.
```
bossmc: "supports"?

sprout/session_expires_helper.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include "constants.h"
+#include "custom_headers.h"
+#include "sproutsasevent.h"
+#include "session_expires_helper.h"
+
+// The RFC states that the minimum session expiry defaults to 90s if no Min-SE
+// header is present.
+const static SessionExpiresHelper::SessionInterval DEFAULT_MIN_SE = 90;
+
+
```
bossmc: Extra blank line (inconsistent with later in file).  Couple more examples below too.

sprout/session_expires_helper.cpp
```diff
@@ -0,0 +1,204 @@
...
+bool SessionExpiresHelper::timer_supported(pjsip_msg* msg)
+{
+  pjsip_supported_hdr* supported_hdr = (pjsip_supported_hdr*)
+    pjsip_msg_find_hdr(msg, PJSIP_H_SUPPORTED, NULL);
+
+  if (supported_hdr != NULL)
+  {
+    for (unsigned ii = 0; ii < supported_hdr->count; ++ii)
+    {
+      if (pj_strcmp(&supported_hdr->values[ii], &STR_TIMER) == 0)
```
bossmc: Case-sensitive?  I know the spec is clear that it's lower case, but when has a spec stopped a UE from doing something weird?

sprout/session_expires_helper.cpp
```diff
@@ -0,0 +1,204 @@
...
+                            min_se_hdr->expires :
+                            DEFAULT_MIN_SE;
+
+  if ((se_hdr != NULL) && (se_hdr->expires < _target_se))
+  {
+    // The request already has a session expires that is below our target. We
+    // don't need to change the value, but we do make a note of it (which is
+    // needed when processing the response).
+    TRC_DEBUG("Session expires already set to %d", se_hdr->expires);
+    _se_on_req = se_hdr->expires;
```
bossmc: I'd prefer if this was done outside the if, so we can't forget to add it to future branches (as the response handling depends on this being filled in).

bossmc:
Looks good.

sprout/session_expires_helper.cpp
AlexHockey: I thought we allowed this to separate out sections for the file to improve readability?

sprout/session_expires_helper.cpp
AlexHockey: That feels weird as the spec is pretty clear you should use 'timer'. Would we also accept typos? ;-)

sprout/session_expires_helper.cpp
AlexHockey: Fair enough.

sprout/session_expires_helper.cpp
bossmc: Fair enough.

Issue #1202, opened 2015-09-30
Call diversions by a B2BUA are not correlated with the original call in SAS


rkd-msw:
My gut feel is that this is due to https://github.com/Metaswitch/sprout/blob/dev/sprout/scscfsproutlet.cpp#L830 (where we create a new AS chain for originating-cdiv requests), which means that our B2BUA correlation mechanism (at https://github.com/Metaswitch/sprout/blob/dev/sprout/aschain.cpp#L182 - relying on associating the new message with the ASChain trail) doesn't work.

Fix is presumably to keep the old trail when creating a new ASChain - let's try that.

rkd-msw:
Retested with #1206, confirmed fixed.

PR #1206, opened 2015-10-02
[Reviewer: Alex] When creating a new AS chain after a retarget, reuse the SAS trail ID

rkd-msw:
Speculative fix to https://github.com/Metaswitch/sprout/issues/1202.

It compiles, but I haven't live-tested - I've reproed https://github.com/Metaswitch/sprout/issues/1202 on the VMware rig, though, so I can live test once it goes into the 'latest' repo.


AlexHockey:
Yes, I think this is the right fix. However I don't like looking at the trail ID as we don't own it. I would rather have the function parameter not take a default and update all the callers.

PR #1207, opened 2015-10-02
[Reviewer: Ellie] Add a config flag to always behave as though the HSS had sent Include…

rkd-msw:
…RegisterRequest and IncludeRegisterResponse in the XML

Some HSSes don't allow you to set these fields, but some TASes require them. This pull request allows you to set `force_third_party_reg_body=Y` in the config, which always behaves as if the HSS had set it.

This makes Clearwater a useful bridge for interoperating between these sort of HSSes and these sort of TASes.

### DIFF
--- a/sprout-base.root/etc/init.d/sprout
+++ b/sprout-base.root/etc/init.d/sprout
@@ -165,6 +165,7 @@ get_daemon_args()
         [ "$enforce_user_phone" != "Y" ] || user_phone_arg="--enforce-user-phone"
         [ "$enforce_global_only_lookups" != "Y" ] || global_only_lookups_arg="--enforce-global-only-lookups"
         [ "$override_npdi" != "Y" ] || override_npdi_arg="--override-npdi"
+        [ "$force_third_party_reg_body" != "Y" ] || force_3pr_body_arg="--force-3pr-body"
 
         # Enable SNMP alarms if informsink(s) are configured
         if [ ! -z "$snmp_ip" ]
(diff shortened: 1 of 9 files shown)


sprout/registration_utils.cpp
```diff
@@ -79,10 +83,16 @@ void send_register_to_as(pjsip_rx_data* received_register,
                          const std::string&,
                          SAS::TrailId);
 
+void RegistrationUtils::init(SNMP::RegistrationStatsTables* _third_party_reg_stats_tables,
+                             bool _force_third_party_register_body)
+{
+  third_party_reg_stats_tables = _third_party_reg_stats_tables;
```
eleanor-merry: Not sure about the names and underscores here

sprout-base.root/etc/init.d/sprout
```diff
@@ -165,6 +165,7 @@ get_daemon_args()
         [ "$enforce_user_phone" != "Y" ] || user_phone_arg="--enforce-user-phone"
         [ "$enforce_global_only_lookups" != "Y" ] || global_only_lookups_arg="--enforce-global-only-lookups"
         [ "$override_npdi" != "Y" ] || override_npdi_arg="--override-npdi"
+        [ "$force_third_party_reg_body" != "Y" ] || force_3pr_body_arg="--force-3pr-body"
```
eleanor-merry: You'll add this to the config options guide too, yes?

eleanor-merry:
LGTM

sprout/registration_utils.cpp
rkd-msw: `third_party_reg_stats_tables_arg` etc. instead?

eleanor-merry: Sure

PR #1212, opened 2015-10-07
Reviewer Alex - Remove CDIV-triggered ACRs

### LINKED ISSUE #1199: On call diversion, Sprout duplicates START ACRs (and doesn't send ENDs)
Run the "Call Diversion AS - Unconditional" clearwater-live-test.  Check the Rf flows.  Often (if not every time), you will see 4 Rf sessions being started(2 originating, 2 terminating) and only 2 being ended (1 originating, 1 terminating).

This actually appears to be due to the flows from Sprout to Ralf - Sprout seems to send multiple START ACRs for the same node type and role, and Ralf (slightly blindly) honors this, creating multiple Rf sessions.  When Sprout sends ENDs, Ralf only sends ENDs for a subset of the Rf sessions (as it only seems to expect to have one session per node type and role.

This issue exhibits with other call diversion scenarios too - the above is just the simplest.

bossmc:
As per the discussion in #1199, this change limits the S-CSCF to perform billing at the beginning of originating and at the end of terminating, removing the billing performed on `cdiv` and removing the spurious INTERIM billing if your `Record-Route` settings cause multiple hops per call leg.

Fixes #1199.

I've tested the `billing-role` change and the `ACR::cancel()` function in UT.  I'll confirm the extra START ACR with a SAS trace on a live system once this is merged.

### DIFF
--- a/include/acr.h
+++ b/include/acr.h
@@ -149,7 +149,8 @@ public:
 
   /// Returns the JSON encoded message in string form.
   ///
-  /// See comments for `_cancelled`.
+  /// If the ACR has been cancelled, this function's behaviour is unspecified
+  /// (though the UTs expect it to return "Cancelled ACR").
   ///
   /// @param   timestamp      Timestamp to be used as Event-Timestamp AVP.
   virtual std::string get_message(pj_time_val timestamp=unspec);
@@ -167,7 +168,8 @@ public:
   /// cancelled.  In general this will be when the relevant transaction or AS
   /// chain has ended.
   /// @param   timestamp      Timestamp to be used as Event-Timestamp AVP.
-  inline void send(pj_time_val timestamp=unspec) {
+  inline void send(pj_time_val timestamp=unspec)
+  {
     if (!_cancelled)
     {
       send_message(timestamp);
@@ -182,13 +184,13 @@ protected:
   /// Tracks if the ACR has been cancelled.
   ///
   /// `send_message()` will not be called if this is true.
-  /// `get_message()` should return an appropriate error string if this is true
+  /// `get_message()` is undefined if this is true
   bool _cancelled;
 
 private:
   /// Called when the ACR message should be triggered.
   ///
-  /// See comments for `_cancelled`.
+  /// This will never be called on a cancelled ACR.
   ///
   /// @param   timestamp      Timestamp to be used as Event-Timestamp AVP.
   virtual void send_message(pj_time_val timestamp=unspec);
--- a/include/acr.h
+++ b/include/acr.h
@@ -147,12 +147,10 @@ public:
   /// @param   caps           Capabiliies as received from I-CSCF.
   virtual void server_capabilities(const ServerCapabilities& caps);
 
-  /// Called when the ACR message should be triggered.  In general this will
-  /// be when the relevant transaction or AS chain has ended.
-  /// @param   timestamp      Timestamp to be used as Event-Timestamp AVP.
-  virtual void send_message(pj_time_val timestamp=unspec);
-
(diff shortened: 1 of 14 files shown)


include/acr.h
```diff
@@ -164,6 +162,37 @@ class ACR
 
   /// Set the default CCF for this ACR.
   virtual void set_default_ccf(const std::string& default_ccf);
+
+  /// Called when the ACR message should be sent if it's not yet been
+  /// cancelled.  In general this will be when the relevant transaction or AS
+  /// chain has ended.
+  /// @param   timestamp      Timestamp to be used as Event-Timestamp AVP.
+  inline void send(pj_time_val timestamp=unspec) {
```
AlexHockey: We don't allow egyptian braces.

sprout/acr.cpp
```diff
@@ -653,6 +658,11 @@ void RalfACR::send_message(pj_time_val timestamp)
 
 std::string RalfACR::get_message(pj_time_val timestamp)
 {
+  if (_cancelled)
+  {
+    return "Cancelled ACR";
+  }
+
```
AlexHockey: Why do we need this change. Who calls `get_message` outside of `send`?

bossmc: Unit tests :disappointed:

include/scscfsproutlet.h
```diff
@@ -284,10 +284,10 @@ class SCSCFSproutletTsx : public SproutletTsx
   bool lookup_ifcs(std::string public_id,
                    Ifcs& ifcs);
 
-  /// Record-Route the S-CSCF sproutlet into a dialog.  The parameter passed
-  /// will be attached to the Record-Route and can be used to recover the
+  /// Record-Route the S-CSCF sproutlet into a dialog.  The third parameter
+  /// passed may be attached to the Record-Route and can be used to recover the
   /// billing role that is in use on subsequent in-dialog messages.
-  void add_record_route(pjsip_msg* msg, NodeRole);
+  void add_record_route(pjsip_msg* msg, bool, NodeRole);
```
AlexHockey: Needs a comment explaining what the bool does

bossmc: True

sprout/scscfsproutlet.cpp
```diff
@@ -1566,6 +1556,39 @@ void SCSCFSproutletTsx::add_record_route(pjsip_msg* msg,
 
     _record_routed = true;
   }
+
+  // Ensure the billing scope flag is set on the RR header.
+  if (billing_rr)
+  {
+    pjsip_route_hdr* rr = (pjsip_route_hdr*)pjsip_msg_find_hdr(msg,
```
AlexHockey: Probably cleaner to put this in an else branch in the above if test?

include/acr.h
```diff
@@ -147,12 +147,10 @@ class ACR
...
   virtual void server_capabilities(const ServerCapabilities& caps);
 
-  /// Called when the ACR message should be triggered.  In general this will
-  /// be when the relevant transaction or AS chain has ended.
-  /// @param   timestamp      Timestamp to be used as Event-Timestamp AVP.
-  virtual void send_message(pj_time_val timestamp=unspec);
-
   /// Returns the JSON encoded message in string form.
+  ///
+  /// See comments for `_cancelled`.
```
AlexHockey: I'm not keen on implementation being referenced on the interface.

sprout/acr.cpp
AlexHockey: Boooo!

AlexHockey:
Looks good. A bit of picky commenting feedback inline.

PR #1214, opened 2015-10-08
Reviewer: Rob] Strip Proxy-Authorization headers

### LINKED ISSUE #1213: A non-REGISTERing PBX origination yields incorrect statistic counts.
When performing a single non-REGISTERing PBX origination, the 'sproutNonRegisterAuthentication-Attempts' and 'sproutNonRegisterAuthenticationSuccesses' statistics are being incremented by 4 instead of 1. This is due to the logic being triggered on all TAS interactions (which includes unnecessary validation requiring extra memcached look-ups). The root cause is the TAS passing through the Proxy-Authorization header back to the S-CSCF. That header should probably be stripped before invoking the AS chain to avoid the potential for this situation.

AlexHockey:
Rob, please could you review this change to strip any Proxy-Authorization headers out of the message before passing the request up out of the auth module. 

I've not done this for Authorization headers as 3rd party REGISTERs may need to include the entire received REGISTER request :-0.

Tested in UT. Fixes #1213.

### DIFF
--- a/sprout/authentication.cpp
+++ b/sprout/authentication.cpp
@@ -831,7 +831,13 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
       if (status == PJ_SUCCESS)
       {
         // Request authentication completed, so let the message through to
-        // other modules.
+        // other modules. Remove any authorization headers first.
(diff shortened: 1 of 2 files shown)


rkd-msw:
Yep, looks good.

Issue #1184, opened 2015-09-04
Confusion about signature of on_tx_request()


dimbleby:
As clang points out

    In file included from scscfplugin.cpp:44:
    /home/clearwater/sprout/include/scscfsproutlet.h:219:16: warning: 'SCSCFSproutletTsx::on_tx_request' hides overloaded virtual function
          [-Woverloaded-virtual]
      virtual void on_tx_request(pjsip_msg* req);
                   ^
    /home/clearwater/sprout/include/sproutlet.h:312:16: note: hidden overloaded virtual function 'SproutletTsx::on_tx_request' declared here: different
          number of parameters (2 vs 1)
      virtual void on_tx_request(pjsip_msg* req, int fork_id) { }
                   ^

Same thing in `ICSCFSproutletTsx` and `BGCFSproutletTsx`.

As well as fixing these, suggest making use of the C++11 `override` specifier when you're intending to override base class methods.

PR #1220, opened 2015-10-19
Compiler warnings

dimbleby:
Includes a fix for #1184

### DIFF
--- a/sprout/custom_headers.cpp
+++ b/sprout/custom_headers.cpp
@@ -68,9 +68,9 @@ typedef int   (*print_fptr)(void *hdr, char *buf, pj_size_t len);
 /*****************************************************************************/
 pjsip_hdr_vptr session_expires_hdr_vptr =
 {
-  (clone_fptr) &pjsip_session_expires_hdr_clone,
-  (clone_fptr) &pjsip_session_expires_hdr_shallow_clone,
-  (print_fptr) &pjsip_session_expires_hdr_print_on,
+   &pjsip_session_expires_hdr_clone,
+   &pjsip_session_expires_hdr_shallow_clone,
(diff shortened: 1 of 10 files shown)


sprout/custom_headers.cpp
```diff
@@ -867,7 +868,9 @@ int pjsip_p_c_v_hdr_print_on(void* h, char* buf, pj_size_t len)
   pj_memcpy(p, "icid-value=", 11);
   p += 11;
   *p++ = '"';
-  pj_memcpy(p, hdr->icid.ptr, hdr->icid.slen);
+  if (hdr->icid.slen) {
+    pj_memcpy(p, hdr->icid.ptr, hdr->icid.slen);
```
bossmc: What warning is this covering?  `memcpy` 0 bytes is specified to be safe:

> Where an argument declared as `size_t n` specifies the length of the array for a function, `n` can have the value zero on a call to that function. Unless explicitly stated otherwise in the description of a particular function in this sub-clause, pointer arguments on such a call shall still have valid values, as described in 7.1.4. On such a call, a function that locates a character finds no occurrence, a function that compares two character sequences returns zero, and a function that copies characters copies zero characters.

Are you worried that `hdr->icid.ptr` would be `NULL` in this case?

dimbleby: Yes, `hdr->icid.ptr` can be `NULL` in that case - and indeed it's the `NULL` rather than the `0` that the sanitizer complains about.  

I wrote the `if` test that way (rather than checking the pointers) for consistency with the surrounding code.

bossmc: Cool, makes sense.

PR #1221, opened 2015-10-19
[Reviewer: AMC] Pass the SAS trail ID into the chronos connection when setting auth timers

AlexHockey:
Very simple trail ID plumbing fix. I tested this by writing a script to lint our SAS protocol flows. I ran the live tests before the fix and checked the lint found the problem, and after the fix to prove the problem was fixed.


bossmc:
Cool.

PR #1223, opened 2015-10-20
[Reviewer: Ellie] Reinstate Bono logging

rkd-msw:
I noticed that no Bono nodes have been logging for a while - this is because short arguments like `-a` are incompatible with using `=` for arguments. This switches all the short-form arguments to long-form ones.


eleanor-merry:
LGTM

PR #1227, opened 2015-10-21
[Reviewer: Andy] Make the S-CSCF override the session ID when deciding to record-route

### LINKED ISSUE #1222: Rf billing incorrect for Term Interim and Stop ACRs when using Sentinel VoLTE TAS.
For a basic VoLTE call it appears that the Term Interim ACR is incomplete and the Term Stop ACR is not generated. This seems to be due to the S-CSCF using the Call-ID for the terminating leg it sends to the TAS for the Start ACR, and the Call-ID for the terminating leg sent by the TAS (to the S-CSCF) for the Interim and Stop ACRs. As a result, Ralf is unable to lookup the session for the latter two cases. All of this is probably due to the TAS acting as a B2BUA. A SAS trace and Wireshark capture of the Rf interface at the CDF will be uploaded to a diagnostics folder.

Also, I do not know if this is intentional or not, but Diameter ACR/ACA flows for Interim records are not displayed in SAS ladder diagrams.

AlexHockey:
Andy, please can you review this PR that fixes #1222. As discussed, I've added a new method to allow the S-CSCF to override the default session ID for an ACR, and made the S-CSCF sproutlet call this whenever it record-routes itself into a dialog to do billing. 

I tested the ACR code in UT. I tested the entire change (including the S-CSCF part) live, by using mangelwurzel to change the call ID and checking in SAS that the correct session IDs were used on INTERIMs and STOPs send by sprout.

### DIFF
--- a/sprout/ut/acr_test.cpp
+++ b/sprout/ut/acr_test.cpp
@@ -122,10 +122,9 @@ protected:
 
     if (!rc)
     {
-//      Json::FastWriter writer;
-  //    printf("JSON comparison failed\nReceived\n%s\nExpected\n%s\n",
-    //         json_output.toStyledString().c_str(),
-      //TODO       json_expected.toStyledString().c_str());
+      printf("JSON comparison failed\nReceived\n%s\nExpected\n%s\n",
+             output.c_str(),
+             exp_str.c_str());
     }
 
     return rc;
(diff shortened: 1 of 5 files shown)


sprout/ut/acr_test.cpp
```diff
@@ -1108,3 +1115,139 @@ TEST_F(ACRTest, SCSCFPublish)
...
+  acr->as_info("sip:as1.homedomain:5060;transport=TCP",
+               "sip:6505559999@homedomain",
+               200,
+               false);
+  ts.msec = 70;
+  acr->tx_response(parse_msg(r200ok.get()), ts);
+
+  // Build and checked the resulting Rf ACR message.
+  string rf_acr = acr->get_message(ts);
+  EXPECT_TRUE(compare_acr(rf_acr, "acr_scscftermcall_start_changed_call_id.json"));
```
bossmc: Have you forgotten to add this `json` file to git?

bossmc:
Testing sounds good, code looks good.

sprout/ut/acr_test.cpp
AlexHockey: Yep.

PR #1233, opened 2015-10-23
Add support for unsolicited outdials

dimbleby:
As discussed.  Built on @mirw's work in a hackathon some time ago.


include/basicproxy.h
```diff
@@ -256,7 +256,7 @@ class BasicProxy
 
     pj_timer_entry       _trying_timer;
     static const int     TRYING_TIMER = 1;
-    pthread_mutex_t      _trying_timer_lock;
+    pthread_mutex_t      _trying_timer_lock = PTHREAD_MUTEX_INITIALIZER;
```
AlexHockey: I'm not sure I like this. Most people are used to looking at the constructor to figure out how members are initialized, not the class definition.

include/sproutlet.h
```diff
@@ -617,6 +624,9 @@ class Sproutlet
 
   /// The host name of this service.
   const std::string _service_host;
+
+  /// The sproutletproxy that hosts this sproutlet.
+  SproutletProxy* _proxy = NULL;
```
AlexHockey: Ditto

include/basicproxy.h
dimbleby: Of course where there's more than one way to do it, there's more than one opinion about how to do it...

I think that the main argument in favour of initializing this way is that it avoids duplication when you have multiple constructors that would all do the same thing; and that when you add a new constructor you can't accidentally miss fields that are already defaulted like this.  Personally I much prefer this style.

Still I'm happy to change it if you want.  

Do you feel strongly enough about this that you'd like me to change it, or are you simply noting that you're not sure what you think?

AlexHockey: The multiple constructors worry is a valid one but I think that constructor delegation is a better solution. 

Given that the code would work if you initialized this in the constructor, this is an existing class, and that I think people may get tripped up by this style of initialization, I would like you to initialize in the constructor please.

AlexHockey:
After a fair bit of discussion, we decided not to take this change into Project Clearwater in it's current form. We think that the right approach is:

* Move the `BasicProxy` functionality into the `SproutletProxy` and delete the `BasicProxy` (which would then be unused). 
* Split the `SproutletProxy::UasTsx` class into a `Context` class that handles scheduling and routing, and a new `UasTsx` that handles the binding to PJSIP. 
* Add a (not SIP aware) `Trigger` interface and change the `UasTsx` to implement it.

PR #1235, opened 2015-10-26
[Reviewer: Andy] Fixes #1209 - Reload enum json on sighup

### LINKED ISSUE #1209: enum.json is not reloaded by a 'service sprout reload'
Unlike bgcf.json and s-cscf.json

This means that /usr/share/clearwater/clearwater-config-manager/scripts/upload_enum_json won't activate any new enum config uploaded (requires an explicit restart of sprout).

Presumably this is down to the fact that an Updater object (as created in scscfselector.cpp and bgcfservice.cpp to take care of s-cscf.json and bgcf.json uploads respectively) hasn't been created in enumservice.cpp

Tom-Rees:
Andy

Would you mind reviewing these changes to allow Sprout to pickup changes to enum.json.
A couple of areas I'm less confident on are:
* The locking using pthread mutexes.
* Memory management - Looking at [scscfselector](https://github.com/Metaswitch/sprout/blob/dev/sprout/scscfselector.cpp) line 151, we don't seem to destroy the old list (_scscfs) in the way I have done in my change. I think this might mean that scscfselector leaks the memory storing the config whenever it is reloaded, although if I'm misunderstanding something then my changes may be wrong instead.

Fixes #1209 

I tested this on our system with both good and broken enum.json files and it behaved as expected.

Thanks,
Tom

### DIFF
--- a/include/enumservice.h
+++ b/include/enumservice.h
@@ -43,9 +43,9 @@
 #include <list>
 #include <string>
 #include <boost/regex.hpp>
+#include <boost/thread.hpp>
 #include <netinet/in.h>
 #include <ares.h>
-#include <pthread.h>
 #include "sas.h"
 #include "baseresolver.h"
 #include "dnsresolver.h"
@@ -96,8 +96,6 @@ public:
   std::string lookup_uri_from_user(const std::string& user, SAS::TrailId trail) const;
 
 private:
-  void destroy_prefixes();
-
   struct NumberPrefix
   {
     std::string prefix;
@@ -105,12 +103,15 @@ private:
     std::string replace;
   };
 
-  std::list<struct NumberPrefix*> _number_prefixes;
-
+  std::vector<NumberPrefix> _number_prefixes;
   std::string _configuration;
-  NumberPrefix* prefix_match(const std::string& number) const;
   Updater<void, JSONEnumService>* _updater;
-  mutable pthread_mutex_t _number_prefixes_rw_lock;
+
+  // Mark as mutable to flag that this can be modified without affecting the
+  // external behaviour of the class, allowing for locking in 'const' methods.
+  mutable boost::shared_mutex _number_prefixes_rw_lock;
+
+  const NumberPrefix* prefix_match(const std::string& number) const;
 };
 
 /// @class DNSEnumService
--- a/include/enumservice.h
+++ b/include/enumservice.h
@@ -45,10 +45,12 @@
 #include <boost/regex.hpp>
 #include <netinet/in.h>
 #include <ares.h>
+#include <pthread.h>
 #include "sas.h"
 #include "baseresolver.h"
 #include "dnsresolver.h"
 #include "communicationmonitor.h"
+#include "updater.h"
 
 /// @class EnumService
 ///
@@ -88,9 +90,14 @@ public:
   JSONEnumService(std::string configuration = "./enum.json");
   ~JSONEnumService();
 
+  // Updates the enum configuration
+  void update_enum();
+
   std::string lookup_uri_from_user(const std::string& user, SAS::TrailId trail) const;
 
 private:
+  void destroy_prefixes();
+
   struct NumberPrefix
   {
     std::string prefix;
@@ -100,8 +107,10 @@ private:
 
   std::list<struct NumberPrefix*> _number_prefixes;
 
+  std::string _configuration;
   NumberPrefix* prefix_match(const std::string& number) const;
-
+  Updater<void, JSONEnumService>* _updater;
+  mutable pthread_mutex_t _number_prefixes_rw_lock;
 };
 
 /// @class DNSEnumService
--- a/sprout/enumservice.cpp
+++ b/sprout/enumservice.cpp
@@ -96,8 +96,6 @@ JSONEnumService::JSONEnumService(std::string configuration):
   _configuration(configuration),
   _updater(NULL)
 {
-  pthread_mutex_init(&_number_prefixes_rw_lock, NULL);
-
   // create and updater which, by default, runs the function when initialized
   _updater = new Updater<void, JSONEnumService>(this,
                                   std::mem_fun(&JSONEnumService::update_enum));
@@ -148,7 +146,7 @@ void JSONEnumService::update_enum()
 
   try
   {
-    std::list<struct NumberPrefix*> new_number_prefixes;
+    std::vector<NumberPrefix> new_number_prefixes;
 
     JSON_ASSERT_CONTAINS(doc, "number_blocks");
     JSON_ASSERT_ARRAY(doc["number_blocks"]);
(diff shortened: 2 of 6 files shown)


include/enumservice.h
```diff
@@ -100,8 +107,10 @@ class JSONEnumService : public EnumService
 
   std::list<struct NumberPrefix*> _number_prefixes;
```
bossmc: This is a `list` of pointers (unlike in `SCSCFSelector` which uses a `vector` of `scscf_t` structs.  This explains why your code needs to allocate the `NumberPrefix*` pointers from the heap and why you need `destroy_prefixes()`.

I think that, given that this code is run rarely, it's probably better to switch this to storing `NumberPrefix` objects (not pointers) and possibly also to store them in a `vector` not a `list` (since those are likely faster to iterate over).

include/enumservice.h
```diff
@@ -100,8 +107,10 @@ class JSONEnumService : public EnumService
...
 
+  std::string _configuration;
   NumberPrefix* prefix_match(const std::string& number) const;
-
+  Updater<void, JSONEnumService>* _updater;
+  mutable pthread_mutex_t _number_prefixes_rw_lock;
```
bossmc: Slinky use of `mutable`, care to add a comment explaining why?

bossmc: Also, this is a standard mutex, not a read-write lock (https://docs.oracle.com/cd/E19455-01/806-5257/6je9h032u/index.html) which means that all readers (service threads) will be serialized on access to this object.  Switching to a read-write lock means that multiple readers can all perform lookups in parallel **unless** there are any writers in which case the writer gets unilateral access to allow it to perform updates as needed.

sprout/enumservice.cpp
```diff
@@ -175,19 +188,24 @@ JSONEnumService::JSONEnumService(std::string configuration)
         // Badly formed number block.
         TRC_WARNING("Badly formed ENUM number block (hit error at %s:%d)",
                     err._file, err._line);
-        CL_SPROUT_ENUM_FILE_INVALID.log(configuration.c_str());
+        CL_SPROUT_ENUM_FILE_INVALID.log(_configuration.c_str());
       }
     }
+
+    // Delete the old prefixes and store the new ones under mutex protection.
+    pthread_mutex_lock(&_number_prefixes_rw_lock);
```
bossmc: This will need to grab the write lock.

sprout/enumservice.cpp
```diff
@@ -175,19 +188,24 @@ JSONEnumService::JSONEnumService(std::string configuration)
...
   {
     TRC_ERROR("Badly formed ENUM configuration data - missing number_blocks object");
-    CL_SPROUT_ENUM_FILE_INVALID.log(configuration.c_str());
+    CL_SPROUT_ENUM_FILE_INVALID.log(_configuration.c_str());
   }
 }
 
-
-JSONEnumService::~JSONEnumService()
+void JSONEnumService::destroy_prefixes()
```
bossmc: This is unnecessary if we switch to storing `NumberPrefixes` inside the `list/vector` rather than pointers.

sprout/enumservice.cpp
```diff
@@ -240,6 +270,10 @@ std::string JSONEnumService::lookup_uri_from_user(const std::string &user, SAS::
 
 JSONEnumService::NumberPrefix* JSONEnumService::prefix_match(const std::string& number) const
 {
+  JSONEnumService::NumberPrefix* xoPrefix = NULL;
```
bossmc: While I've seen the `xo...` prefix before, it's not something we use in our code, can you use a more descriptive name?  `found_prefix`?

sprout/enumservice.cpp
```diff
@@ -240,6 +270,10 @@ std::string JSONEnumService::lookup_uri_from_user(const std::string &user, SAS::
...
+
+  pthread_mutex_lock(&_number_prefixes_rw_lock);
```
bossmc: This can take just the read lock.

bossmc:
Reviewed, I've tried to explain the two questions you had, feel free to contact me if you want to discuss.    On a side note, I think that our other two `Updaters` don't correctly handle multi-threading which is a bit disconcerting... how hard would it be for you to add an RWLock to each of them?

Tom-Rees:
Andy,

I've re-written the locking to use boost's RAII-style RWlocks. I also changed enumservice to use a vector instead of a list of pointers.

I couldn't see anywhere that we had used these before so I tested the locking behaviour with:

```c++
#include <boost/thread.hpp>
#include <iostream>
#include <string>
#include <vector>

boost::shared_mutex lMutex;

void read_lock_loop(int x) {
 for (int i=1; i<5; ++i) {
    {
      boost::shared_lock<boost::shared_mutex> read_lock(lMutex);
      std::cout << x << ": Read lock taken.\n" << std::flush;
      sleep(4);
      std::cout << x << ": Releasing read lock\n" << std::flush;
    }
    sleep(1);
  }
}

void take_write_lock() {
  std::cout << "Attempting to take write lock\n" << std::flush;
  boost::lock_guard<boost::shared_mutex> write_lock(lMutex);
  std::cout << "Write lock taken\n" << std::flush;
}

int main() {
  std::string input;
  std::vector<boost::thread*> lThreads;

  for (int i = 0; i < 10; ++i) {
    boost::thread read1(read_lock_loop, i);
    lThreads.push_back(&read1);
    sleep(1);
  }

  boost::thread write1(take_write_lock);
  write1.join();

  std::cout << "Write done, joining readers\n" << std::flush;

  for(std::vector<boost::thread*>::iterator i = lThreads.begin(); i!= lThreads.end(); ++i) {
    (*i)->join();
  }

  return 0;
}
```

Which demonstrates the desired behaviour.

Please could you review again?

Cheers,
Tom

sprout/enumservice.cpp
```diff
@@ -238,26 +252,26 @@ std::string JSONEnumService::lookup_uri_from_user(const std::string &user, SAS::
...
 
-    if (number.compare(0, len, (*it)->prefix, 0, len) == 0)
+    if (number.compare(0, len, it->prefix, 0, len) == 0)
     {
       // Found a match, so return it (at the moment we assume the entries are
       // ordered with most specific matches first so we can stop as soon as we
       // have one match).
       TRC_DEBUG("Match found");
-      return *it;
+      return &*it;
```
bossmc: There's a lifetime issue here, the memory that this point to could be freed/invalidated/changed by the `Updater` while the caller is using it.  Better to return by value rather than pointer (or, and this would be my comment if this were lifetime-safe, reference).  This was an existing problem, but as we're here it would be good to fix it now (and presumably in the other two places).

bossmc:
One comment in-line, I think there's still a threading issue in the return values of the lookup functions.  The boost locking API stuff looks good to me.

BTW, your test code has a few problems:

* you're storing pointers to `boost::thread`s in your `std::vector` but those pointers all point to the same stack variable... so the `join` calls don't work as expected.
    * I discovered `emplace_back` on `std::vector` which allows you to make the `boost::thread` directly into the vector which resolves this nicely).
* holding the read lock for such a long time prevents the write lock from ever getting in on the action so I was seeing writer starvation
    * Making the "work" in the read loop much smaller (I used setting a `volatile int` to 5) meant that the writer got in quickly even with loads of read threads

sprout/enumservice.cpp
Tom-Rees: I was initially going to return a value, not a pointer. The problem I had was that it would require some notion of a 'not found' struct for when we didn't find a match. I couldn't think of a  neat way of representing this and so I moved the locking to the callee.

I think there's two options to resolve this:
1. Add a comment to flag that this function assumes the callee has taken the read lock. 
2. Create something to represent a number not found. I had assumed that this would be a well known problem, but i couldn't find any good solutions. One possibility might be to extend the struct and give both structs an equality operator, but it feels a bit 'hacky'.

Tom-Rees:
Comment added to remind people to take read lock before calling.

The scscf and bgcf read methods return by value so don't hit this issue.

bossmc:
Looks good!  Nice one.

PR #1237, opened 2015-10-27
In-tree sproutlet builds

dimbleby:
This adds the ability to develop sproutlets without having to make changes to the core sprout repository.

The approach is that:
- Plugins live in a `plugins/my_plugin` directory
- They provide their own makefiles which run their own builds, packaging, tests
  - So eg while developing `hello_world` you’ll mostly live in the `plugins/hello_world` directory and run `make`, `make test`, `make deb`, etc from there
- However when you go up a directory and build, test or package sprout, sprout's makefiles will look for plugins and do whatever-we’re-doing-for-sprout for the plugins too

I'm intending also to offer you an example 'hello world' sproutlet that works within this infrastructure.


dimbleby:
See sample app [here](https://github.com/dimbleby/greeter).  

(Will need transferring to a Metaswitch repository if / when happy).

sprout-base.root/etc/init.d/sprout
```diff
@@ -275,6 +275,12 @@ get_daemon_args()
         [ "$stateless_proxies" = "" ]             || DAEMON_ARGS="$DAEMON_ARGS --stateless-proxies=$stateless_proxies"
         [ "$ralf_threads" = "" ]                  || DAEMON_ARGS="$DAEMON_ARGS --ralf-threads=$ralf_threads"
         [ "$non_register_authentication" = "" ]   || DAEMON_ARGS="$DAEMON_ARGS --non-register-authentication=$non_register_authentication"
+
+        for script in /usr/share/clearwater/sprout/plugin_conf.d/*.plugin_conf
```
bossmc: Who creates this folder?  What if it's empty (I think `$script` becomes `/usr/share/clearwater/sprout/plugin_conf.d/*.plugin_conf` and the body of the loop errors)?

bossmc:
One comment (and the code doesn't cleanly merge) but otherwise this looks good.

sprout-base.root/etc/init.d/sprout
mbt-msw: The directory is only there if a plugin installs something there, but that's ok as the expansion becomes `/usr/share/clearwater/sprout/plugin_conf.d/*.plugin_conf`.   I've added a check for the expansion value, which also makes sure we don't fail to start sprout if someone installs a non-executable plugin_conf script.

bossmc:
One final request, can you add some docs covering how to use this build system to the `docs` folder?  This might just require a copy-paste of @rkd-msw's blog post or @dimbleby's sample Sproutlet's README.

PR #1240, opened 2015-10-28
Fix PJUtils::get_header_value

dimbleby:


sprout/pjutils.cpp
```diff
@@ -1606,11 +1606,15 @@ std::string PJUtils::get_header_value(pjsip_hdr* header)
   // pjsip_hdr_print_on doesn't appear to null-terminate the string - do this by hand
   buf2[len] = '\0';
```
bossmc: Since you're tracking `len` below, this is now unnecessary.

bossmc: And potentially dangerous (if the header is exactly `MAX_HDR_SIZE` bytes long).

bossmc:
One comment.

PR #1241, opened 2015-10-28
Keep sproutlet requests in order

dimbleby:
Sproutlets might reasonably expect that if they request that messages
1, 2 and 3 be sent then they will be sent in that order.

But the use of an unordered_map prevents this from being true.
Therefore prefer a plain map.


bossmc:
This introduces an interesting implementation assumption that's visible on the API (namely that `fork_ids` are numerically increasing within a `Tsx`).

What's the use case for assuming messages will be sent in order?  If they're forks of a call, there's no guarantee anything downstream of the sproutlet seeing them in order/treating them as ordered, if they're requests in the same dialog then they're invalid (glare).

dimbleby:
Eg you're dealing with an MRF, have sent an INVITE and got a 200 OK.  You now want to send an ACK, and also an INFO to do something (say to create a conference).

You'd very much like to send the ACK and the INFO in that order.

PR #1244, opened 2015-10-29
Correct Reason parameter in History-Info header

### LINKED ISSUE #873: Reason Header Field of History-Info Header added for CDIV is malformed.
It appears that the Reason Header Field of a History-Info added for CDIV is malformed. IR-92 references TS 24.604 (4.5.2.6.2.2, (b)) which says the following for the History-Info header: “If the diversion is based on a SIP response from the served user, a Reason header in escaped form shall be included in accordance with IETF RFC 7044 [3].” Looking at 7044 it shows an example of a “embedded” Reason header in section 5 (```History-Info: <sip:UserA@ims.example.com?Reason=SIP%3Bcause%3D302>;index=1.1,…```) and points to 3326 for details. RFC 3326 defines the grammar for the Reason Header Field in section 2 (which lines up with the example in 7044). When Sprout sends a diverted Invite (for busy), the first History-Info header looks like: ```History-Info: <sip:4144910002@vm2.richlab.datcon.co.uk;text=Busy%20Here;cause=486;Reason=SIP>;index=1```. This does not seem to follow the grammar as expected. 

Alex did some digging and found the following: RFC 4244 (the precursor to 7044) contains two different examples for how to encode the cause and text. Section 4.2 uses this form:
```
<sip:UserA@ims.example.com?Reason=SIP%3Bcause%3D302>;index=1;foo=bar
```
However the flows in section 4.5 use this representation, which is the one Sprout is using.
```
<sip:User2@UA2.example.com?Reason=SIP;cause=408;text="RequestTimeout">;index=1.1.1
```
However 7044 only mentions the first form, so I think this is the one we should use. (In fact I think this was probably the intention of the original RFC 4244 – that RFC did not place any special importance on the “cause” or “text” parameters, while RFC 3326 (Reason header) does which makes me think that RFC 4244 intended implementers use the section 4.2 form, and the section 4.5 form was included in error).

chris-elford-metaswitch:
This pull request fixes #873 by correcting the Reason field in the History-Info header. The old behavior was to include each part of the Reason field as a separate parameter in the URI. The fixed behavior is to URL escape the values into a single parameter as per RFC 7044.

I've checked the RFC and confirmed that I agree with the interpretation in the issue. I've run the live tests against my code with no errors.

### DIFF
--- a/sprout/basicproxy.cpp
+++ b/sprout/basicproxy.cpp
@@ -1733,6 +1733,8 @@ void BasicProxy::UACTsx::cancel_pending_tsx(int st_code)
       if (_tdata->msg->line.req.method.id == PJSIP_INVITE_METHOD)
       {
         TRC_DEBUG("Sending CANCEL request");
+
+        // See issue 1232.
         pjsip_tx_data *cancel = PJUtils::create_cancel(stack_data.endpt,
                                                        _tsx->last_tx,
                                                        st_code);
--- a/sprout/bono.cpp
+++ b/sprout/bono.cpp
@@ -2863,7 +2863,7 @@ void UACTransaction::cancel_pending_tsx(int st_code)
       // See issue 1232.
       pjsip_tx_data* cancel = PJUtils::create_cancel(stack_data.endpt,
                                                      _tsx->last_tx,
-                                                     _tsx->status_code)
+                                                     _tsx->status_code);
 
       if (trail() == 0)
       {
--- a/sprout/bono.cpp
+++ b/sprout/bono.cpp
@@ -2860,18 +2860,11 @@ void UACTransaction::cancel_pending_tsx(int st_code)
     TRC_DEBUG("Found transaction %s status=%d", name(), _tsx->status_code);
     if (_tsx->status_code < 200)
     {
-      pjsip_tx_data *cancel;
-      pjsip_endpt_create_cancel(stack_data.endpt, _tsx->last_tx, &cancel);
-      if (st_code != 0)
-      {
-        char reason_val_str[96];
-        const pj_str_t* st_text = pjsip_get_status_text(st_code);
-        sprintf(reason_val_str, "SIP ;cause=%d ;text=\"%.*s\"", st_code, (int)st_text->slen, st_text->ptr);
-        pj_str_t reason_name = pj_str("Reason");
-        pj_str_t reason_val = pj_str(reason_val_str);
-        pjsip_hdr* reason_hdr = (pjsip_hdr*)pjsip_generic_string_hdr_create(cancel->pool, &reason_name, &reason_val);
-        pjsip_msg_add_hdr(cancel->msg, reason_hdr);
-      }
+      // See issue 1232.
+      pjsip_tx_data* cancel = PJUtils::create_cancel(stack_data.endpt,
+                                                     _tsx->last_tx,
+                                                     _tsx->status_code)
+
       if (trail() == 0)
       {
         TRC_ERROR("Sending CANCEL request with no SAS trail");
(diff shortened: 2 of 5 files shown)


sprout/pjutils.cpp
```diff
@@ -1825,25 +1822,35 @@ void PJUtils::update_history_info_reason(pjsip_uri* history_info_uri, pj_pool_t*
...
     {
+      // Need to build up a std::string containing the parameter value
+      // so that it can be manipulated using C++ utility functions.
+      std::stringstream value_builder;
+      std::string reason_value_string;
+      pj_str_t reason_value;
+
+      const pj_str_t* reason_text = pjsip_get_status_text(code);
+
+      value_builder << "SIP;cause=";
```
rkd-msw: Can you chain the `<<`s - `value_builder << "SIP;cause=" << code << ";text=\""...`?

rkd-msw: Or an `snprintf`, or "align your chevrons".

sprout/basicproxy.cpp
```diff
@@ -1733,6 +1733,8 @@ void BasicProxy::UACTsx::cancel_pending_tsx(int st_code)
       if (_tdata->msg->line.req.method.id == PJSIP_INVITE_METHOD)
       {
         TRC_DEBUG("Sending CANCEL request");
+
+        // See issue 1232.
```
rkd-msw: Not sure this comment adds much value, but I don't mind that much.

sprout/bono.cpp
```diff
@@ -2860,18 +2860,11 @@ void UACTransaction::cancel_pending_tsx(int st_code)
     TRC_DEBUG("Found transaction %s status=%d", name(), _tsx->status_code);
     if (_tsx->status_code < 200)
     {
-      pjsip_tx_data *cancel;
-      pjsip_endpt_create_cancel(stack_data.endpt, _tsx->last_tx, &cancel);
-      if (st_code != 0)
-      {
-        char reason_val_str[96];
```
rkd-msw: I was a bit nervous about this change but then realised that create_cancel already does this.

sprout/pjutils.cpp
```diff
@@ -1825,25 +1822,35 @@ void PJUtils::update_history_info_reason(pjsip_uri* history_info_uri, pj_pool_t*
     pjsip_sip_uri* history_info_sip_uri = (pjsip_sip_uri*)history_info_uri;
     if (pj_list_empty(&history_info_sip_uri->other_param))
     {
+      // Need to build up a std::string containing the parameter value
+      // so that it can be manipulated using C++ utility functions.
+      std::stringstream value_builder;
+      std::string reason_value_string;
+      pj_str_t reason_value;
+
+      const pj_str_t* reason_text = pjsip_get_status_text(code);
```
rkd-msw: Feels like this is partly common with PJUtils::add_reason - could we refactor the common bit out, to have a single function that takes a status code and returns `SIP;cause=777;text="whatever"`?

rkd-msw:
Looks good - might be worth refactoring to share code with PJUtils::add_reason.

rkd-msw:
(P.S. Convention is to include `[Reviewer: Rob]` in the title as well as assigning)

sprout/pjutils.cpp
chris-elford-metaswitch: I've stuck it all on one line. It goes over the 80-limit but it's neater that way.

chris-elford-metaswitch:
Rob, I've followed your suggestion and factored out the reason string building part of the code. I this what you expected?

sprout/pjutils.cpp
```diff
@@ -59,6 +59,11 @@ extern "C" {
 #include "enumservice.h"
 #include "uri_classifier.h"
 
+/// Definitions of static functions.
+namespace PJUtils {
```
rkd-msw: Not worth bothering with the namespace if it's static.

rkd-msw:
One comment, then good to merge.

PR #1248, opened 2015-11-02
Move MockSproutletProxy etc to make them generally available

dimbleby:
On the other hand I have a pull request that will make its way across to you in due course to re-introduce these mocks in cpp_common, where they can be used by sproutlet unit tests too.

### DIFF
--- a/sprout/ut/mock_sproutlet.h
+++ b/sprout/ut/mock_sproutlet.h
@@ -1,0 +1,82 @@
+/**
+ * @file mock_sproutlet.h Mock Sproutlet
+ *
+ * Project Clearwater - IMS in the Cloud
+ * Copyright (C) 2015  Metaswitch Networks Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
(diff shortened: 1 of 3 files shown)


dimbleby:
Decided that these mocks did belong in this repository rather than in cpp-common, but put them in their own header files so as to make them generally available.

Includes mocks for methods introduced at #1233, so likely doesn't make sense to take this before that.

bossmc:
@AlexHockey - Continuation of the work in #1233 which you're reviewing.  I think this change is fine once you've reviewed that.   Passing to you to click the big green button when ready.

rkd-msw:
Is https://github.com/Metaswitch/clearwater-app-servers/tree/b57cc36bf11ef1c2035777144051158d5a0aa0cf a sensible place for them?

bossmc:
Nope, that repository hold the application server API which this isn't.  I question whether that repo is adding much or making life more difficult...

AlexHockey:
Not sure how this builds given `sproutletproxy_test.cpp` doesn't include the new headers...

mbt-msw:
It compiles because MockSproutlet and friends are not used in the existing code.  They are, however, useful for other sproutlets to use in their unit tests, which is why they're being retained in their own headers.

AlexHockey:
Cool.

PR #1249, opened 2015-11-04
[Reviewer: Rob] Copy URI setting as well

eleanor-merry:
Fix to copy the _uri as well when copying an AoR

Tested in UTs


rkd-msw:
+1

PR #1250, opened 2015-11-04
[Reviewer: Alex] UT - don't increase our token rate spuriously during slow periods

rkd-msw:
UT for https://github.com/Metaswitch/cpp-common/pull/397. Also fixes up some unrealistic behaviour in UT, where we call `request_complete()` despite never having called `admit_request()`.


AlexHockey:
Looks good.

PR #1252, opened 2015-11-06
[Reviewer: AMC] Flush trails on requests that are handled statelessly

AlexHockey:
As discussed, please can you review this PR to send a flush marker on trails that are handled statelessly (ACKs, and requests rejected by bono). I tested this by running the live tests and checking that they worked, and that my SAS lint reported no unflushed trails related to ACK processing.

### DIFF
--- a/sprout/bono.cpp
+++ b/sprout/bono.cpp
@@ -245,6 +245,31 @@ static pj_status_t add_path(pjsip_tx_data* tdata,
 static NodeRole acr_node_role(pjsip_msg *req);
 
 
+// Utility class that automatically flushes a trail ID when it goes out of
+// scope (unless the user decides it isn't required). This is useful when a
+// function has multiple exit points that should all cause the trail to be
+// flushed.
+class TrailFlusher
+{
+public:
+  TrailFlusher(SAS::TrailId trail) : _trail(trail), _flush_required(true) {}
+
+  void set_flush_required(bool required) { _flush_required = required; }
+
+  ~TrailFlusher()
+  {
+    if (_flush_required)
+    {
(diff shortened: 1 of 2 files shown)


sprout/bono.cpp
```diff
@@ -245,6 +245,31 @@ static pj_status_t add_path(pjsip_tx_data* tdata,
 static NodeRole acr_node_role(pjsip_msg *req);
 
 
+// Utility class that automatically flushes a trail ID when it goes out of
+// scope (unless the user decides it isn't required). This is useful when a
+// function has multiple exit points that should all cause the trail to be
+// flushed.
```
bossmc: Nice!

sprout/sproutletproxy.cpp
```diff
@@ -449,6 +449,21 @@ SproutletProxy::UASTsx::~UASTsx()
   }
   _timers.clear();
 
+  if (_trail != 0)
+  {
+    // Flush the trail so it appears promptly in SAS. Note that we also log an
+    // end marker when the transaction moves into completed state, which also
+    // flushes the trail (see on_tsx_complete). We need to log a flush marker
```
bossmc: Could we `0` the `_trail` in `on_tsx_complete`?   Or are you worried that someone might try to raise events/markers after that (in which case, this flush would also be a good idea)?

bossmc:
Looks nice, one question.

sprout/sproutletproxy.cpp
AlexHockey: Logging an extra flush marker isn't an issue, but the risk of dropping some events could be really bad, so I'd rather not null out the trail ID.

PR #1256, opened 2015-11-11
Pass memento notify URL to memento

mbt-msw:
Implements configuration of https://github.com/Metaswitch/memento/pull/105

The URL to notify is configured in the `memento_notify_url` setting in shared config.


AlexHockey:
Looks good. Please could you update https://github.com/Metaswitch/clearwater-readthedocs/blob/master/docs/Clearwater_Configuration_Options_Reference.md to document the new config option. It is probably an "Advanced option".

PR #1257, opened 2015-11-13
[Reviewer: Alex] Just talk to the memcached proxy on localhost, and let that handle GR

rkd-msw:
This:

* sets up a single Memcached connection to the proxy on localhost
* removes the whole idea of a "remote store" or a "backup store"

I've tested live - I can successfully register a subscriber (using a normal memcached instance on 127.0.0.1 - which should have the same interface as the proxy).
Passes UT and code coverage. (Some of the UTs for remote store code also hit necessary code coverage paths, so I've repurposed them for code coverage.)

### DIFF
--- a/include/bono.h
+++ b/include/bono.h
@@ -355,9 +355,7 @@ private:
   static const int LIVENESS_TIMER = 1;
 };
 
-pj_status_t init_stateful_proxy(RegStore* registrar_store,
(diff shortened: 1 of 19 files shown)


AlexHockey:
We agreed to put this on hold as the current thinking is we will not remove the GR code from sprout after all.

PR #1264, opened 2015-11-17
Reviewer AMC: Move sprout/sprout to sprout/src

graemerobertson:
I've run `make` in the top level directory and `make` in src and `make test` in src


bossmc:
Looks like fun. :+1:

PR #1265, opened 2015-11-18
[Reviewer: Rob] Fix up UT failure

eleanor-merry:
I've occasionally seen the RequestComplete UT fail as it unexpectedly updates its rate. I think this is because the load monitor is created before time is controlled - this PR fixes this.


rkd-msw:
LGTM.

Issue #1251, opened 2015-11-05
Our initial incoming request rate is probably a bad estimate


rkd-msw:
https://github.com/Metaswitch/sprout/blob/dev/sprout/main.cpp#L1334 sets our initial permitted request rate to 100 requests/sec. This is probably too low: assuming metrics of a re-REGISTER (2 requests) every 5 minutes and a call (~5 requests?) every 30 minutes per subscriber, each subscriber generates about 0.01 requests/second, so we can only handle ~10k subscribers with 100 requests/sec.

Previously, https://github.com/Metaswitch/cpp-common/issues/384 meant that that would creep up from 100 to a higher value, but we've fixed that issue now. To make sure we don't have load problems at start-of-day, we should:

- do some stress testing and figure out how many allowed requests we converge on on a single-core system
- set the initial rate value, as a Sprout command-line argument, to that figure times the number of cores (or maybe 75% of that value, to allow for perf fluctuations/different clock speeds)

rkd-msw:
From testing REGISTER traffic (probably a higher rate of work per request than calls) against an m3.medium Sprout node:

```
18-11-2015 17:39:04.548 UTC Status load_monitor.cpp:251: Maximum incoming request rate/second increased to 594.927612 (based on a smoothed mean latency of 3294 and 0 upstream overload responses)
18-11-2015 17:39:07.550 UTC Status load_monitor.cpp:251: Maximum incoming request rate/second increased to 604.551331 (based on a smoothed mean latency of 3763 and 0 upstream overload responses)
18-11-2015 17:39:10.550 UTC Status load_monitor.cpp:251: Maximum incoming request rate/second increased to 614.146301 (based on a smoothed mean latency of 4050 and 0 upstream overload responses)
18-11-2015 17:39:13.549 UTC Status load_monitor.cpp:251: Maximum incoming request rate/second increased to 623.878601 (based on a smoothed mean latency of 2677 and 0 upstream overload responses)
18-11-2015 17:39:18.547 UTC Status load_monitor.cpp:251: Maximum incoming request rate/second increased to 633.478882 (based on a smoothed mean latency of 3997 and 0 upstream overload responses)
18-11-2015 17:39:21.548 UTC Status load_monitor.cpp:251: Maximum incoming request rate/second increased to 643.256958 (based on a smoothed mean latency of 2219 and 0 upstream overload responses)
[...]

PR #1266, opened 2015-11-19
[Reviewer: Alex] Default the initial token rate to 250 * num cores

rkd-msw:
In https://github.com/Metaswitch/cpp-common/pull/397, we discussed that our initial token rate was probably too low.

I've now taken measurements suggesting 250 req/sec/core is sensible (and an underestimate, if anything) - see https://github.com/Metaswitch/sprout/issues/1251#issuecomment-157803199. This sets it to that.

### DIFF
--- a/sprout-base.root/etc/init.d/sprout
+++ b/sprout-base.root/etc/init.d/sprout
@@ -102,7 +102,6 @@ get_settings()
         default_session_expires=600
         signaling_dns_server=127.0.0.1
         chronos_hostname=127.0.0.1:7253
-        . /etc/clearwater/config
 
         # Set up a default cluster_settings file if it does not exist.  The local
         # IP address needs to be surrounded by square brackets if it is IPv6.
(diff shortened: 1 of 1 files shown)


sprout-base.root/etc/init.d/sprout
```diff
@@ -126,6 +126,11 @@ get_settings()
         num_pjsip_threads=1
         num_worker_threads=$(($(grep processor /proc/cpuinfo | wc -l) * 50))
         num_http_threads=$(($(grep processor /proc/cpuinfo | wc -l) * 50))
+
+        # Testing has shown that we can handle at least 250 requests per second per core - this is
+        # probably an underestimate, but is a reasonable start.
+        init_token_rate=$(($(grep processor /proc/cpuinfo | wc -l) * 250))
+
```
AlexHockey: I think this overwrites the value on `/etc/clearwater/shared_config`.

PR #1268, opened 2015-11-20
[Reviewer: Alex] Fix running without clearwater-cluster-manager

mirw:
Alex,

Please can you review this fix to https://github.com/Metaswitch/sprout/commit/b3eb91d98e6cb517b83478fcdec1e9eeff0a24ad (which you reviewed yesterday)?

Basically, by moving dotting in `/etc/clearwater/config` down, we had no longer set up `local_ip` before we tried to use it - this meant that our `/etc/clearwater/cluster_settings` file was broken if we were running without `clearwater-cluster-manager` (which would otherwise set up this file).

Not yet live-tested - will do so on Docker once the build completes.

Matt

### DIFF
--- a/sprout-base.root/etc/init.d/sprout
+++ b/sprout-base.root/etc/init.d/sprout
@@ -103,23 +103,6 @@ get_settings()
         signaling_dns_server=127.0.0.1
         chronos_hostname=127.0.0.1:7253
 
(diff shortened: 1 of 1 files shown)


AlexHockey:
Looks good. I'm happy with your plan for testing.

PR #1270, opened 2015-12-01
[Reviewer: Andy] Fix authentication crash

rkd-msw:
A Sprout node got a re-REGISTER with the following Authorization header:

`Authorization: Digest username="6510803201@great-lakes2.datcon.co.uk",realm="great-lakes2.datcon.co.uk",nonce="1ecf59a675a7162d",uri="sip:great-lakes2.datcon.co.uk",response="5e279ba845e1a866d8fc87080165091d",cnonce="44448888",opaque="222481f80f449504",qop=auth,nc=00000002`

This crashed Sprout, because:

- there was a response, so we tried to challenge it
- there was an expired nonce, so we got NULL as our AV
- there was no explicit algorithm, so we tried to call `av->HasMember("digest")`

This fixes that up (defaulting to Digest in that case). I confirmed my UT reproduces the crash.

### DIFF
--- a/src/authentication.cpp
+++ b/src/authentication.cpp
@@ -710,20 +710,22 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
       {
         auth_stats_table = auth_stats_tables->ims_aka_auth_tbl;
       }
-    }
-    if (auth_stats_table == NULL)
-    {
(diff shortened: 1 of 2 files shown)


bossmc:
Looks good.

Issue #1274, opened 2015-12-02
Perimeta Rejects Notify from Sprout with an Apparent Invalid AOR 


dcollins991:
For the "REGISTER - Network Dereg" test scenario, a UE registers successfully(300 expiry) and then SUBSCRIBES to registration events.  The UE receives a NOIFY and then sends a 200 OK.  The UE then waits 5 mutes and waits on a NOTIFY.  After 5 minutes Sprout issues a NOTIFY and Perimeta rejects with a 403 stating, 
"A call to an unknown subscriber was rejected because the adjacency does 
not permit such calls. Check the public ID and local ID to see if an 
attempt to register has been rejected. 
Public ID = sip:7030000010@mvoip.richlab.datcon.co.uk 
Local ID = sip:quaff@192.168.162.158:32144;transport=udp;ob"
Not sure if this is the root cause of the issue but the aor is invalid and appears as, 
registration aor="aor" id="209243057466376192" state="active"  or, 
registration aor="aor" id="209243057466376192" state="terminated"
Sprout earlier reported an AOR in the SAS log of, 
RegStore SET for AoR: "sip:7030000010@mvoip.richlab.datcon.co.uk" succeeded.  
None of the AOR values in the Notify xml are properly set.  In Release Rob Roy it was set in the register  Notify but not in the dereg NOTIFY.

I have SAS traces and diags.  Please clarify as to whether Sprout is expected to support the registration aor field.

mirw:
@eleanor-merry, I think you've done some work in this area - please can you take a look?

rkd-msw:
I've taken a look at this.

* The AOR should indeed be "sip:7030000010@mvoip.richlab.datcon.co.uk" rather than "aor", and this looks like a small fix.
* However, I think this is a longstanding bug and not a regression (though the issue suggests otherwise, at least in some cases) - so I don't think this is the root cause. I wonder if in fact the problem is that we're trying to send a NOTIFY to a UE saying that that UE's registration has expired (which is obviously doomed).

@dcollins991 - I can't see an issue1274 folder with diags in the usual place. can you upload them? If possible, can you include a trace from the Rob Roy release, so I can diagnose the differences you're reporting in that release?

Can you also confirm if you have registration monitoring on in the P-CSCF or not?

dcollins991:
[...]

PR #1275, opened 2015-12-03
[Reviewer: Ellie] Fix subscriptions for expired users

rkd-msw:
This fixes the two issues reported in #1274:

* Use the actual AoR instead of the literal string 'aor'
* When binding A expires, don't send a NOTIFY to binding A announcing the subscription is terminated - that's pointless

I think the UTs cover all the right behaviour here - I don't think there's much value in testing live.


rkd-msw:
@eleanor-merry , ping?

eleanor-merry:
LGTM

PR #1278, opened 2015-12-09
Reviewer Matt - Don't clean FV tests when doing top-level clean

bossmc:
This is a holding pattern fix (while I determine what we do with the code in `tests`).  Prevents errors when running `make clean` at the top level of Sprout's codebase.


mirw:
Makefile looks makefile.

PR #1282, opened 2015-12-10
Update gcovr

dimbleby:


dimbleby:
Hold on - I think that we can avoid a gcovr update after all...

build-infra pull to follow instead.

PR #1283, opened 2015-12-11
[Reviewer: Andy] Add performance test

mirw:
Andy,

Please can you review my changes to add performance testing?  Most of this is just refactoring (and only worth skim-reading) - key points are

*   new sip-perf.xml SIPp scenario
*   clearwater-sip-stress package refactored into clearwater-sipp base and clearwater-sip-stress scripts
*   new clearwater-sip-perf package containing sip-perf.xml SIPp scenario.

All live-tested.

Thanks,

Matt


bossmc:
Looks good to me.

PR #1286, opened 2015-12-17
[Reviewer AMC] If a route header is present, only route to a sproutlet if this is NOT a SUBSCRIBE

steelyvoid:
This fixes a SUBSCRIBE routing issue whereby, if I-CSCF identifies the target S-CSCF for a SUBSCRIBE as the local Sprout, it routes the request to the S-CSCF Sproutlet, thereby bypassing the subscription module and causing the SUBSCRIBE to be routed (erroneously) to the subscribee instead of being processed on the local Sprout

This fix stops the I-CSCF from attempting to use sproutlets, instead sending it to the local S-CSCF the long way round (so that it is sent to the subscription module correctly by PJSIP).

Tested live using a P-CSCF configuration that sent re-event SUBSCRIBEs to the I-CSCF and sending in reg-event subscribes for local subscribers.

### DIFF
--- a/src/sproutletproxy.cpp
+++ b/src/sproutletproxy.cpp
@@ -123,7 +123,9 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
   }
   else
   {
-    if (PJSIP_URI_SCHEME_IS_SIP(route->name_addr.uri))
+    // TODO: Again, needs to change once the subscription manager is a sproutlet.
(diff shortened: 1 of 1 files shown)


bossmc:
Hmmm, CI shows that this change has broken non-`reg-event` subscription (in particular end-to-end subscriptions loop forever since they can't be handled by the subscription module and this new code means they can't be handled by a sproutlet),  I'm reverting this change for now until we come up with a solution that works.

PR #1293, opened 2016-01-05
[Reviewer: Rob] Omit trailing --non-registering-pbxes parameter

mirw:
Minor tweak - just to reduce confusion with the trailing --non-registering-pbxes parameter.

Tested live.

Matt


rkd-msw:
+1

PR #1295, opened 2016-01-06
Reviewer AJH: 3-site GR memcached client changes

graemerobertson:
Hi Alex,

Please can you review my changes to move Sprout over to use the new memcached architecture, and to enable 3-site GR.

Note that there are still a couple of spec issues to resolve, but I don't think that affects your ability to review this code - but I haven't dropped them.

The change to the registrar module, subscription module and the handlers ***should*** be identical so this PR looks bigger than it is. However, the UTs only really test the handlers version of the code (since they were basically already in place).

I haven't live tested yet. This also isn't ready to be merged into dev.

Apologies about the commit history :( :(

Thanks,
Graeme


include/cfgoptions.h
```diff
@@ -94,8 +94,8 @@ struct options
   std::string                          hss_server;
   std::string                          xdm_server;
   std::string                          chronos_service;
-  std::string                          store_servers;
-  std::string                          remote_store_servers;
+  std::string                          registration_store;
+  std::vector<std::string>             remote_registration_stores;
```
AlexHockey: Should be updated if we change the name of the config variables / command line parameters.

src/handlers.cpp
```diff
@@ -58,7 +58,7 @@ static bool sdm_access_common(SubscriberDataManager::AoRPair** aor_pair,
                               bool& previous_aor_pair_alloced,
                               std::string aor_id,
                               SubscriberDataManager* current_sdm,
-                              SubscriberDataManager* remote_sdm,
+                              std::vector<SubscriberDataManager*> remote_sdms,
                               SubscriberDataManager::AoRPair** previous_aor_pair,
                               SAS::TrailId trail)
 {
```
AlexHockey: We discussed this at length. I think that this function should be changed so that the only AoR it passes out is in `aor_pair`. This means that `previous_aor_pair` becomes a single pointer, we can bin `previous_aor_pair_allocated`, and simplify the code in this function significantly.

AlexHockey: It's also worth a comment that either `previous_aor_pair` must be NULL, or `remote_sdms` must be empty.

src/main.cpp
```diff
@@ -651,13 +653,32 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
...
-      break;
+      {
+        std::string stores_str = std::string(pj_optarg);
+        std::stringstream ss(stores_str);
+        std::istream_iterator<std::string> begin(ss);
+        std::istream_iterator<std::string> end;
+        std::vector<std::string> stores_vector(begin, end);
+        std::copy(stores_vector.begin(),
+                  stores_vector.end(),
+                  std::ostream_iterator<std::string>(std::cout, ","));
```
AlexHockey: I have no idea what this code is trying to do.

src/main.cpp
```diff
@@ -651,13 +653,32 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
...
 
-    case 'm':
-      options->remote_store_servers = std::string(pj_optarg);
-      TRC_INFO("Using remote memcached store with configuration file %s", pj_optarg);
+          TRC_INFO("Using memcached registration stores %s", pj_optarg);
+        }
+        else
+        {
+          // I'm not sure we can hit this branch.
+          TRC_ERROR("Unable to parse registration-stores parameter");
```
AlexHockey: Can you work this out and delete the branch if not?

src/main.cpp
```diff
@@ -651,13 +653,32 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
...
+        if (!stores_vector.empty())
+        {
+          options->registration_store = stores_vector.front();
+          stores_vector.erase(stores_vector.begin());
+          options->remote_registration_stores = stores_vector;
 
-    case 'm':
-      options->remote_store_servers = std::string(pj_optarg);
-      TRC_INFO("Using remote memcached store with configuration file %s", pj_optarg);
+          TRC_INFO("Using memcached registration stores %s", pj_optarg);
```
AlexHockey: This log would more more useful if you output the result of the parse, e.g.

```
Primary store: example.com
Backup stores:
  - example2.com
  - exmaple3.com
```

src/main.cpp
```diff
@@ -1293,7 +1320,7 @@ LoadMonitor* load_monitor = NULL;
 HSSConnection* hss_connection = NULL;
 Store* local_data_store = NULL;
 SubscriberDataManager* local_sdm = NULL;
-SubscriberDataManager* remote_sdm = NULL;
+std::vector<SubscriberDataManager*> remote_sdms = {};
```
AlexHockey: FYI, the default constructor for a vector is for it to be empty, so no need to initialize here.

src/main.cpp
```diff
@@ -1935,38 +1964,33 @@ int main(int argc, char* argv[])
...
+      if (!opt.remote_registration_stores.empty())
       {
         // Use remote memcached store too.
-        TRC_STATUS("Using remote memcached compatible store with ASCII protocol");
-
-        remote_data_store = (Store*)new MemcachedStore(true,
-                                                       opt.remote_store_servers,
-                                                       memcached_remote_comm_monitor,
-                                                       remote_vbucket_alarm);
+        TRC_STATUS("Using remote memcached compatible stores with ASCII protocol");
```
AlexHockey: We use the binary protocol now.

src/registrar.cpp
```diff
@@ -257,22 +258,49 @@ SubscriberDataManager::AoRPair* write_to_store(
...
-          (backup_sdm->has_servers()))
-      {
-        backup_aor = backup_sdm->get_aor_data(aor, trail);
-        backup_aor_alloced = (backup_aor != NULL);
-      }
+      bool found_binding = false;
 
       if ((backup_aor != NULL) &&
           (backup_aor->get_current() != NULL) &&
           (!backup_aor->get_current()->bindings().empty()))
```
AlexHockey: Worth putting these three conditions into a sub-function.

src/registrar.cpp
```diff
@@ -257,22 +258,49 @@ SubscriberDataManager::AoRPair* write_to_store(
...
       {
+        found_binding = true;
+      }
+      else
+      {
+        std::vector<SubscriberDataManager*>::iterator it = backup_sdms.begin();
+
+        while ((it != backup_sdms.end()) && (!found_binding))
+        {
+          if (backup_aor_alloced)
```
AlexHockey: This flag feels a bit messy. I can think of two ways to tidy it up:
* Instead of having a flag that relates to another variable, just have a `local_backup_aor` variable, that is initialized to NULL, and unconditionally freed when it is overwritten and at the end of the function . `backup_aor` gets set to `local_backup_aor` when want to use it going forward. 

* Subfunction the logic to copy the bindings and subscriptions. This means you could have two branches - one for is a backup AoR is passed in, and one where you have to look one up. In the latter branch you can free the backup AoR as soon as you've copied stuff out of it.

src/subscription.cpp
```diff
@@ -176,23 +177,62 @@ SubscriberDataManager::AoRPair* write_subscriptions_to_store(
...
+            else
+            {
+              ++it;
+            }
+          }
+        }
+      }
+
+      if (found_subscription)
+      {
```
AlexHockey: Same comment about tidyness as above.

AlexHockey:
Generally looks good. You we're working in some difficult code, and I think we should tidy it up rather than let it degrade further. That is what most of my comments are about. 

The unit tests look good. Should we have a testcase for when remote site 1 doesn;t have an AoR but remote site 2 does? 

Remember to also update the config options reference when you're done with all the changes.

PR #1297, opened 2016-01-12
[Reviewer: Andy] Use ENT logs in CM

eleanor-merry:
Corresponding changes to https://github.com/Metaswitch/cpp-common/pull/441. If these are OK, I'll make the same changes to the other repos

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1691,33 +1691,33 @@ int main(int argc, char* argv[])
 
     chronos_comm_monitor = new CommunicationMonitor(new Alarm("sprout", AlarmDef::SPROUT_CHRONOS_COMM_ERROR,
                                                                         AlarmDef::MAJOR),
-                                                              "Sprout",
-                                                              "Chronos");
+                                                    "Sprout",
+                                                    "Chronos");
 
     enum_comm_monitor = new CommunicationMonitor(new Alarm("sprout", AlarmDef::SPROUT_ENUM_COMM_ERROR,
                                                                      AlarmDef::MAJOR),
-                                                           "Sprout",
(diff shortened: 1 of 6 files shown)


src/main.cpp
```diff
@@ -1690,22 +1690,34 @@ int main(int argc, char* argv[])
     // Create Sprout's alarm objects.
 
     chronos_comm_monitor = new CommunicationMonitor(new Alarm("sprout", AlarmDef::SPROUT_CHRONOS_COMM_ERROR,
-                                                                        AlarmDef::MAJOR));
+                                                                        AlarmDef::MAJOR),
+                                                              "Sprout",
+                                                              "Chronos");
```
bossmc: Indent fail (and following).

src/ut/communicationmonitor_test.cpp
```diff
@@ -92,7 +91,7 @@ TEST_F(CommunicationMonitorTest, UpdateNoAlarmSuccessCleared)
 TEST_F(CommunicationMonitorTest, UpdateNoAlarmFailureSet)
 {
   cwtest_advance_time_ms(31000);
-  EXPECT_CALL(*_ma, alarmed()).Times(2).WillRepeatedly(Return(true));
+  EXPECT_CALL(*_ma, alarmed()).WillOnce(Return(true));
```
bossmc: I'm not a big fan of this, it's not mandated that the implementation calls `alarmed()` only once, just that it doesn't call `set()` or `clear()`... Can't you set up a `WillRepeatedly` on `alarmed()`?

bossmc:
Fairly uninteresting

PR #1298, opened 2016-01-12
Reviewer AJH: Use specced config options

graemerobertson:
Hi Alex,

Please can you review the changes I've made to Sprout to address the following:
* The AV store can be configured separately to the registration store.
* sprout_registation_store is of the format <site_name>=<domain>,<site_name>=<domain>,...

I'm looking to merge this into my 3site_GR branch. I haven't tested any of that yet, but I'll test all the different combinations of configuration options when I do that.

I'll port the relevant changes over to Ralf once you've reviewed.

Thanks,
Graeme


src/main.cpp
```diff
@@ -1599,6 +1612,104 @@ int main(int argc, char* argv[])
...
+          // This is the local site's store.
+          registration_store_location = store;
+        }
+        else
+        {
+          // A remote site store.
+          remote_registration_stores_locations.push_back(store);
+        }
+      }
+    }
```
AlexHockey: I don;t care too much, but the cases with "site1=foo.com", seems like a degenerate case of "site1=foo.com,site2=bar.com" and could be handled with the same code.

AlexHockey: Aren't we going to need the same logic to parse config options in ralf and cedar. Should this live in cpp-common?

src/main.cpp
```diff
@@ -1599,6 +1612,104 @@ int main(int argc, char* argv[])
...
+    else if (opt.av_store.substr(0, pos) == opt.local_site_name)
+    {
+      av_store_location = opt.av_store.substr(pos+1);
+    }
+    else
+    {
+      // No local site AV store, so Sprout has been misconfigured.
+      TRC_ERROR("No local site AV store specified");
+      return 1;
+    }
```
AlexHockey: Likewise, this block of code is very similar to the bock above and could be subfunctioned.

src/main.cpp
```diff
@@ -1599,6 +1612,104 @@ int main(int argc, char* argv[])
...
+  }
+
+  if (av_store_location != "")
+  {
+    TRC_INFO("Using memcached AV store %s", av_store_location.c_str());
+  }
```
AlexHockey: What if no AV store is configured. We still use an AV store, so it would be nice to have a log in that case too.

src/main.cpp
```diff
@@ -2059,7 +2170,20 @@ int main(int argc, char* argv[])
...
       // relevant challenge is sent.
       TRC_STATUS("Initialise S-CSCF authentication module");
-      av_store = new AvStore(local_data_store);
+
+      if (av_store_location != "")
+      {
+        av_memstore = (Store*)new TopologyNeutralMemcachedStore(av_store_location,
+                                                                astaire_resolver,
+                                                                memcached_comm_monitor);
+        av_store = new AvStore(av_memstore);
```
AlexHockey: It seems slightly odd to me that we can only create an AV store if a reg store was configured, but I suppose the code has always been like that.

src/main.cpp
```diff
@@ -2309,6 +2433,10 @@ int main(int argc, char* argv[])
     delete *it;
   }
   remote_data_stores.clear();
+  if (av_memstore != NULL)
+  {
+    delete av_memstore;
+  }
```
AlexHockey: This is unnecessary - `delete NULL` is safe.

AlexHockey:
Looks good, most of my comments are about sub-functioning repeated code.

src/main.cpp
graemerobertson: We do, later. But it makes more sense to have both logs here so I'll do that.

PR #1302, opened 2016-01-15
[Reviewer: Chris] Fixes Iss1140: Sprout adds wrong header field parameter for orig-cdiv

AME2:


chris-elford-metaswitch:
This looks good to me. Only one tweak required to the code, then this is good to go.

PR #1303, opened 2016-01-18
[Reviewer: Ellie] Astaire alarm log updates

johadalin:
I'd particularly like you to look at the UTs. I've completely replaced what was there before. The diff below makes it look like i edited them about, but they are just all new, trying to test the error states rather than using the 'alarmed' check.


johadalin:
Cpp- common changes are here: https://github.com/Metaswitch/cpp-common/pull/444

johadalin:
Astaire changes here: https://github.com/Metaswitch/astaire/pull/44

src/main.cpp
```diff
@@ -2076,9 +2062,8 @@ int main(int argc, char* argv[])
 
     if (local_data_store == NULL)
     {
-      CL_SPROUT_MEMCACHE_CONN_FAIL.log();
```
eleanor-merry: What is this condition? Why doesn't it get an ENT log?

src/ut/communicationmonitor_test.cpp
```diff
@@ -68,54 +68,129 @@ class CommunicationMonitorTest : public ::testing::Test
   CommunicationMonitor _cm;
 };
 
-TEST_F(CommunicationMonitorTest, SuccessNoUpdate)
+// State Definitions:
+//
+// NO_ERRORS == No failures, at least one success
```
eleanor-merry: Can you explain more about what the test is doing here, rather than using the state definitions

eleanor-merry:
Looks good.

PR #1307, opened 2016-01-20
[Reviewer: Rob] Fixes iss1185: UTs now clean up

AME2:


src/ut/basetest.hpp
```diff
@@ -45,11 +45,16 @@
 #include "fakelogger.h"
 #include "stack.h"
 #include "snmp_event_accumulator_table.h"
+#include "statistic.h"
+#include "zmq_lvc.h"
+
+#include "basetest.hpp"
```
rkd-msw: Should you be including basetest.hpp in itself?

src/ut/basicproxy_test.cpp
```diff
@@ -3691,6 +3691,9 @@ TEST_F(BasicProxyTest, BlacklistOnTimeout)
 // out.
 TEST_F(BasicProxyTest, StatelessProxyNoBlacklistOnTimeout)
 {
+  // This server doesn't respond, so advance time to trigger the timeout.
```
rkd-msw: You should remove this code rather than commenting it out.

src/ut/basetest.cpp
```diff
@@ -1,70 +0,0 @@
...
- * software and licensed under the OpenSSL Licenses.
- * "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
- * under which the OpenSSL Project distributes the OpenSSL toolkit software,
- * as those licenses appear in the file LICENSE-OPENSSL.
- */
-
-///
-///----------------------------------------------------------------------------
-
-#include "gmock/gmock.h"
```
rkd-msw: On reflection I think we should keep the separate base_test.cpp, as it makes compilation faster - the destructor code only needs to be compiled once, not once for every file that includes the header. (See https://github.com/Metaswitch/homestead/pull/150 for more information.)

src/ut/basetest.cpp
```diff
@@ -1,70 +0,0 @@
...
-{
-  stack_data.stats_aggregator = new LastValueCache(num_known_stats,
-                                                   known_statnames,
-                                                   "6666",
-                                                   10);  // Short period to reduce shutdown delays.
-}
-
-BaseTest::~BaseTest()
-{
-  // Destroy the LVC before calling cw_reset_time, otherwise ZeroMQ
```
rkd-msw: Worth noting that this comment pinpoints the issue we spent a while tracking down earlier - spotting this then might have saved a little time.

src/ut/basetest.hpp
```diff
@@ -45,11 +45,16 @@
 #include "fakelogger.h"
 #include "stack.h"
 #include "snmp_event_accumulator_table.h"
+#include "statistic.h"
+#include "zmq_lvc.h"
```
rkd-msw: Because you've removed the call to `new LastValueCache`, you don't need to include `zmq_lvc.h` in any of these files.

rkd-msw:
A few comments, otherwise looks good.

As a general point it's worth saying more in your pull requests - someone looking at this in a few months won't understand why you've removed the `LastValueCache` objects or `basetest.cpp` in a pull request that says "UTs now clean up".

AME2:
One of the issues controlling time caused in the Unit Tests was whenever a new LastValueCache object was created (in both the basetest and siptest classes) the zmq_poll function would wait until timeout, so if a UT had advanced time zmq_poll would wait this amount of time.
To fix this I removed the `LastValueCache` objects.

src/Makefile
```diff
@@ -108,7 +108,6 @@ sprout_test_SOURCES := ${SPROUT_COMMON_SOURCES} \
                        faketransport_tcp.cpp \
                        fakednsresolver.cpp \
                        fakechronosconnection.cpp \
-                       basetest.cpp \
```
rkd-msw: You'll want to reinstate this.

rkd-msw:
One final comment, then good to merge.

(Another point to note here is that the LastValueCache objects weren't used in the tests - they're a remnant of old-style statistics - which is why it was sensible to remove them.)

PR #1308, opened 2016-01-20
[Reviewer: Chris] Stats api update

johadalin:
This is the bulk of the work. 

I've added the AoR class methods for getting the number of bindings and subscriptions, and the next expiry time, as we decided. 
The SubscriberDataManager, and SubscriberDataManagerTests are where the biggest changes are. 
I'd particularly appreciate a look at the send_timers logic, and the tests associated with sending timers that i've added, to make sure i'm not missing some areas of testing.

There's also a number of minor changes throughout:
- changed RegSubTimeoutTask to AoRTimeoutTask
- made modifications to support the map of tags throughout.
- cleared out a couple of bits of whitespace i saw Andrew introduced in scscf-sproutlet code.

Let me know if anything isn't clear. Feedback as soon as is plausible would be good so i can try to get it in this week (assuming there aren't major markups).
I'll begin live testing it early tomorrow, at least initial tests, and will let you know if things aren't working and i'm going to have to make more code changes.

### DIFF
--- a/include/subscriber_data_manager.h
+++ b/include/subscriber_data_manager.h
@@ -202,10 +202,10 @@ public:
     inline const Subscriptions& subscriptions() const { return _subscriptions; }
 
     // Return the number of bindings in the AoR.
-    uint32_t get_bindings_count() { return _bindings.size(); }
+    inline uint32_t get_bindings_count() const { return _bindings.size(); }
 
     // Return the number of subscriptions in the AoR.
-    uint32_t get_subscriptions_count() { return _subscriptions.size(); }
+    inline uint32_t get_subscriptions_count() const { return _subscriptions.size(); }
 
     // Return the expiry time of the binding or subscription due to expire next.
     int get_next_expires();
--- a/include/subscriber_data_manager.h
+++ b/include/subscriber_data_manager.h
@@ -201,6 +201,15 @@ public:
     /// Retrieve all the subscriptions.
     inline const Subscriptions& subscriptions() const { return _subscriptions; }
 
+    // Return the number of bindings in the AoR.
+    uint32_t get_bindings_count() { return _bindings.size(); }
+
+    // Return the number of subscriptions in the AoR.
+    uint32_t get_subscriptions_count() { return _subscriptions.size(); }
+
+    // Return the expiry time of the binding or subscription due to expire next.
+    int get_next_expires();
+
     /// CSeq value for event notifications for this AoR.  This is initialised
     /// to one when the AoR record is first set up and incremented every time
     /// the record is updated while there are active subscriptions.  (It is
@@ -209,6 +218,9 @@ public:
     /// CSeq=1, and once a subscription dialog is established it should
     /// receive every NOTIFY for the AoR.)
     int _notify_cseq;
+
+    // Chronos Timer ID
+    std::string _timer_id;
 
   private:
     /// Map holding the bindings for a particular AoR indexed by binding ID.
@@ -383,44 +395,22 @@ public:
   private:
     ChronosConnection* _chronos_conn;
 
+    /// Build the tag info map from an AoR
+    virtual void build_tag_info(AoR* aor,
+                                std::map<std::string, uint32_t>& tag_map);
+
     /// Create the Chronos Timer request
     ///
     /// @param aor_id       The AoR ID
-    /// @param object_id    What the JSON ID should be
     /// @param timer_id     The Timer ID
     /// @param expiry       Timer length
-    /// @param id_type      Binding/subscription ID
     /// @param tags         Any tags to add to the Chronos timer
     /// @param trail        SAS trail
     virtual void set_timer(const std::string& aor_id,
-                           std::string object_id,
                            std::string& timer_id,
                            int expiry,
-                           std::string id_type,
-                           std::vector<std::string> tags,
+                           std::map<std::string, uint32_t> tags,
                            SAS::TrailId trail);
-
-    /// Create and send any Chronos requests for bindings
-    ///
-    /// @param aor_id       The AoR ID
-    /// @param aor_pair     The AoR pair to send requests for
-    /// @param now          The current time
-    /// @param trail        SAS trail
-    virtual void send_binding_timers(const std::string& aor_id,
-                                     AoRPair* aor_pair,
-                                     int now,
-                                     SAS::TrailId trail);
-
-    /// Create and send any Chronos requests for subscriptions
-    ///
-    /// @param aor_id       The AoR ID
-    /// @param aor_pair     The AoR pair to send requests for
-    /// @param now          The current time
-    /// @param trail        SAS trail
-    virtual void send_subscription_timers(const std::string& aor_id,
-                                          AoRPair* aor_pair,
-                                          int now,
-                                          SAS::TrailId trail);
   };
 
   /// @class SubscriberDataManager::NotifySender
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -150,7 +150,7 @@ static void report_sip_all_register_marker(SAS::TrailId trail, std::string uri_s
 }
 
 //LCOV_EXCL_START - don't want to actually run the handlers in the UT
-void RegSubTimeoutTask::run()
+void AoRTimeoutTask::run()
 {
   if (_req.method() != htp_method_POST)
   {
@@ -251,7 +251,7 @@ void DeregistrationTask::run()
   delete this;
 }
 
-void RegSubTimeoutTask::handle_response()
+void AoRTimeoutTask::handle_response()
 {
   bool all_bindings_expired = false;
   SubscriberDataManager::AoRPair* aor_pair = set_aor_data(_cfg->_sdm,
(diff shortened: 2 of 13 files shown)


johadalin:
The CPP-common and ralf changes are at the PRs below, but the changes are so minor it isn't really worth reviewing assuming live testing works.
https://github.com/Metaswitch/cpp-common/pull/449
https://github.com/Metaswitch/ralf/pull/198

include/subscriber_data_manager.h
```diff
@@ -201,6 +201,15 @@ class SubscriberDataManager
     /// Retrieve all the subscriptions.
     inline const Subscriptions& subscriptions() const { return _subscriptions; }
 
+    // Return the number of bindings in the AoR.
+    uint32_t get_bindings_count() { return _bindings.size(); }
```
chris-elford-metaswitch: Should these functions be marked inline and const?

include/subscriber_data_manager.h
```diff
@@ -201,6 +201,15 @@ class SubscriberDataManager
     /// Retrieve all the subscriptions.
     inline const Subscriptions& subscriptions() const { return _subscriptions; }
 
+    // Return the number of bindings in the AoR.
```
chris-elford-metaswitch: At line 171, the function common_constructor looks like it should be private.

src/handlers.cpp
```diff
@@ -343,7 +343,7 @@ SubscriberDataManager::AoRPair* RegSubTimeoutTask::set_aor_data(
 }
 
 // Retrieve the aor and binding ID from the opaque data
```
chris-elford-metaswitch: This comment is now out of date.

src/subscriber_data_manager.cpp
```diff
@@ -731,6 +733,41 @@ std::string SubscriberDataManager::AoR::Binding::pub_gruu_quoted_string(pj_pool_
   return ret;
 }
 
+// Utility function to return the expiry time of the binding or
+// subscription due to expire next.
+int SubscriberDataManager::AoR::get_next_expires()
```
chris-elford-metaswitch: Surely get_next_expiry?

src/subscriber_data_manager.cpp
```diff
@@ -731,6 +733,41 @@ std::string SubscriberDataManager::AoR::Binding::pub_gruu_quoted_string(pj_pool_
   return ret;
 }
 
+// Utility function to return the expiry time of the binding or
+// subscription due to expire next.
```
chris-elford-metaswitch: It doesn't get the expiry time. It gets the time until expiry. Pedantic, but important.

src/subscriber_data_manager.cpp
```diff
@@ -731,6 +733,41 @@ std::string SubscriberDataManager::AoR::Binding::pub_gruu_quoted_string(pj_pool_
...
+    {
+      _next_expires = s->second->_expires;
+    }
+  }
+
+  // If nothing has altered the _next_expires, the AoR is empty and invalid.
+  // Return 0 to indicate there is nothing to expire.
+  if (_next_expires == INT_MAX)
+  {
+    return 0;
```
chris-elford-metaswitch: The fact that this function returns 0 when there is no binding or subscription is an important part of the interface of this function, so it at least needs to be documented in the comment at the top of the function.

chris-elford-metaswitch: You are going to add a log here to help debugging cases where this function is called incorrectly.

src/subscriber_data_manager.cpp
```diff
@@ -1213,29 +1264,68 @@ SubscriberDataManager::ChronosTimerRequestSender::~ChronosTimerRequestSender()
...
+  std::map<std::string, uint32_t> old_tags;
+  std::map<std::string, uint32_t> new_tags;
+
+  build_tag_info(aor_pair->get_orig(), old_tags);
+  build_tag_info(aor_pair->get_current(), new_tags);
+
+  std::string& timer_id = aor_pair->get_current()->_timer_id;
+  // An AoR with no bindings is invalid, and the timer should be deleted.
+  // We do this before getting next_expires to save on processing.
+  if (new_tags["BIND"] == 0)
```
chris-elford-metaswitch: This is a bit roundabout. You want to know whether there are any bindings in the AoR, so rather than asking it directly, you look inside the data prepared for transmission to Chronos and (relying on your knowledge of the Chronos interface) pull out the relevant data.

I think you've done this to avoid a couple of function calls, but this feels like premature optimization to me. You've optimized for speed at the cost of reduced readability and a more leaky abstraction.

If you are worried about the function call overhead, you can make the binding_count function inline.
If you are worried about de-referencing aor_pair too many times, you can store the current AoR pointer in a local variable.

src/subscriber_data_manager.cpp
```diff
@@ -1213,29 +1264,68 @@ SubscriberDataManager::ChronosTimerRequestSender::~ChronosTimerRequestSender()
...
+  {
+    if (timer_id != "")
+    {
+      _chronos_conn->send_delete(timer_id, trail);
+    }
+    return;
+  }
+
+  int old_next_expires = aor_pair->get_orig()->get_next_expires();
+  int new_next_expires = aor_pair->get_current()->get_next_expires();
```
chris-elford-metaswitch: In fact, given how many times you call get_orig and get_current, I'd be inclined to use local variables for both of these pointers.

src/subscriber_data_manager.cpp
```diff
@@ -1213,29 +1264,68 @@ SubscriberDataManager::ChronosTimerRequestSender::~ChronosTimerRequestSender()
...
-                                    std::vector<std::string> tags,
+                                    std::map<std::string, uint32_t> tags,
                                     SAS::TrailId trail)
 {
   std::string temp_timer_id = "";
   HTTPCode status;
-  std::string opaque = "{\"aor_id\": \"" + aor_id + "\", \"" +
-                              id_type + "\": \""+ object_id +"\"}";
+  // Previous versions of handlers.cpp asserted timers contain a "binding_id".
+  // We no longer want this information, but it is included to simplify upgrade.
```
chris-elford-metaswitch: Do we have a plan for removing this in the long term. Can you document it somewhere?

src/subscriber_data_manager.cpp
```diff
@@ -1268,128 +1358,6 @@ void SubscriberDataManager::ChronosTimerRequestSender::set_timer(
...
-                s_id,
-                s->_timer_id,
-                s->_expires - now,
-                "subscription_id",
-                SubscriberDataManager::TAGS_SUB,
-                trail);
-    }
-  }
-}
-
```
chris-elford-metaswitch: I love massive sections of deleted code. This all looks a lot simpler now.

src/ut/chronosconnection_test.cpp
```diff
@@ -101,7 +101,7 @@ TEST_F(ChronosConnectionTest, SendPostWithTags)
   fakecurl_responses["http://10.42.42.42:80/timers"] = Response(headers);
 
   std::string opaque = "{\"aor_id\": \"aor_id\", \"binding_id\": \"binding_id\"}";
-  std::vector<std::string> tags {"TAG1", "TAG2"};
+  std::map<std::string, uint32_t> tags; tags["TAG1"]++; tags["TAG2"]++;
```
chris-elford-metaswitch: This is a little busy. I was surprised it worked, but I'm slowly getting my head around C++ :smile: .

chris-elford-metaswitch:
Is it worth having a unit test which at least checks that we don't barf when receiving a genuine binding ID in a timer?

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -747,7 +759,7 @@ TEST_F(SubscriberDataManagerChronosRequestsTest, SubscriptionTestsDeleteWithoutC
   b1->_cid = std::string("gfYHoZGaFaRNxhlV0WIwoS-f91NoJ2gq");
```
chris-elford-metaswitch: At 754, why de-reference this rather than using _store directly?

chris-elford-metaswitch:
Only minor feedback - I'd like to know what you do about the points where I'm not sure above. Otherwise, I don't need to see this again.

include/subscriber_data_manager.h
johadalin: No idea if it should. i'm going to leave it as it is for now, as i haven't changed that section.

src/subscriber_data_manager.cpp
johadalin: It returns the expiry time, which is stored in _expires in the bindings and subscriptions; the function therefore is getting the next _expires.

I have however seen that i was not setting the time to be relative rather than absolute in my previous code, so i've added a step just  before set_timer to subtract 'now' from the '_expires' value to give a relative time. This may well have been what threw you.

src/subscriber_data_manager.cpp
johadalin: I dont believe there is a place where we can log things to be done at later dates. It also greatly depends on the upgrade state and strategy of clearwater core. I think this is the best we can do for now.

src/ut/subscriber_data_manager_test.cpp
johadalin: I don't know. I was following the structure of the rest of the tests.

PR #1309, opened 2016-01-21
[Reviewer Andy] Add aka resync to sprout

### LINKED ISSUE #1299: SQN resync in AKA doesn't work
In order to successfully authenticate, a UE may need to resynchronize it's AKA sequence number with the HSS, the flow looks like:

* UE sends a standard initial REGISTER to the S-CSCF
* S-CSCF does a MAR/MAA with the HSS which provides an AKA challenge, containing an encoded SQN
* The S-CSCF sends back the challenge to the UE
* The UE spots that the SQN needs resyncing (it's too far out from the local SQN record)
* The UE sends a response register to the S-CSCF but includes an `auts` parameter encoding it's local sequence number
* The S-CSCF should forward this parameter to the HSS to complete the resync and get a new challenge which the UE should happily respond to.

Homestead has support for receiving the `auts` parameter over HTTP and passing it to the HSS, but Sprout's code that would send the `auts` parameter to Homestead (https://github.com/Metaswitch/sprout/blob/3dfa8cc68f54573419fec5b93f5dbe9deb382488/src/authentication.cpp#L792-L831) is only invoked if the UE is already authenticated (which is impossible since the SQN is out of sync).  The comment here suggests this is a DOS-protection decision, but it seems it's just incorrect.

Tom-Rees:
Fixes #1299  - Where we were rejecting resync responses instead of passing them through to the HSS.

Tested live.

Andy - Please could you review along with related Homestead PR.

Cheers,
Tom

### DIFF
--- a/src/authentication.cpp
+++ b/src/authentication.cpp
@@ -266,7 +266,7 @@ pj_status_t user_lookup(pj_pool_t *pool,
       std::string response = "";
       if (((*av)["aka"].HasMember("response")) &&
           ((*av)["aka"]["response"].IsString()) &&
-          auts_param != NULL)
+          auts_param == NULL)
       {
         response = (*av)["aka"]["response"].GetString();
       }
--- a/src/authentication.cpp
+++ b/src/authentication.cpp
@@ -60,6 +60,7 @@ extern "C" {
 #include "authentication.h"
 #include "avstore.h"
 #include "snmp_success_fail_count_table.h"
+#include "base64.h"
 
 
 //
@@ -814,6 +815,11 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
           TRC_DEBUG("AKA SQN resync request from UE");
           std::string auts = PJUtils::pj_str_to_string(&p->value);
           std::string nonce = PJUtils::pj_str_to_string(&credentials->nonce);
+
+          // Convert the auts and nonce to binary for manipulation
+          nonce = base64_decode(nonce);
+          auts  = base64_decode(auts);
+
           if ((auts.length() != 14) ||
               (nonce.length() != 32))
           {
@@ -828,8 +834,9 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
           {
             // auts and nonce are as expected, so create the resync string
             // that needs to be passed to the HSS, and act as if no
-            // authentication information was received.
-            resync = nonce.substr(0,16) + auts;
+            // authentication information was received. The resync string
+            // should be RAND || AUTS.
+            resync = base64_encode(reinterpret_cast<const unsigned char*>((nonce.substr(0, 16) + auts).c_str()), 30);
             status = PJSIP_EAUTHNOAUTH;
             sc = unauth_sc;
           }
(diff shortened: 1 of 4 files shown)


src/authentication.cpp
```diff
@@ -822,8 +834,9 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
           {
             // auts and nonce are as expected, so create the resync string
             // that needs to be passed to the HSS, and act as if no
-            // authentication information was received.
-            resync = nonce.substr(0,16) + auts;
+            // authentication information was received. The resync string
+            // should be RAND || AUTS.
+            resync = base64_encode(reinterpret_cast<const unsigned char*>((nonce.substr(0, 16) + auts).c_str()), 30);
```
bossmc: Sad that this needs such a mammoth cast :disappointed:...

bossmc: It would be nicer if we had a overloaded `base64_encode(const std::string&)` (one overload would just call into the other).

src/ut/authentication_test.cpp
```diff
@@ -296,25 +302,30 @@ std::string AuthenticationMessage::hash2str(md5_byte_t* hash)
...
 
-
-void AuthenticationMessage::calculate_digest_response()
+string AuthenticationMessage::calculate_digest_response(
+    string algorithm, bool force_aka,
+    string auth_user, string key,
+    string method, string uri,
+    string nonce, string nc,
+    string cnonce, string qop,
+    string auth_realm)
```
bossmc: Presumably most (if not all) of these strings can be `const string&`?

src/ut/authentication_test.cpp
```diff
@@ -1260,7 +1282,7 @@ TEST_F(AuthenticationTest, AKAAuthResyncSuccess)
   EXPECT_EQ(2,((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_AUTHENTICATION_STATS_TABLES.ims_aka_auth_tbl)->_attempts);
   EXPECT_EQ(2,((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_AUTHENTICATION_STATS_TABLES.ims_aka_auth_tbl)->_successes);
 
-  _hss_connection->delete_result("/impi/6505550001%40homedomain/av/aka?impu=sip%3A6505550001%40homedomain&autn=876543218765432132132132132132");
+  _hss_connection->delete_result("/impi/6505550001%40homedomain/av/aka?impu=sip%3A6505550001%40homedomain&autn=f3beb9e37db5f3beb9e37db5f3beb9e3df6d77db5df6d77db5df6d77db5d");
```
bossmc: `autn` doesn't exist any more...

bossmc:
Looks pretty reasonable, couple of comments inline.

src/authentication.cpp
```diff
@@ -255,12 +256,18 @@ pj_status_t user_lookup(pj_pool_t *pool,
     pj_strdup(pool, &cred_info->username, acc_name);
     if (av->HasMember("aka"))
     {
+      pjsip_param* auts_param = pjsip_param_find(&credentials->other_param,
+                                        &STR_AUTS);
```
bossmc: Indenting

Issue #1267, opened 2015-11-19
Adding a second P-Asserted-Identity isn't quite right


eleanor-merry:
We add a second P-Assserted-Identity header if:
* There's a single P-Asserted-Identity header that's a SIP URI
* The SIP URI has a Tel URI alias (determined at https://github.com/Metaswitch/sprout/blob/dev/src/scscfsproutlet.cpp#L1798).

This logic means that we only add the second header if the Tel URI exactly matches the SIP URI (including any +'s). This isn't right - the aliases don't have to exactly match. 

Instead we should:
* Do comparisons on pj_strs, not strings (so we can use the URIclassifier and get the global enforce flag for free)
* Check if there's an exact match in the alias list, and use that Tel URI if so
* If not, use the first Tel URI in the aliases list.

(We also add a SIP URI header if there's a single Tel URI P-Asserted-Identity header - this function doesn't need to change)

PR #1310, opened 2016-01-22
[Reviewer: Rob] Fixes iss1267: Adds a tel URI alias to the P-Asserted-Identity header

AME2:
Fixes iss1267 
https://github.com/Metaswitch/sprout/issues/1267
Instead of only adding a tel URI alias if it matches the user of the SIP URI, we now check if there is a tel URI alias whose username matches the SIP URI. If there is we add it to the P-Asserted-Identity header and if there isn't we add the first tel URI in the alias list.

### DIFF
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -1830,16 +1830,18 @@ void SCSCFSproutletTsx::add_second_p_a_i_hdr(pjsip_msg* msg)
       }
       else
       {
-        for (unsigned int i=0; i<_aliases.size(); i++)
+        for (std::vector<std::string>::iterator it = _aliases.begin();
+             it != _aliases.end();
+             ++it)
         {
-          std::string tel_URI = "tel:";
-          std::size_t found = _aliases[i].rfind(tel_URI.c_str(), 4); 
-          if (found!=std::string::npos)
+          std::string tel_URI_prefix = "tel:";
+          bool has_tel_prefix = ((*it).rfind(tel_URI_prefix.c_str(), 4) != std::string::npos); 
+          if (has_tel_prefix)
           {
-            TRC_DEBUG("Add second P-Assered Identity for %s", _aliases[i].c_str());
+            TRC_DEBUG("Add second P-Asserted Identity for %s", (*it).c_str());
             PJUtils::add_asserted_identity(msg,
                                            get_pool(msg),
-                                           _aliases[i],
+                                           *it,
                                            asserted_id->name_addr.display);
             break;
           }
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -1791,7 +1791,7 @@ void SCSCFSproutletTsx::add_second_p_a_i_hdr(pjsip_msg* msg)
     // header.
     return;
   }
-
+  
   // Look for P-Asserted-Identity header.
   pjsip_routing_hdr* asserted_id =
     (pjsip_routing_hdr*)pjsip_msg_find_hdr_by_name(msg,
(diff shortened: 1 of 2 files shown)


src/scscfsproutlet.cpp
```diff
@@ -1825,6 +1828,23 @@ void SCSCFSproutletTsx::add_second_p_a_i_hdr(pjsip_msg* msg)
                                        new_p_a_i_str,
                                        asserted_id->name_addr.display);
       }
+      else
+      {
+        for (unsigned int i=0; i<_aliases.size(); i++)
```
rkd-msw: We'd normally iterate over a vector using a C++ iterator - see https://github.com/Metaswitch/sprout/blob/dev/src/aschain.cpp#L267 for an example.

src/scscfsproutlet.cpp
```diff
@@ -1825,6 +1828,23 @@ void SCSCFSproutletTsx::add_second_p_a_i_hdr(pjsip_msg* msg)
...
+        {
+          std::string tel_URI = "tel:";
```
rkd-msw: Is it clearer to call this `tel_prefix` or `tel_uri_prefix`?

src/scscfsproutlet.cpp
```diff
@@ -1825,6 +1828,23 @@ void SCSCFSproutletTsx::add_second_p_a_i_hdr(pjsip_msg* msg)
...
+          std::size_t found = _aliases[i].rfind(tel_URI.c_str(), 4);
```
rkd-msw: I would probably write this as `bool has_tel_prefix = (_aliases[i].rfind(tel_prefix, 4) != std::string::npos)` - your if test then becomes `if (has_tel_prefix)` which makes the overall intent of the code clearer.

src/scscfsproutlet.cpp
```diff
@@ -1825,6 +1828,23 @@ void SCSCFSproutletTsx::add_second_p_a_i_hdr(pjsip_msg* msg)
...
       }
+      else
+      {
+        for (unsigned int i=0; i<_aliases.size(); i++)
+        {
+          std::string tel_URI = "tel:";
+          std::size_t found = _aliases[i].rfind(tel_URI.c_str(), 4); 
+          if (found!=std::string::npos)
+          {
+            TRC_DEBUG("Add second P-Assered Identity for %s", _aliases[i].c_str());
```
rkd-msw: Typo - "Assered" -> "Asserted"

src/ut/scscf_test.cpp
```diff
@@ -7056,6 +7056,25 @@ TEST_F(SCSCFTest, TestAddSecondTelPAIHdr)
   doSuccessfulFlow(msg, testing::MatchesRegex(".*wuntootreefower.*"), hdrs, false);
 }
 
+TEST_F(SCSCFTest, TestAddSecondTelPAIHdriWithAlias)
```
rkd-msw: I think there's a typo in the test name - an extra 'i' in between 'HdrWith'.

rkd-msw: It's generally good practice to add a short comment before a testcase, to make it clear what you're testing - in this case, that a tel: URI alias is added even when the username is different from the original SIP URI.

src/ut/scscf_test.cpp
```diff
@@ -7075,6 +7094,24 @@ TEST_F(SCSCFTest, TestAddSecondSIPPAIHdr)
   doSuccessfulFlow(msg, testing::MatchesRegex(".*wuntootreefower.*"), hdrs, false);
 }
 
+TEST_F(SCSCFTest, TestAddSecondSIPPAIHdrNoSIPUri)
```
rkd-msw: Same comment about having a comment before the test explaining its purpose - a sentence or two of normal writing is a lot better for that than a test name like `TestAddSecondSIPPAIHdrNoSIPUri`,

rkd-msw:
A couple of comments inline. One other thing is that I think it's worth having more testcases to cover the more complicated cases, in particular:

* when you have multiple TEL URI aliases, none of which share the same userpart as the SIP URI, the first one in the list should be used
* when you have multiple TEL URI aliases, one of which shares the same userpart as the SIP URI, it should be used even if it's not the first one in the list

AME2:
Applied mark-ups

rkd-msw:
Couple of minor comments, good to merge once those are fixed.

PR #1311, opened 2016-01-28
[Reviewer AMC] Add unit test for missing Content-Length header

Neil-Bruce:


src/ut/siptest.cpp
```diff
@@ -370,7 +370,14 @@ void SipTest::inject_msg(const string& msg, TransportFlow* tp)
   log_pjsip_buf(buf, rdata->pkt_info.packet, rdata->pkt_info.len);
   pj_size_t size_eaten = pjsip_tpmgr_receive_packet(rdata->tp_info.transport->tpmgr,
                                                     rdata);
-  EXPECT_EQ((pj_size_t)rdata->pkt_info.len, size_eaten);
+  if (expected == -99)
```
bossmc: Ugh, gross...

bossmc: In particular because this is the mainline... and there's only one off-mainline use.  I'd prefer to see a specific test (that's mostly a copy-paste of this function) rather than coding in edge-cases into the generic framework like this.

src/ut/common_sip_processing_test.cpp
```diff
@@ -192,6 +194,8 @@ class CommonProcessingTest : public SipTest
 
       string requri = target;
       string route = _route.empty() ? "" : _route + "\r\n";
+      char   length[128];
```
bossmc: `content_length`?

bossmc:
I like that you've written a UT for this.  I'm worried that embedding this behaviour as optional behaviour for the common functionality is going to make maintenance harder in the long term so I suggest just making a specific test for this specific issue.

Neil-Bruce:
Review markups applied, back to Andy for review.

src/ut/common_sip_processing_test.cpp
```diff
@@ -548,3 +552,17 @@ TEST_F(CommonProcessingTest, DeathTest_Message200FailsHealthCheck)
...
 }
+
+TEST_F(CommonProcessingTest, NoContentLengthDropped)
+{
+  // Tests that a malformed request with no content length is just dropped
+
+  // Inject a request with no content length header.
+  Message msg1;
+  msg1._contentlength = false;
+  inject_msg_failure(msg1.get_request(), _tp, -PJSIP_EMISSINGHDR);
```
bossmc: Much nicer, thanks!

PR #1312, opened 2016-02-02
[Reviewer AJH] Add unit test to cover failure to open log files.

chris-elford-metaswitch:
Uses cw_control_fopen to allow fopen failures to be simulated.
Adds test which uses this to force a logger to re-open its log file.


AlexHockey:
The test you've added looks good. It might also be worth testing:

* When monotonic time starts at 0 - make sure we try to open the file right away. 
* When monotonic time != wall time.

chris-elford-metaswitch:
I've added a test to cover starting before 5s have passed. I'm not convinced that the monotonic != wall test is worth adding so I haven't. Are you happy with this?

AlexHockey:
Yeah, OK.

Issue #1296, opened 2016-01-08
Bono doesn't reject REGISTERs correctly during quiescence


mirw:
I was trying to quiesce my Bono node to upgrade it.  Unfortunately, my Bono node was being being polled with UDP SIP REGISTERs (about once every 10s) by a malicious host out on the Internet.

There were two things wrong about its behavior.

*   It (correctly) returned a 305 Use Proxy response, but didn't specify its proxy URI in a Contact header - it should.
*   Each poll kept a UDP transport alive for a further few minutes, meaning that Bono never quiesced.

mirw:
The first bullet is covered by #480.  This just covers bullet 2 - keeping UDP transports alive.

eleanor-merry:
I've changed the bono restart script to use stop instead of quiesce (https://github.com/Metaswitch/sprout/pull/1313). When this bug is fixed the fixer should change this back

eleanor-merry:
When Bono quiesces it first disables the untrusted TCP port.

It then waits for all its flows (REGISTERs) to finish (which can take up to reg_max_expires). Any new REGISTERs it receives during this time it rejects with a 305, and Bono doesn't update any flows based on these REGISTERs. 

Bono then (only for flows using UDP) waits for 10 minutes to see if there's any further activity on a flow, and only then destroys the flow.

This means that the quiescing flows part takes around 15 minutes, but can take much longer if reg_max_expires is set higher than the default.

The next stage is that Bono disables the trusted TCP port, and waits for all its connections to be torn down. If there's only TCP traffic then this is quick, however if there's UDP traffic then as Bono's UDP ports are still open (both trusted and untrusted), then the connections don't get torn down as they're still in use. However after 30 seconds monit kills us as poll_bono fails three times in a row (as that's TCP to the trusted port).

I think therefore we shouldn't recommend using bono quiesce, as we can't control how long this will take (as reg_max_expires could be set to anything). The UDP ports still being open is a bit weird, but I'm not sure it's worth changing if we don't recommend bono quiesce.

PR #1313, opened 2016-02-02
[Reviewer: Graeme] Update bono_restart

eleanor-merry:
Use stop rather than quiesce to restart Bono (as this is bugged - https://github.com/Metaswitch/sprout/issues/1296)


graemerobertson:
Looks fine, but might be worth a comment linking to the issue, and also possibly an explicit note in sprout 1296 to change this back once it's fixed.

PR #1320, opened 2016-02-12
[Reviewer: Ellie] handle invalid session expires value better

johadalin:
Just want to make sure this isn't going to break things. 
I've run UTs and they all still pass

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -898,7 +898,7 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
       options->default_session_expires = atoi(pj_optarg);
       if (options->default_session_expires < 90)
       {
-        TRC_INFO("Error, invalid default session expires value %d. Minimum value is 90 seconds",
+        TRC_INFO("Error, invalid default session expires value %d. Using default value.",
                   options->default_session_expires);
         options->default_session_expires = 90;
       }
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -309,9 +309,9 @@ static void usage(void)
        "     --sub-max-expires <expiry>\n"
        "                            The maximum allowed subscription period (in seconds)\n"
        "     --default-session-expires <expiry>\n"
-       "                            The session expiry period to request (in seconds)\n"
+       "                            The session expiry period to request (in seconds. Min 90)\n"
```
eleanor-merry: Can you put the default value in as well?

src/main.cpp
```diff
@@ -894,12 +894,22 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
 
     case OPT_DEFAULT_SESSION_EXPIRES:
       options->default_session_expires = atoi(pj_optarg);
+      if (options->default_session_expires < 90)
+      {
+        TRC_INFO("Error, invalid default session expires value. Minimum value is 90 seconds");
```
eleanor-merry: Can you put the invalid value, and say that we're using 90 seconds instead?

eleanor-merry:
Looks good, I've put some picky comments on wording

Issue #1201, opened 2015-09-29
HSS deregisters due to Chronos timer pops are not logged to SAS


bossmc:
This is because we don't pass the trail ID to the HTTP stack (https://github.com/Metaswitch/sprout/blob/dev/sprout/handlers.cpp#L269).

Fairly annoying as it makes debugging registration timeouts harder.

rkd-msw:
OK, I've got them logged to SAS with #1329, but the detailed timeline view isn't very good, so I'm going to add a high-level SAS log before closing this.

![2016-02-23_20h22_42](https://cloud.githubusercontent.com/assets/4363492/13263802/81c69eaa-da6b-11e5-8ae4-fd4da96b5c18.png)

rkd-msw:
Now done.

PR #1329, opened 2016-02-22
[Reviewer: Alex] Pass trail ID on deregistration

rkd-msw:
This is a trivial change for #1201.

I think this is actually so trivial that I'm not proposing to test it before merging - I'll merge, then register a client on the VMware system and make it fail (i.e. kill my X-Lite process), and make sure I get a sensible SAS log.


AlexHockey:
Code looks good, and I'm happy with your test plan. 

(Note at first I didn't think this change would work as there might be nothing to correlate the timer pop with, but then I saw that we mark these pops as if they were fresh registrations, which isn't ideal but does at least work).

PR #1331, opened 2016-02-23
[Reviewer: Graeme] Pull in memento-as not memento

eleanor-merry:
This PR pulls in memento-as instead of memento, and adds the memento UTs


graemerobertson:
Looks good. You need to update our update-cpp-common script to at least reflect the fact that Sprout no longer has a submodule called Memento, and possibly also to update the new submodule structure (although I think you were leaning towards not doing that).

PR #1342, opened 2016-03-09
[Reviewer: Andrew] Route Tel URIs in the BGCF correctly

### LINKED ISSUE #1316: The BGCF fails to route tel URIs that don't include a routing number.
If the URI classifier determines that the request URI to be forwarded to the BGCF is not a SIP URI and there is no routing number contained within the URI, the BGCF will not attempt to find a route for the number.
This means that if we have a wildcard routing record to catch any tel URIs that need to be routed, for example:
{
  "name" : "Wildcard number matching"
  "number" : "*"
  "route" : ["sip:10.225.84.205:5058"]
}
and the BGCF receives a tel URI with no routing number, the above record will not be matched and no routing target selected. This results in Sprout reporting a '404'- no route to target error.

The code for selecting a BGCF route based off the domain or number contained within the request URI is found in bgcfsproutlet:
https://github.com/Metaswitch/sprout/blob/dev/src/bgcfsproutlet.cpp#L159
This needs to be modified to assign bgcf_routes in cases where we have a tel URI with no routing number.

eleanor-merry:
This PR lets the BGCF route Tel URIs to the default domain. The actual fix is very small, but I pulled out the BGCF tests in scscf_test to a separate UT file, and commonised some of the UT helper functions.

Fixes #1316 (tested through UTs)


src/bgcfsproutlet.cpp
```diff
@@ -163,6 +163,14 @@ void BGCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
...
     bgcf_routes = _bgcf->get_route_from_domain(routing_value, trail());
   }
+  else if ((uri_class == LOCAL_PHONE_NUMBER) ||
+           (uri_class == GLOBAL_PHONE_NUMBER))
+  {
+    // Find the downstream routes based on the domain - this only matches
+    // any wild card routing set up
+    routing_value = "";
+    bgcf_routes = _bgcf->get_route_from_domain(routing_value, trail());
+  }
```
AME2: This looks like the correct logic needed to fix the issue.

src/ut/bgcf_test.cpp
```diff
@@ -0,0 +1,420 @@
...
+TEST_F(BGCFTest, TestSimpleTelURI)
+{
+  add_host_mapping("ut.cw-ngv.com", "10.9.8.7");
+  SCOPED_TRACE("");
+  BGCFMessage msg;
+  msg._toscheme = "tel";
+  msg._to = "16505551234";
+  msg._todomain = "";
+  list<HeaderMatcher> hdrs;
+  hdrs.push_back(HeaderMatcher("Route", ".*10.0.0.2:5060.*"));
```
AME2: Could it be worth a comment here saying the route we use is the same as the one in the Wildcard routing rule in src/ut/test_bgcf_sproutlet_bgcf.json ?

AME2:
All looks good to me, I didn't review the code copied from scscf_test. 
My only suggestion is about adding a comment to make it clearer when we are testing for the routing rules contained in src/ut/test_bgcf_sproutlet_bgcf.json. I can see where we are referencing these rules but not how they are compared against or used.

PR #1344, opened 2016-03-11
[Reviewer: Alex] Replace a lock with a check for PJSIP transport thread

rkd-msw:
This:

* formalises the long-established practice that we can only have one PJSIP transport thread
* adds a function to tell you whether you're on that thread
* changes the SproutletProxy timer functions to check they're on that thread, instead of using a lock

Connection tracking stuff looked more complicated, I'll come and talk to you about that.

### DIFF
--- a/src/basicproxy.cpp
+++ b/src/basicproxy.cpp
@@ -1444,10 +1444,7 @@ void BasicProxy::UASTsx::cancel_trying_timer()
   // We expect to only be called on the PJSIP transport thread, and our data
   // race/locking safety is based on this assumption. Raise an error log if
   // this is not the case.
-  if (!is_pjsip_transport_thread())
-  {
-    TRC_ERROR("Function expected to be called on PJSIP transport thread - has been called on different thread");
-  }
+  CHECK_PJ_TRANSPORT_THREAD();
 
   if (_trying_timer.id == TRYING_TIMER)
   {
@@ -1466,10 +1463,7 @@ void BasicProxy::UASTsx::trying_timer_expired()
   // We expect to only be called on the PJSIP transport thread, and our data
   // race/locking safety is based on this assumption. Raise an error log if
   // this is not the case.
-  if (!is_pjsip_transport_thread())
-  {
-    TRC_ERROR("Function expected to be called on PJSIP transport thread - has been called on different thread");
-  }
+  CHECK_PJ_TRANSPORT_THREAD();
 
   TRC_DEBUG("Trying timer expired for %s, transaction state = %s",
             name(),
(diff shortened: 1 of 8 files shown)


src/basicproxy.cpp
```diff
@@ -1443,7 +1441,13 @@ void BasicProxy::UASTsx::exit_context()
 /// Cancel the trying timer.
 void BasicProxy::UASTsx::cancel_trying_timer()
 {
-  pthread_mutex_lock(&_trying_timer_lock);
+  // We expect to only be called on the PJSIP transport thread, and our data
+  // race/locking safety is based on this assumption. Raise an error log if
+  // this is not the case.
+  if (!is_pjsip_transport_thread())
+  {
+    TRC_ERROR("Function expected to be called on PJSIP transport thread - has been called on different thread");
```
AlexHockey: I thought we were planning to assert if we were on the wrong thread?

AlexHockey: I'm a bit worried that this is an error log as we could end up spamming the log file.

rkd-msw: I polled @steelyvoid , who had the view that anything that might affect service (like asserting) was dangerous.

AlexHockey: Assuming we do tick with a log, I think we should output the current thread ID, and the ID of the transport thread (which might argue for a `CHECK_ON_TRANSPORT_THREAD` macro to reduce code duplication). I also thing the log text should give a stronger hint that anyone seeing this log should raise it as an issue.

rkd-msw: I think that's a sensible risk - this is an error log because we never expect it to get called. I think spamming the log file is a better outcome than having a too-low log level, missing issues here, and risking data corruption.

AlexHockey:
A few questions inline, but I like that you've removed the "break everything" config option.

src/basicproxy.cpp
AlexHockey: I suppose that's OK given that we don't rely on logs for debugging production deployments (so we're not going to swamp useful diags) and we have the log cleanup script to keep disk usage in check.

steelyvoid: Is there any way we could have both behaviours, perhaps controlled by a config or compile flag?  I can see that asserting in FV tests or UTs could be very powerful.

rkd-msw: There is, but I'm reluctant to do so:

* I don't think we'd actually run the same threading model in UT (threaded UTs in general just make things unpredictable), so I don't think assertions there would help
* so we'd need to actually build packages with this compile flag, which runs the risk of us either actually releasing that, or testing with the compile flag then releasing without it and having the compile flag mask some other bug

rkd-msw:
Reassigned to you as you're investigating the issue.

PR #1345, opened 2016-03-11
[Reviewer AMC] Switch ENT logging channel to local7

steelyvoid:
Switch to a less congested logging facility for ENT logs


steelyvoid:
Can you rereview to check that you're happy with the factorisation?  If so, I'll go ahead and make the same changes to the other components.  Note that I have tested that removing the closelogs doesn't have any impact (ENT logging works fine after restarting sprout)

bossmc:
Looks fine.

PR #1349, opened 2016-03-15
[Reviewer: Alex] Replace AvStore with ImpiStore (API changes)

mirw:
Please can you review my interface changes for replacing the AvStore with the ImpiStore?

Obviously, not tested yet (or even code-complete) - please don't merge!

Matt


include/impistore.h
```diff
@@ -0,0 +1,265 @@
...
+    AuthChallenge(const Type _type, const std::string& _nonce, uint64_t _expires) :
+      type(_type),
+      nonce(_nonce),
+      nonce_count(1),
+      expires(_expires),
+      branch_id(),
+      _cas(0) {};
+
+    /// Destructor must be virtual as we're going to extend this class.
+    virtual ~Av() {};
```
AlexHockey: Should be `~AuthChallenge`

include/impistore.h
```diff
@@ -0,0 +1,265 @@
...
+
+    /// Type of the AV
+    enum Type type;
+
+    /// Nonce used for this challenge.
+    std::string nonce;
+
+    /// Minimum nonce count we will accept - any nonce count lower than this
+    /// might be a reply and must be rejected.
```
AlexHockey: reply -> replay

include/impistore.h
```diff
@@ -0,0 +1,265 @@
...
+      nonce(_nonce),
+      realm(_realm),
+      qop(_qop),
+      ha1(_ha1),
+      expires(_expires) {};
+
+    /// Destructor.
+    virtual ~DigestAuthChallenge() {};
+
+    /// Digest realm (public)
```
AlexHockey: What do public and private mean here? At first I thought you meant the C++ sense - now I think you mean "whether the element should be sent to the UE" but I'm still not 100% sure.

include/impistore.h
```diff
@@ -0,0 +1,265 @@
...
+    /// Constructor.
+    /// @param _nonce        Nonce used for this challenge.
+    /// @param _challenge    AKA challenge.
+    /// @param _response     AKA response.
+    /// @param _expires      Absolute expiry time in seconds since the epoch.
+    AKAAuthChallenge(const std::string& _nonce,
+                     const std::string& _challenge,
+                     const std::string& _response,
+                     uint64_t _expires) :
+      type(AuthChallenge::Type::AKA),
```
AlexHockey: You will need to call the `AuthChallenge` constructor here rather than initializing its member variables in the subclass.

include/impistore.h
mirw: Yes, I'd meant whether it was public or private knowledge from a security perspective.  I'm happy just to remove these comments.

include/impistore.h
mirw: Thanks!  I'd spotted and fixed this just after I'd send it to you for review, but I saw you were already reviewing and didn't want to change things under your feet.

include/impistore.h
```diff
@@ -0,0 +1,265 @@
...
+      challenge(_challenge),
+      response(_response),
+      expires(_expires) {};
+
+    /// Destructor.
+    virtual ~AKAAuthChallenge() {};
+
+    /// AKA challenge (public)
+    std::string challenge;
```
AlexHockey: I thought this was the same as the nonce for AKA. Have I missed something?

include/impistore.h
```diff
@@ -0,0 +1,265 @@
...
+    /// List of authentication challenges that can be used with this IMPI
+    std::vector<AuthChallenge> auth_challenges;
+
+  private:
+    /// Memcached CAS value.
+    uint64_t _cas;
+
+    // The IMPI store is a friend so it can read our CAS value.
+    friend class ImpiStore;
+  };
```
AlexHockey: I think it would be useful to have a `get_challenge(nonce)` method on this class. Thoughts?

include/impistore.h
```diff
@@ -0,0 +1,265 @@
...
+    /// Minimum nonce count we will accept - any nonce count lower than this
+    /// might be a reply and must be rejected.
+    uint32_t nonce_count;
+
+    /// Expiry time - absolute in seconds since the Epoch
+    uint64_t expires;
+
+    /// Branch ID of request on which challenge was issued, or empty if not
+    /// applicable.
+    std::string branch_id;
```
AlexHockey: I fell like the store shouldn't know about things like the SIP transport and SAS correlation. Should we just call this `correlator`?

include/impistore.h
```diff
@@ -0,0 +1,265 @@
...
+  Store* _data_store;
+
+  /// The mode to use when accessing the data store.
+  Mode _mode;
+};
+
+// Utility function - retrieves the "branch" field from the give challenge
+// and raises a correlating transaction marker in the given trail.
+void correlate_branch_from_challenge(ImpiStore::AuthChallenge* auth_challenge,
+                                     SAS::TrailId trail);
```
AlexHockey: As above, I'm not sure if this should be part of the store or not.

AlexHockey:
Looks good. :smile: 

The old store had a tombstone flag - I'm fairly sure that I can just use the nonce count to have the same effect but I'd like to quickly run that past you and check you agree. 

Am I right in thinking that when we drop support for upgrading to the data format, we will bin the `_mode` and `get_impi_with_nonce` methods?

include/impistore.h
mirw: Good spot.  Removed.

include/impistore.h
mirw: Yes, I wondered about this.  Added.

include/impistore.h
mirw: Renamed.

include/impistore.h
mirw: I've reworded to be less "branch"-oriented, but left in - it feels like a useful utility to have, and I can't see anywhere better to put it.

mirw:
Thanks for your comments.

I don't think we need a per-challenge tombstone flag, although I've added a (private) list of nonces to the `Impi` class, so that we can detect if (and which) challenges have been deleted.

When written to the store, both the IMPI and the AVs need a tombstone flag (as before, for the AVs), but they need not appear on the API.

Is that OK?

mirw:
...and yes, we'll remove (or cut back) the `mode` options and the `get_impi_with_nonce` methods once we no longer need to support upgrade.

mirw:
@AlexHockey, FYI, I'm not expecting you to review the further code I've just pushed to this branch, but please grab me to sync up before you start development here.

Issue #1347, opened 2016-03-14
Sprout should have tests that ACRs are send at the right point


AlexHockey:
At present sprout has tests for the `ACR` class that check the JSON produced by the class when fed various message flows. However there are no tests that:

* ACRs are generated at the correct points in time (e.g. that the registrar sends an EVENT ACR when it responds to a REGISTER). 
* ACR flows are always sensible (e.g. sprout always sends an END if it sends a START). 

This issue tracks adding this framework and writing a tactical set of tests.

PR #1350, opened 2016-03-15
[Reviewer: Andy] Send ACRs when the registrar rejects a REGISTER. 

AlexHockey:
Andy, please can you review this PR to make sure the registrar sends ACRs even when it rejects REGISTERs. This contains a big refactor to move some enums into the `ACR` class to avoid them polluting the namespace (I started to hit this because I included `acr.h` in `pjutils.h` and found some places where some identifiers defined in `acr.h` conflicted with identifiers in cpp files). 

I haven't unit tested this due to https://github.com/Metaswitch/sprout/issues/1347. I have checked the existing UTs work.

This is surprisingly difficult to live test as its difficult to get a REGISTER to fail in the registrar without also having it fail earlier on. In the end I run the following testcase:

* Set up deployment to not use an I-CSCF. 
* Send REGISTER, receive 401. 
* Kill homestead. 
* Send REGISTER, receive rejection from the registrar (jn this case a 403). 

This gave me two ACRs (one for the REGISTER/401 from the authentication module, the other for the REGISTER/403 from the registrar). Here's the second ACR:

```javascript
{"peers":{"ccf":["cdf.ajh1.cw-ngv.com"]},"event":{"Accounting-Record-Type":1,"User-Name":"3010000839@ajh1.cw-ngv.com","Event-Timestamp":1458043857,"Service-Information":{"Subscription-Id":[{"Subscription-Id-Type":2,"Subscription-Id-Data":"sip:3010000839@ajh1.cw-ngv.com"}],"IMS-Information":{"Event-Type":{"SIP-Method":"REGISTER","Expires":3600},"Role-Of-Node":0,"Node-Functionality":0,"User-Session-Id":"3fa769708ec3f9907f5283930cbe05fa","Called-Party-Address":"sip:3010000839@ajh1.cw-ngv.com","Time-Stamps":{"SIP-Request-Timestamp":1458043857,"SIP-Request-Timestamp-Fraction":817,"SIP-Response-Timestamp":1458043857,"SIP-Response-Timestamp-Fraction":820},"IMS-Charging-Identifier":"3fa769708ec3f9907f5283930cbe05fa","Cause-Code":403,"From-Address":"<sip:3010000839@ajh1.cw-ngv.com>;tag=7c8297cbe2219c2f1675b1394f887ac6","IMS-Visited-Network-Identifier":"ajh1.cw-ngv.com","Route-Header-Received":"<sip:sprout.ajh1.cw-ngv.com:5054;transport=TCP;lr;orig>","Instance-Id":"<urn:uuid:30d86d6f-d52c-59d2-b678-1182f3c65fbb>"}}}}
```

This ACR looks sensible to me - note in particular that the cause code is correctly reported as 403. 

I haven't been able to test all the other cases where we reject a REGISTER, but hopefully this testcase is convincing enough that the approach I've taken works.

### DIFF
--- a/include/acr.h
+++ b/include/acr.h
@@ -54,12 +54,6 @@ extern "C" {
 #include "ralf_processor.h"
 #include "servercaps.h"
 
-typedef enum { SCSCF=0, PCSCF=1, ICSCF=2, BGCF=5, AS=6, IBCF=7 } Node;
-
-typedef enum { CALLED_PARTY=0, CALLING_PARTY=1 } Initiator;
-
-typedef enum { NODE_ROLE_ORIGINATING=0, NODE_ROLE_TERMINATING=1 } NodeRole;
-
 /// Class tracking state required for Rf ACR messages.  An instance of this
 /// class is created for each SIP transaction that requires accounting, and
 /// the class is passed messages and other data during processing of the
@@ -90,6 +84,10 @@ typedef enum { NODE_ROLE_ORIGINATING=0, NODE_ROLE_TERMINATING=1 } NodeRole;
 class ACR
 {
 public:
+  typedef enum { SCSCF=0, PCSCF=1, ICSCF=2, BGCF=5, AS=6, IBCF=7 } Node;
+  typedef enum { CALLED_PARTY=0, CALLING_PARTY=1 } Initiator;
+  typedef enum { NODE_ROLE_ORIGINATING=0, NODE_ROLE_TERMINATING=1 } NodeRole;
+
   /// Unspecified timestamp value.
   static const pj_time_val unspec;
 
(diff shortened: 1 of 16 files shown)


bossmc:
Refactor looks good, if tedious.  The fixes look fine, but I'm worried that there's nothing stopping us introducing the same issue in the future if we add more exit points from the REGISTER handling function... is there something we can do here (Mike considered using the ACR as a guard but then got scared of handling failures in destructors)?

AlexHockey:
We discussed. Adding guards doesn't actually gain too much - even if you can guarantee to send an ACR in the right places, you can't easily guarantee it is correctly populated.

PR #1353, opened 2016-03-16
Make auth challenges be a vector of pointers to avoid object slicing

AlexHockey:


mirw:
Does the rest of sprout/UTs still build with this change?  (I don't feel too strongly about this, but wanted to check.)

AlexHockey:
It does now.

PR #1356, opened 2016-03-18
[Reviewer: Matt] Code and UTs for non-store nonce-count support.

AlexHockey:
Hi Matt. My nonce count code is ready to be reviewed and merged into the main sto1119 branch. The current status is:

* I have written all the code and UTs I think I need to. 
* UTs are all passing
* Coverage in the authentication module and handlers 100%
* Valgrind is reporting some leaked authentication challenges but I think this is just because the version of the `ImpiStore` in my version of the code is incomplete. 

Thanks, Alex.

### DIFF
--- a/include/authentication.h
+++ b/include/authentication.h
@@ -56,6 +56,8 @@ enum struct NonRegisterAuthentication
   IF_PROXY_AUTHORIZATION_PRESENT
 };
 
+typedef int(*binding_expiry_func)(pjsip_contact_hdr* contact, pjsip_expires_hdr* expires);
+
 pj_status_t init_authentication(const std::string& realm_name,
                                 ImpiStore* impi_store,
                                 HSSConnection* hss_connection,
@@ -64,7 +66,8 @@ pj_status_t init_authentication(const std::string& realm_name,
                                 NonRegisterAuthentication non_register_auth_mode_param,
                                 AnalyticsLogger* analytics_logger,
                                 SNMP::AuthenticationStatsTables* auth_stats_tables,
-                                bool nonce_count_supported_arg);
+                                bool nonce_count_supported_arg,
+                                binding_expiry_func expiry_func);
 
 void destroy_authentication();
 
--- a/include/authentication.h
+++ b/include/authentication.h
@@ -56,7 +56,8 @@ enum struct NonRegisterAuthentication
   IF_PROXY_AUTHORIZATION_PRESENT
 };
 
-typedef int(*binding_expiry_func)(pjsip_contact_hdr* contact, pjsip_expires_hdr* expires);
+typedef int(*get_expiry_for_binding_fn)(pjsip_contact_hdr* contact,
+                                        pjsip_expires_hdr* expires);
 
 pj_status_t init_authentication(const std::string& realm_name,
                                 ImpiStore* impi_store,
@@ -67,7 +68,7 @@ pj_status_t init_authentication(const std::string& realm_name,
                                 AnalyticsLogger* analytics_logger,
                                 SNMP::AuthenticationStatsTables* auth_stats_tables,
                                 bool nonce_count_supported_arg,
-                                binding_expiry_func expiry_func);
+                                get_expiry_for_binding_fn get_expiry_for_binding_arg);
 
 void destroy_authentication();
 
--- a/include/authentication.h
+++ b/include/authentication.h
@@ -63,7 +63,8 @@ pj_status_t init_authentication(const std::string& realm_name,
                                 ACRFactory* rfacr_factory,
                                 NonRegisterAuthentication non_register_auth_mode_param,
                                 AnalyticsLogger* analytics_logger,
-                                SNMP::AuthenticationStatsTables* auth_stats_tables);
+                                SNMP::AuthenticationStatsTables* auth_stats_tables,
+                                bool nonce_count_supported_arg);
 
 void destroy_authentication();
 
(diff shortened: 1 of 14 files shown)


include/authentication.h
```diff
@@ -63,7 +66,9 @@ pj_status_t init_authentication(const std::string& realm_name,
                                 ACRFactory* rfacr_factory,
                                 NonRegisterAuthentication non_register_auth_mode_param,
                                 AnalyticsLogger* analytics_logger,
-                                SNMP::AuthenticationStatsTables* auth_stats_tables);
+                                SNMP::AuthenticationStatsTables* auth_stats_tables,
+                                bool nonce_count_supported_arg,
+                                get_expiry_for_binding_fn get_expiry_for_binding_arg);
```
mirw: Why do we need to make this function configurable?  Why can't we just make the authentication module hard-coded to call into a utility function?

src/authentication.cpp
```diff
@@ -149,6 +157,31 @@ static pjsip_digest_credential* get_credentials(const pjsip_rx_data* rdata)
   return credentials;
 }
 
+/// Given a request that has passed authentication, calculate the time at which
+/// to expire the challenge.
+///
+/// @param rdata - The request in question.
+///
+/// @return The time (in seconds) to store the challenge for.
```
mirw: Worth clarifying that this gives an absolute (not relative) time?

src/authentication.cpp
```diff
@@ -756,38 +786,115 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
...
+        if (nonce_count_supported && is_register)
+        {
+          new_expiry = calculate_challenge_expiration_time(rdata);
+        }
+
+        Store::Status store_status;
+        do
+        {
+          // Work out the next nonce count and expiry to use (in case another
+          // UE authenticated using this challenge at the same time. We don't
```
mirw: Missing close parenthesis.

src/authentication.cpp
```diff
@@ -756,38 +786,115 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
...
+          // currently police against this, but we don't want the expiration or
+          // nonce counts to travel backwards.
```
mirw: Is it OK that we don't police this?  Should we CAS the new AV and only accept the request if it succeeds?  Maybe we should just track this as a future?  It looks like we didn't police this previously either.

src/handlers.cpp
```diff
@@ -456,11 +460,12 @@ HTTPCode DeregistrationTask::handle_request()
...
-                                                          it->first,
-                                                          it->second,
-                                                          aor_pair,
-                                                          NULL);
+          deregister_bindings(_cfg->_remote_sdm,
+                              it->first,
+                              it->second,
+                              aor_pair,
+                              NULL,
+                              NULL);
```
mirw: Is it OK to ignore the impis_to_delete from this deregister_bindings call?

src/main.cpp
```diff
@@ -220,6 +221,7 @@ const static struct pj_getopt_option long_opt[] =
   { "pidfile",                      required_argument, 0, OPT_PIDFILE},
   { "plugin-option",                required_argument, 0, 'N'},
   { "impi-store-mode",              required_argument, 0, OPT_IMPI_STORE_MODE},
+  { "impi-store-mode",              no_argument,       0, OPT_NONCE_COUNT_SUPPORTED},
```
mirw: Looks like a typo in the option name.

mirw:
Looks good.  A few comments inline.

sprout-base.root/etc/init.d/sprout
```diff
@@ -283,6 +283,8 @@ get_daemon_args()
         [ "$stateless_proxies" = "" ]             || DAEMON_ARGS="$DAEMON_ARGS --stateless-proxies=$stateless_proxies"
         [ "$ralf_threads" = "" ]                  || DAEMON_ARGS="$DAEMON_ARGS --ralf-threads=$ralf_threads"
         [ "$non_register_authentication" = "" ]   || DAEMON_ARGS="$DAEMON_ARGS --non-register-authentication=$non_register_authentication"
+        [ "$impi_store_mode" = "" ]               || DAEMON_ARGS="$DAEMON_ARGS --impi-store-mode=$impi_store_mode"
+        [ "$nonce_count_supported" = "Y" ]        && DAEMON_ARGS="$DAEMON_ARGS --nonce-count-supported"
```
mirw: Not sure it matters here, but generally better to write `[ "$nonce_count_supported" != "Y" ] || ...` as this means the expression is always true.  Otherwise, if `set -e` is set, the whole script would fail if `nonce_count_supported` was not `Y`.

include/authentication.h
AlexHockey: I wanted to decouple the authentication and registrar modules (for unit test in particular).

sprout-base.root/etc/init.d/sprout
AlexHockey: Thanks for the tip. I will make that change.

src/authentication.cpp
AlexHockey: Yes (I changed the implementation and forgot to update the comment).

src/handlers.cpp
AlexHockey: My thinking when I wrote this was challenges are not GR so we should only pay attention to the local store. But given that IMPUs and IMPIs can be stored on different replicas you could lose the IMPU in the local site and still have some IMPIs. So it would probably be safer to add the IMPIs learned from the remote site.

src/main.cpp
AlexHockey: I'm not going to change this here as you fixed it in a separate PR.

PR #1358, opened 2016-03-19
[Reviewer: Rob] AS alarms

AlexHockey:
HI Rob, please could you review this change to add AS alarms to sprout? A couple of things to note:

* I tweaked all the S-CSCF UTs to check that AS communication tracker was being invoked correctly. I did this as the tests involve a lot of interesting AS topologies that I wanted to test.  I also decided to add a lot of missing 100 trying flows (from the AS -> Sprout) so that the majority of testcases were treated as AS successes (as that is less confusing).
* I am hoping to add the feature that support requested to log the error reason, but I'll do that as a separate PR. 

Tested in UT so far. I'll test live when you have reviewed the code (or at least gotten far enough through to say "this doesn't need a fundamental re-write").

### DIFF
--- a/include/as_communication_tracker.h
+++ b/include/as_communication_tracker.h
@@ -57,8 +57,8 @@ public:
   ///
   /// The object takes ownership of the alarm and the logs passed to it.
   AsCommunicationTracker(Alarm* alarm,
-                         PDLog1<const char*>* as_failed_log,
-                         PDLog1<const char*>* as_ok_log);
+                         const PDLog1<const char*>* as_failed_log,
+                         const PDLog1<const char*>* as_ok_log);
 
   /// Destructor.
   virtual ~AsCommunicationTracker();
@@ -96,8 +96,8 @@ private:
   // and when they are considered to be OK.
   //
   // Each log takes the URI of the AS as a parameter.
-  PDLog1<const char*>* _as_failed_log;
-  PDLog1<const char*>* _as_ok_log;
+  const PDLog1<const char*>* _as_failed_log;
+  const PDLog1<const char*>* _as_ok_log;
 
   /// Check if any application servers are healthy. If so, log them and
   /// consider clearing the alarm.
--- a/include/as_communication_tracker.h
+++ b/include/as_communication_tracker.h
@@ -1,0 +1,111 @@
+/**
+ * @file as_communication_tracker.h
+ *
+ * Project Clearwater - IMS in the Cloud
+ * Copyright (C) 2016  Metaswitch Networks Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version, along with the "Special Exception" for use of
+ * the program along with SSL, set forth below. This program is distributed
+ * in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details. You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ *
+ * The author can be reached by email at clearwater@metaswitch.com or by
+ * post at Metaswitch Networks Ltd, 100 Church St, Enfield EN2 6BQ, UK
+ *
+ * Special Exception
+ * Metaswitch Networks Ltd  grants you permission to copy, modify,
+ * propagate, and distribute a work formed by combining OpenSSL with The
+ * Software, or a work derivative of such a combination, even if such
+ * copying, modification, propagation, or distribution would otherwise
+ * violate the terms of the GPL. You must comply with the GPL in all
+ * respects for all of the code used other than OpenSSL.
+ * "OpenSSL" means OpenSSL toolkit software distributed by the OpenSSL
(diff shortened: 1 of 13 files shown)


AlexHockey:
A couple more things:

* The design is slightly different from the one you reviewed, to accommodate the request to have separate logs for each AS. 
* I was worried enough about this that I wanted to test the logs were generated correctly, so I did end up using a mock log (see https://github.com/Metaswitch/cpp-common/pull/481)

include/as_communication_tracker.h
```diff
@@ -0,0 +1,111 @@
...
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+
+#include <map>
+#include <string>
+
+#include "pdlog.h"
+#include "alarm.h"
+
+#ifndef AS_COMMUNICATION_TRACKER_H__
```
rkd-msw: I know I'm fighting a losing battle, but [double underscores are reserved to the compiler](http://stackoverflow.com/a/228797).

src/as_communication_tracker.cpp
```diff
@@ -0,0 +1,168 @@
...
+    as_iter->second++;
+  }
+  else
+  {
+    // This is the first time we've spotted that the AS has failed, so log this
+    // fact.
+    TRC_DEBUG("First failure for this AS - generate log");
+    _as_failed_log->log(as_uri.c_str());
+
+    _as_failures.emplace(as_uri, 1);
```
rkd-msw: Does `_as_failures[as_uri] = 1;` not work?

src/scscfsproutlet.cpp
```diff
@@ -568,35 +594,61 @@ void SCSCFSproutletTsx::on_rx_response(pjsip_msg* rsp, int fork_id)
       // The AS chain isn't complete, so the response must be from an
       // application server.  Check to see if we need to trigger default
       // handling.
-      if ((!_cancelled) &&
+      if ((!_seen_response) &&
```
rkd-msw: Is it right that a 100 Trying then a 503 Service Unavailable counts as a success for that AS? Maybe the best choice of a bad lot, but worth a comment.

src/ut/as_communication_tracker_test.cpp
```diff
@@ -0,0 +1,267 @@
...
+    delete _mock_error_log;
+    delete _mock_ok_log;
+  }
+
+  const std::string AS1 = "as1";
+  const std::string AS2 = "as2";
+};
+
+// Advance time to the next AsCommunicationTracker checking period. This is
+// slightly encapsulation breaking (sincec it means the UT knows how long this
```
rkd-msw: sincec

src/as_communication_tracker.cpp
```diff
@@ -0,0 +1,168 @@
...
+          }
+          else
+          {
+            curr_as->second = 0;
+          }
+
+          curr_as = next_as;
+        }
+
+        if (_as_failures.empty())
```
rkd-msw: I think our alarm reliability approach requires that this should clear the alarm when all app servers are healthy, even if it wasn't previously raised. I think just moving this out of the `if (!_as_failures.empty())` block would do this.

src/ut/as_communication_tracker_test.cpp
```diff
@@ -0,0 +1,267 @@
...
+// As such each on_success call checks the AS health in
+// the previous period. So the first call checks the period in which the AS
+// was failing (so does not treat the AS as being healthy) wheras only on the
+// second call do we see the AS being healthy.
+//
+// This looks worse in UT than in real life!
+//
+
+// Test that is an AS is healthy, nothing happens.
+TEST_F(AsCommunicationTrackerTest, SuccessIsIdempotent)
```
rkd-msw: Good test - wouldn't have spotted the issue at https://github.com/Metaswitch/sprout/pull/1358/files#r56808017 without seeing this.

src/ut/mock_as_communication_tracker.h
```diff
@@ -0,0 +1,53 @@
+/**
+ * @file mock_as_communication_tracker.h
+ *
+ * project clearwater - ims in the cloud
+ * copyright (c) 2013  metaswitch networks ltd
+ *
+ * this program is free software: you can redistribute it and/or modify it
```
rkd-msw: Classic @AlexHockey copyright statement here.

rkd-msw:
Looks good - relatively few comments for a 1,000 line PR.

src/as_communication_tracker.cpp
AlexHockey: Yes it will. I made some fixes in this case and there was a reason for using emplace originally but it doesn't apply any more.

src/as_communication_tracker.cpp
AlexHockey: As discussed, I'm going to remove the `if (!_as_failures.empty())` test entirely as `Alarm` is idempotent.

src/scscfsproutlet.cpp
AlexHockey: Yes that is correct according to 24.229 section 5.4.3.2/5.4.3.3, default handling is applied if the AS 

> receives a 408 (Request Timeout) response or a 5xx response from the AS without previously receiving a 1xx response to the original SIP request, and without previously receiving a SIP request from the AS that contained the same original dialog identifier as the original request;

I will add a comment.

AlexHockey: Ah, except that 1xx doesn't include 100 does it.

AlexHockey:
Thanks Rob. I lost a few updates from my Vim snippets recently which explains the double-underscore and the lowercase copyright statement.

PR #1361, opened 2016-03-21
[Reviewer: Alex] Nonce-count support changes

mirw:
Alex,

Please can you review my nonce-count support changes?  Note that this is not against dev, but against a specific branch I've created for review (so that it excludes the nonce-count changes you made yourself).

This has all been live-tested as discussed, and UT coverage of the new code is 100%.

Matt


mirw:
One concern I had: is it safe to delete IMPIs without CAS?  I'm not sure (in fact, I think it probably isn't!), but would like to discuss with you.

include/impistore.h
```diff
@@ -0,0 +1,393 @@
...
+    };
+
+    /// Initial nonce count, when creating new challenges.
+    static const int INITIAL_NONCE_COUNT = 1;
+
+    /// Default expiry time, if no value can be read from the store.  This
+    /// should always be long enough for the UE to respond to the
+    /// authentication challenge, and means that on authentication timeout our
+    /// 30-second Chronos timer should pop before it expires.
+    static const int DEFAULT_EXPIRES = 40;
```
AlexHockey: I think this is another field that is only needed on upgrade (since when using the IMPI-only mode all challenges should have an expires specified by the user of the store). If so, is it worth adding a comment saying this?

include/impistore.h
```diff
@@ -0,0 +1,393 @@
...
+    /// Write to JSON writer (IMPI format).
+    virtual void write_json(rapidjson::Writer<rapidjson::StringBuffer>* writer);
+
+    /// Deserialization from JSON (IMPI format).
+    static ImpiStore::AuthChallenge* from_json(rapidjson::Value* json);
+
+    /// Serialization to JSON (AV format).
+    std::string to_json_av();
+
+    /// Write to JSON writer (IMPI format).
```
AlexHockey: Av format, not IMP?I format?

include/impistore.h
```diff
@@ -0,0 +1,393 @@
...
+    virtual void write_json_av(rapidjson::Writer<rapidjson::StringBuffer>* writer);
+
+    /// Deserialization from JSON (AV format).
+    static ImpiStore::AuthChallenge* from_json_av(const std::string& nonce, const std::string& json);
+
+    /// Deserialization from JSON (AV format).
+    static ImpiStore::AuthChallenge* from_json_av(const std::string& nonce, rapidjson::Value* json);
+
+    /// Memcached CAS value.  Only used for Mode::READ_AV_IMPU_WRITE_AV_IMPI.
```
AlexHockey: IMPI not IMPU

include/impistore.h
```diff
@@ -0,0 +1,393 @@
...
+
+    /// Helper - get authentication challenge for a given nonce.
+    /// @returns the authentication challenge, or NULL if not found
+    /// @param nonce         The nonce to look up.
+    ImpiStore::AuthChallenge* get_auth_challenge(const std::string& nonce);
+
+    /// Private ID
+    std::string impi;
+
+    /// List of authentication challenges that can be used with this IMPI
```
AlexHockey: Might be worth a comment about ownership of the challenges.

include/impistore.h
```diff
@@ -0,0 +1,393 @@
...
+  };
+
+  /// Constructor.
+  /// @param data_store    A pointer to the underlying data store.
+  /// @param mode          The mode to use when accessing the data store.
+  ImpiStore(Store* data_store, Mode mode);
+
+  /// Destructor.
+  virtual ~ImpiStore();
+
```
AlexHockey: Again, some comments about ownership would be useful.

src/authentication.cpp
```diff
@@ -89,6 +89,9 @@ pjsip_module mod_authentication =
 // Configuring PJSIP with a realm of "*" means that all realms are considered.
 const pj_str_t WILDCARD_REALM = pj_str("*");
 
+// Initial expiry time (in seconds) for authentication challenges.
+const uint32_t AUTH_CHALLENGE_INIT_EXPIRES = 30;
```
AlexHockey: Shouldn't this be the same as the `DEFAULT_EXPIRY` defined in the store?

src/authentication.cpp
```diff
@@ -504,6 +547,7 @@ void create_challenge(pjsip_digest_credential* credentials,
       chronos->send_post(timer_id, 30, "/authentication-timeout", chronos_body, get_trail(rdata));
```
AlexHockey: Update this to use the `AUTH_CHALLENEG_INIT_EXPIRES` constant?

src/authentication.cpp
```diff
@@ -483,19 +523,22 @@ void create_challenge(pjsip_digest_credential* credentials,
...
-    branch_value.SetString(branch.c_str(), (*av).GetAllocator());
-    (*av).AddMember("branch", branch_value, (*av).GetAllocator());
-
-    // Write the authentication vector (as a JSON string) into the AV store.
-    TRC_DEBUG("Write AV to store");
-    uint64_t cas = 0;
-    Store::Status status = av_store->set_av(impi, nonce, av, cas, get_trail(rdata));
+    // Write the new authentication challenge to the IMPI store
+    TRC_DEBUG("Write authentication challenge to IMPI store");
+    ImpiStore::Impi* impi_obj = impi_store->get_impi(impi, get_trail(rdata));
```
AlexHockey: Should we have a CAS loop here in case two challenges get issued at almost the same time?

src/main.cpp
```diff
@@ -1075,6 +1082,27 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
...
+        TRC_INFO("IMPI store mode set to: av-impi");
+      }
+      else if (stricmp(pj_optarg, "impi") == 0)
+      {
+        options->impi_store_mode = ImpiStore::Mode::READ_IMPI_WRITE_IMPI;
+        TRC_INFO("IMPI store mode set to: impi");
+      }
+      else
+      {
+        TRC_ERROR("Unknown IMPI store mode: %s", pj_optarg);
```
AlexHockey: I wondered if we should set the mode to av-impi if the option cannot be parsed on the basis that the operator is probably trying to set it to av-impi but mis-typed, and not migrating data to the new format would be safer in this case. Thoughts?

src/impistore.cpp
```diff
@@ -0,0 +1,923 @@
...
+  ImpiStore::AuthChallenge* auth_challenge = NULL;
+  if (json->IsObject())
+  {
+    // First, identify what type of AuthChallenge this is, and call through to
+    // that class to deserialize.
+    std::string type = "";
+    JSON_SAFE_GET_STRING_MEMBER(*json, JSON_TYPE, type);
+    if (type == JSON_TYPE_DIGEST)
+    {
+      auth_challenge = ImpiStore::DigestAuthChallenge::from_json(json);
```
AlexHockey: It's worth a comment somewhere that these objects are deserialized bottom-to-top rather than top-to-bottom (especially as the latter is how C++ classes are initialized).

src/impistore.cpp
```diff
@@ -0,0 +1,923 @@
...
+      JSON_SAFE_GET_STRING_MEMBER(*json, JSON_CORRELATOR, auth_challenge->correlator);
+      if (auth_challenge->nonce_count == 0)
+      {
+        // We should always have a nonce_count, but to ease version
+        // forward-compatibility, default it if not found.
+        TRC_WARNING("No \"%s\" field in JSON authentication challenge - defaulting to %u",
+                    JSON_NONCE_COUNT, INITIAL_NONCE_COUNT);
+        auth_challenge->nonce_count = INITIAL_NONCE_COUNT;
+      }
+      if (auth_challenge->expires == 0)
```
AlexHockey: Any chance of having a blank line before this one? I saw this as an else branch to the preceeding if because there was white space separating them.

src/impistore.cpp
```diff
@@ -0,0 +1,923 @@
...
+                JSON_HA1);
+    delete auth_challenge; auth_challenge = NULL;
+  }
+  return auth_challenge;
+}
+
+void ImpiStore::DigestAuthChallenge::write_json_av(rapidjson::Writer<rapidjson::StringBuffer>* writer)
+{
+  // Write all the DigestAuthChallenge fields to JSON, in AV format.  We call
+  // into the superclass to write base AuthChallenges fields.
```
AlexHockey: We don't actually *call* into the superclass.

AlexHockey: D'oh, yes we do. Ignore me.

src/impistore.cpp
```diff
@@ -0,0 +1,923 @@
...
+
+int ImpiStore::Impi::get_expires()
+{
+  // Spin through the AuthChallenges, finding the latest expires time.
+  int expires = 0;
+  for (std::vector<ImpiStore::AuthChallenge*>::iterator it = auth_challenges.begin();
+       it != auth_challenges.end();
+       it++)
+  {
+    expires = ((*it)->expires > expires) ? (*it)->expires : expires;
```
AlexHockey: `std::max` would be more expressive.

src/impistore.cpp
```diff
@@ -0,0 +1,923 @@
...
+          SAS::Event event(trail, SASEvent::IMPISTORE_AV_SET_FAILURE, 0);
+          event.add_var_param(impi->impi);
+          event.add_var_param(nonce.c_str());
+          event.add_static_param(local_status);
+          SAS::report_event(event);
+          // Update status, but only if it's not already DATA_CONTENTION - that's
+          // the most significant status.
+          if (status != Store::Status::DATA_CONTENTION)
+          {
+            status = local_status;
```
AlexHockey: Is it right that OK can overwrite error?

src/ut/impistore_test.cpp
```diff
@@ -0,0 +1,770 @@
...
+  };
+};
+
+/// Constant strings.
+const std::string IMPI = "private@example.com";
+const std::string NONCE1 = "nonce1";
+const std::string NONCE2 = "nonce2";
+
+/// Example IMPI, with a single digest authentication challenge.
+ImpiStore::Impi* example_impi1()
```
AlexHockey: Could you make these names a bit more descriptive?

src/ut/impistore_test.cpp
```diff
@@ -0,0 +1,770 @@
...
+}
+
+TYPED_TEST(ImpiTwoStoreTest, Set1DeleteAC1Get2)
+{
+  ImpiStore::Impi* impi1 = example_impi3();
+  Store::Status status = this->scenario->store1->set_impi(impi1);
+  ASSERT_EQ(Store::Status::OK, status);
+  ImpiStore::Impi* impi2 = this->scenario->store1->get_impi(IMPI);
+  ASSERT_TRUE(impi2 != NULL);
+  ASSERT_EQ(2, impi2->auth_challenges.size());
```
AlexHockey: Why can't you `expect_impis`equal` here?

AlexHockey:
Phew, that was a long PR! I thought it looked good, most of my feedback is about adding a bit more commenting in places. 

I commented on whether the store should return OK if some writes succeed and others fail with an ERROR. I'm not sure what to do, but I'm happy to discuss. 

On unit testing:

* My eyes started to glaze over a bit, which I think is because each test is a solid block of code with no blank lines to separate different logical actions. 
* Should we have some tests for data contention (I believe the `LocalStore` can simulate it)?

include/impistore.h
mirw: Ownership of what?  Sorry, it wasn't clear which code you were referring to.

src/authentication.cpp
mirw: No, the AuthChallenges expires needs to be longer so that it's still around when this timer pops.

include/impistore.h
AlexHockey: I was referring to the impis passed into and returned from these functions.

src/authentication.cpp
AlexHockey: OK. `AUTH_CHALLENGE_INIT_EXPIRES` is currently also set to 30.

src/impistore.cpp
mirw: No, it can't - local_status must not be OK, otherwise we wouldn't be in this branch.

src/authentication.cpp
mirw: Thanks!  As discussed, @AlexHockey, you're going to pick up this mark-up.  (It was highlighted as part of my refactoring, where I directly translated the old calls into new ones, but I think the nature of the store has changed here,)

src/impistore.cpp
AlexHockey: ah, of course. Good good.

src/authentication.cpp
mirw: Yes, I've fixed that to be 40 (as the AvStore was).

include/impistore.h
mirw: No, I'd like to keep this even after we've upgraded.  From experience, we've found that being able to default non-critical fields makes future upgrade easier.

src/main.cpp
mirw: I'd prefer not to try second-guessing here - either ignore the parameter or fail to parse and abort.

src/ut/impistore_test.cpp
mirw: Added.  I've left in the ASSERT_EQ, though, because the next line is to delete 1 of them.

PR #1365, opened 2016-03-22
[Reviewer: Matt] Cope with data contention when creating a new challenge

AlexHockey:
Matt, please can you review my change to cope with data contention when creating auth challenges. Tested in UT.

### DIFF
--- a/src/authentication.cpp
+++ b/src/authentication.cpp
@@ -555,9 +555,14 @@ void create_challenge(pjsip_digest_credential* credentials,
 
       if (challenge != NULL)
       {
-        // The IMPI already has a challenge. This shouldn't happen, but if it
-        // does be defensive and update the existing one (making sure the nonce
-        // count and expiry don't move backwards).
+        // The IMPI already has a challenge. This shouldn't happen in mainline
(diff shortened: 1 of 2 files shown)


src/authentication.cpp
```diff
@@ -531,14 +531,65 @@ void create_challenge(pjsip_digest_credential* credentials,
...
+
+      // Check whether the IMPI has an existing auth challenge.
+      ImpiStore::AuthChallenge* challenge =
+        impi_obj->get_auth_challenge(auth_challenge->nonce);
+
+      if (challenge != NULL)
+      {
+        // The IMPI already has a challenge. This shouldn't happen, but if it
+        // does be defensive and update the existing one (making sure the nonce
+        // count and expiry don't move backwards).
```
mirw: Should we TRC_WARNING here?  It feels so unlikely that it might be worth raising.

mirw: Actually, the far more common reason for this happening is data contention (as you hit), so maybe it's better just to update the comment to reflect that?

mirw:
Looks good.  Just one comment about commenting.

mirw:
Once you've applied this fix, please can you merge into sto1119?

Issue #1346, opened 2016-03-14
Sprout does not apply default handling to ASs if it has received a 100 trying.


AlexHockey:
TS 24.229 states that an S-CSCF should apply default handling if it:

> a)	fails to receive a SIP response within a configurable time; or
> b)	receives a 408 (Request Timeout) response or a 5xx response from the AS without previously     receiving a 1xx response to the original SIP request, and without previously receiving a SIP request     from the AS that contained the same original dialog identifier as the original request;

Sprout does not implement the bit that says "without previously receiving a 1xx response to the original SIP request, and without previously receiving a SIP request from the AS that contained the same original dialog identifier as the original request;". 

This means that if a UE sends a 5xx response that gets passed back to sprout via an AS, sprout will apply default handling when it shouldn't.

mirw:
Flagging as critical because this is call-service affecting.  @JimBunch, please shout if you disagree!

JimBunch:
I agree.

AlexHockey:
I have an update on this issue. On closer inspection of the code it looks like sprout treats an AS as responsive (so won't apply default handling) if:

* It gets a request from the AS with the same ODI token it put on the request to the AS (this is correct). 
* It gets a 100 trying back from the AS (this is wrong as IMS does not treat 100 trying as a 1xx response - yes, really). 

If an AS sent a 100 response followed by a 5xx, the specs say the AS should have default handling applied to it, but sprout would not do so. This means that actually sprout applies default handling less frequently than it should.

AlexHockey:
This is potentially quite a big bug fix (~10 hours) as it will involve re-doing the AS alarm S-CSCF UTs which will break when the fix is implemented.

PR #1373, opened 2016-03-30
Fix up sprout's logic for determining if an AS is responsive

AlexHockey:
Hi Rob, please can you review my fix for #1346. This also involves a change to when the S-CSCF calls into the AS comm tracker to make it consistent with the rest of default handling processing. 

As part of this PR, I have removed a number of expectations on the AS communication tracker being called. In retrospect I think adding these to every test was a mistake as they make them too fragile. My change broke most of the tests because they did not include a non-100 response flow, which meant the test treated the flow as an AS failure (when the transaction was torn down at the end of the test). I don't want to 1xx flows to every test case as most of them don't really care about the response. However I have kept the expectations in tests that do interesting things with default handling, and the mainline tests. 

I'm not planning on testing live as I don't think this will add anything beyond the UTs.

### DIFF
--- a/src/aschain.cpp
+++ b/src/aschain.cpp
@@ -230,8 +230,8 @@ void AsChainLink::on_response(int status_code)
   }
   else if (status_code < PJSIP_SC_OK)
   {
-    // A 1xx response (which does *not* include 100) means that the AS should
-    // be treated as responsive.
+    // A 1xx response (which does *not* include 100 - see TS 24.229 section 3.2)
+    // means that the AS should be treated as responsive.
     _as_chain->_responsive[_index] = true;
   }
   else
--- a/src/aschain.cpp
+++ b/src/aschain.cpp
@@ -226,13 +226,17 @@ void AsChainLink::on_response(int status_code)
 {
   if (status_code == PJSIP_SC_TRYING)
   {
-    // The AS has returned a 100 Trying response, which means it must be
-    // viewed as responsive.
+    // Intentionally do nothing on a 100 trying.
+  }
+  else if (status_code < PJSIP_SC_OK)
+  {
+    // A 1xx response (which does *not* include 100) means that the AS should
+    // be treated as responsive.
     _as_chain->_responsive[_index] = true;
   }
-  else if (status_code >= PJSIP_SC_OK)
+  else
   {
-    // Store the status code returned by the AS.
+    // Final response. Store the status code returned by the AS.
     _as_chain->_as_info[_index].status_code = status_code;
   }
 }
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -365,7 +365,6 @@ SCSCFSproutletTsx::SCSCFSproutletTsx(SproutletTsxHelper* helper,
   _record_routed(false),
   _req_type(req_type),
   _seen_1xx(false),
-  _seen_response(false),
   _impi(),
   _auto_reg(false),
   _se_helper(stack_data.default_session_expires)
@@ -631,19 +630,18 @@ void SCSCFSproutletTsx::on_rx_response(pjsip_msg* rsp, int fork_id)
       }
       else
       {
-        // Default handling will not be triggered.
-        if (!_seen_response)
+        // Default handling will not be triggered. If this is the first non-100
+        // response we've seen from an AS track it as a successful
+        // communication. This means that no matter how many 1xx responses we
+        // receive we only track one success.
+        if ((st_code > 100) && (!_seen_1xx))
         {
-          // This is the first response we've seen from the AS, so track this is
-          // as a successful communication.
           _scscf->track_app_serv_comm_success(_as_chain_link.uri(),
                                               _as_chain_link.default_handling());
         }
       }
     }
   }
-
-  _seen_response = true;
 
   if (st_code > 100)
   {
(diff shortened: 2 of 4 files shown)


src/aschain.cpp
```diff
@@ -226,13 +226,17 @@ void AsChainLink::on_response(int status_code)
 {
   if (status_code == PJSIP_SC_TRYING)
   {
-    // The AS has returned a 100 Trying response, which means it must be
-    // viewed as responsive.
+    // Intentionally do nothing on a 100 trying.
+  }
+  else if (status_code < PJSIP_SC_OK)
+  {
+    // A 1xx response (which does *not* include 100) means that the AS should
```
rkd-msw: Spec reference for it not including 100?

src/aschain.cpp
```diff
@@ -226,13 +226,17 @@ void AsChainLink::on_response(int status_code)
 {
   if (status_code == PJSIP_SC_TRYING)
```
rkd-msw: Is `if ((status_code > PJSIP_SC_TRYING) && (status_code < PJSIP_SC_OK)` a clearer test, rather than this empty if branch?

src/scscfsproutlet.cpp
```diff
@@ -631,11 +630,12 @@ void SCSCFSproutletTsx::on_rx_response(pjsip_msg* rsp, int fork_id)
       }
       else
       {
-        // Default handling will not be triggered.
-        if (!_seen_response)
+        // Default handling will not be triggered. If this is the first non-100
+        // response we've seen from an AS track it as a successful
+        // communication. This means that no matter how many 1xx responses we
+        // receive we only track one success.
+        if ((st_code > 100) && (!_seen_1xx))
```
rkd-msw: Use PJSIP_SC_TRYING here too?

src/ut/scscf_test.cpp
```diff
@@ -3645,11 +3637,136 @@ TEST_F(SCSCFTest, DefaultHandlingContinueResponsiveError)
...
+  ASSERT_NO_FATAL_FAILURE(r2.matches(out));
+
+  tpBono.expect_target(current_txdata(), false);
+  EXPECT_EQ("sip:wuntootreefower@10.114.61.213:5061;transport=tcp;ob", r2.uri());
+  EXPECT_EQ("", get_headers(out, "Route"));
+
+  free_txdata();
+
+  // ---------- AS1 now rejects the request with a 500 response.  The AS is not
+  // bypassed as the INVITE it sent back to sprout indicates that it is live.
```
rkd-msw: Do we assert anywhere that the AS is bypassed? (In other words, how would this test fail if the AS wasn't bypassed?)

src/ut/scscf_test.cpp
```diff
@@ -2247,9 +2257,157 @@ TEST_F(SCSCFTest, SimpleISCMainline)
...
+
+  // Target sends back 100 Trying
+  inject_msg(respond_to_txdata(current_txdata(), 100), &tpBono);
+
+  // Build the two responses we are going to send.
+  std::string resp_180 = respond_to_txdata(current_txdata(), 180);
+  std::string resp_200 = respond_to_txdata(current_txdata(), 200);
+  free_txdata();
+
+  // Send a 180 ringing back down the chain to finish the transaction. This is a
```
rkd-msw: 180 Ringing doesn't finish a transaction.

src/ut/scscf_test.cpp
```diff
@@ -2247,9 +2257,157 @@ TEST_F(SCSCFTest, SimpleISCMainline)
...
+  // more realistic test of AS communication tracking.
+  for (int ii = 0; ii < 2; ++ii)
+  {
+    inject_msg(resp_180);
+
+    ASSERT_EQ(1, txdata_count());
+    RespMatcher(180).matches(current_txdata()->msg);
+
+    char buf[0xffff];
```
rkd-msw: ?????

src/ut/scscf_test.cpp
```diff
@@ -3647,7 +4004,8 @@ TEST_F(SCSCFTest, DefaultHandlingContinue1xxThenError)
 }
 
 
-// Test DefaultHandling=CONTINUE for a responsive AS that returns an error.
+// Test DefaultHandling=CONTINUE for a responsive AS that passes the INVITE
+// back tot he S-CSCF but then returns an error.
```
rkd-msw: to the

src/ut/scscf_test.cpp
```diff
@@ -403,6 +403,39 @@ class SCSCFTest : public SipTest
...
+    std::string curr_response = response;
+
+    for (int ii = 0; ii < num_hops; ++ii)
+    {
+      inject_msg(curr_response);
+
+      ASSERT_EQ(1, txdata_count());
+      RespMatcher(status_code).matches(current_txdata()->msg);
+
+      char buf[0xffff];
```
rkd-msw: ?????

rkd-msw:
Looks good.

src/ut/scscf_test.cpp
AlexHockey: I think you mean "how would the test fail if the AS *were* bypassed". Good spot, I should assert the outbound message queue is empty at the end of the test.

rkd-msw: Yes, I did.

src/ut/scscf_test.cpp
AlexHockey: It's a character buffer for printing the message to, obvs ;-)

Issue #1325, opened 2016-02-19
Possible deadlock - worker thread and transport thread can take _trying_timer_lock and other locks in different orders


rkd-msw:
After doing some analysis with Helgrind, we think some of Sprout's locks are only taken on the transport thread (which makes them unnecessary):
- the connection manager lock
- the trying_timer_lock on a SIP transaction

We've seen warnings in Helgrind about the lock ordering being wrong for these locks (which could cause deadlock, if more than one thread took them).

We should remove these locks and replace them with a check that we're running on the PJSIP transport thread. (We should output an error log if not, since using actual assertins would risk crashing production systems, which we don't want.)

JimBunch:
Marked as critical because it will be service affecting if Sprout deadlocks.

rkd-msw:
OK, I've experimented with this and the checks have done their job - they've confirmed that neither of these locks are exclusively taken on the PJSIP transport thread.

The `_trying_timer_lock` can be taken on a worker thread, when `on_tsx_request` calls exit_context and that decrements the last reference to the group lock:

```
/usr/share/clearwater/bin/sprout(_ZN6Logger9backtraceEPKc+0x6d)[0x4dbe5d]
/usr/share/clearwater/bin/sprout(_ZN3Log9backtraceEPKcz+0x10d)[0x5883dd]
/usr/share/clearwater/bin/sprout(_Z14signal_handleri+0x2c)[0x5e03cc]
/lib/x86_64-linux-gnu/libc.so.6(+0x36d40)[0x7f90dc119d40]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0x39)[0x7f90dc119cc9]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x148)[0x7f90dc11d0d8]
/lib/x86_64-linux-gnu/libc.so.6(+0x2fb86)[0x7f90dc112b86]
/lib/x86_64-linux-gnu/libc.so.6(+0x2fc32)[0x7f90dc112c32]
/usr/share/clearwater/bin/sprout[0x5a1045]
/usr/share/clearwater/bin/sprout(_ZN10BasicProxy6UASTsxD1Ev+0x34)[0x5a1084]
/usr/share/clearwater/bin/sprout(_ZN14SproutletProxy6UASTsxD1Ev+0x12a)[0x5bdaea]
[...]

PR #1376, opened 2016-03-31
Make sure the trying timer is not running when deleting a UASTsx. 

AlexHockey:
Hi Rob, please can you review my fix for #1325? This builds on your earlier PR (#1344) which I've already reviewed, so it probably makes sense for you to review my final commit. 

I have checked that:

* The UTs all pass successfully. 
* The live tests work and the error log is not produced.

### DIFF
--- a/include/basicproxy.h
+++ b/include/basicproxy.h
@@ -256,7 +256,6 @@ protected:
 
     pj_timer_entry       _trying_timer;
     static const int     TRYING_TIMER = 1;
-    pthread_mutex_t      _trying_timer_lock;
 
     friend class UACTsx;
   };
--- a/include/basicproxy.h
+++ b/include/basicproxy.h
@@ -211,6 +211,10 @@ protected:
     /// and vice-versa.  This must be called before destroying either transaction.
     void dissociate(UACTsx* uac_tsx);
 
+    /// Unbind this UASTSx from the PJSIP transaction (if any) and cancel any
+    /// timers that should only be running when the UASTsx is bound.
+    void unbind_from_pjsip_tsx();
+
     /// Creates a new downstream UACTsx object for this transaction.
     virtual BasicProxy::UACTsx* create_uac_tsx(size_t index);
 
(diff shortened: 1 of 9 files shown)


src/basicproxy.cpp
```diff
@@ -438,19 +438,41 @@ BasicProxy::UASTsx::UASTsx(BasicProxy* proxy) :
...
+
+  if (_tsx != NULL)
+  {
+    // Unbind from the PJSIP transaction.
+    _proxy->unbind_transaction(_tsx);
+    _tsx = NULL;
+
+    // The trying timer should be running when we have a PJSIP transaction, so
+    // cancel it as well.
+    cancel_trying_timer();
```
rkd-msw: As this is the only place where `cancel_trying_timer` is called, can we just inline it? I'm nervous about someone calling it from some other place and unwittingly violating our safety invariant.

If you'd like to keep it as a separate function, can we stick `CHECK_PJ_TRANSPORT_THREAD();` in there too/instead?

include/basicproxy.h
```diff
@@ -211,6 +211,10 @@ class BasicProxy
     /// and vice-versa.  This must be called before destroying either transaction.
     void dissociate(UACTsx* uac_tsx);
 
+    /// Unbind this UASTSx from the PJSIP transaction (if any) and cancel any
+    /// timers that should only be running when the UASTsx is bound.
+    void unbind_from_pjsip_tsx();
```
rkd-msw: "unbind" doesn't feel like the best name to me, but I can't think of a better one.

rkd-msw:
Looks good.

src/basicproxy.cpp
AlexHockey: I like the idea of inlining, especially as the code to start the timer is also inlined.

AlexHockey:
I'm not merging this into release-94-fixes, so I'm going to close the PR and make a new one into master.

PR #1377, opened 2016-03-31
[Reviewer: Rob] Fix possible deadlock in quiescing

### LINKED ISSUE #1351: Possible deadlock - worker thread and quiescing thread can take connection manager lock and transport lock in different orders
Split out from https://github.com/Metaswitch/sprout/issues/1325 to keep issues separate.

johadalin:
This merges the quiescing thread with the PJSIP transport thread. 
This should avoid a possible deadlock condition. 

The relocation of stack.cpp:L714/753 is to fix crashes that were seen on stopping or quiescing sprout. 

These changes have been tested on a live deployment, and quiescing shows no crashing or locking.
Once https://github.com/Metaswitch/sprout/pull/1344 is merged, it may be worth adding in CHECK_PJ_TRANSPORT_THREAD() before running the quiesce calls, to ensure we are being safe.

Will fix #1351

### DIFF
--- a/include/connection_tracker.h
+++ b/include/connection_tracker.h
@@ -84,9 +84,6 @@ public:
   void unquiesce();
 
 private:
-  // This must be held when accessing any of this object's member variables.
-  pthread_mutex_t _lock;
-
   // A map of all the connections known to the connection manager, and their
   // state listeners.  This is a set of pjsip transports, but only includes
   // connection-based transports (not datagram transports).
--- a/include/stack.h
+++ b/include/stack.h
@@ -134,6 +134,26 @@ inline SAS::TrailId get_trail(const pjsip_transaction* tsx)
   return (SAS::TrailId)tsx->mod_data[stack_data.sas_logging_module_id];
 }
 
+/*
+// Variable to store our current quiescing state
+static pj_bool_t quiescing = PJ_FALSE;
+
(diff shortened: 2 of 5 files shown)


johadalin:
Closing, as should be against fixes branch, not dev

johadalin:
I've added CHECK_PJ_TRANSPORT_THREAD at the start of the pjsip_tread_function. 
@rkd-msw are you happy where i've put it? 

If so, i'll merge this into dev and close #1351

rkd-msw:
Sadly I'm not - `pjsip_thread_func` is by definition always on the PJSIP transport thread (it *is* the PJSIP transport thread). Can you put `CHECK_PJ_TRANSPORT_THREAD` at the top of the ConnectionTracker methods instead?

johadalin:
Removed the CHECK_PJ_TRANSPORT_THREAD from the previous location, and added it in place of the lock in connection tracker methods. 

Tested on a live deployment by adding a debug log alongside the see when the methods were called, and ensuring that the CHECK did not raise any errors when in normal use, and when stopped or quiesced.

Does this now look good to merge in?

rkd-msw:
Can you also remove the lock from the header file (https://github.com/Metaswitch/sprout/blob/dev/include/connection_tracker.h#L87)? Then good to merge.

PR #1378, opened 2016-03-31
[Reviewer: Rob] Fix possible deadlock in quiescing

### LINKED ISSUE #1351: Possible deadlock - worker thread and quiescing thread can take connection manager lock and transport lock in different orders

johadalin:
[Supercedes #1377]
This merges the quiescing thread with the PJSIP transport thread. 
This should avoid a possible deadlock condition.

These changes have been tested on a live deployment, and quiescing shows no crashing or locking.
Once #1344 is merged, it may be worth adding in CHECK_PJ_TRANSPORT_THREAD() before running the quiesce calls, to ensure we are being safe.

Will fix #1351

### DIFF
--- a/src/stack.cpp
+++ b/src/stack.cpp
@@ -148,6 +148,7 @@ const static std::string _known_statnames[] = {
 const std::string* known_statnames = _known_statnames;
 const int num_known_stats = sizeof(_known_statnames) / sizeof(std::string);
 
+// Variable to hold quiescing state, and interfaces to alter it
 static pj_bool_t quiescing = PJ_FALSE;
 
 extern void set_quiescing_true()
@@ -183,17 +184,15 @@ static int pjsip_thread_func(void *p)
     new_quiescing = quiescing;
     if (curr_quiescing != new_quiescing)
     {
-      TRC_DEBUG("quiescing state changed");
+      TRC_DEBUG("Quiescing state changed");
       curr_quiescing = new_quiescing;
 
       if (new_quiescing)
       {
-        TRC_DEBUG("calling quiescing manager quiesce");
         quiescing_mgr->quiesce();
       }
       else
       {
-       TRC_DEBUG("calling quiescing manager unquiesce");
        quiescing_mgr->unquiesce();
       }
     }
--- a/src/stack.cpp
+++ b/src/stack.cpp
@@ -148,6 +148,20 @@ const static std::string _known_statnames[] = {
 const std::string* known_statnames = _known_statnames;
 const int num_known_stats = sizeof(_known_statnames) / sizeof(std::string);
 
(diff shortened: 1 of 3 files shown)


johadalin:
Moving crash fix to a separate PR

src/stack.cpp
```diff
@@ -158,9 +173,30 @@ static int pjsip_thread_func(void *p)
...
   TRC_DEBUG("PJSIP thread started");
 
+  pj_bool_t curr_quiescing = PJ_FALSE;
+  pj_bool_t new_quiescing = quiescing;
+
   while (!quit_flag)
   {
     pjsip_endpt_handle_events(stack_data.endpt, &delay);
+
+    // Check if our quiescing state has changed, and act appropriately
```
rkd-msw: It might be nice if this code lived inside ConnectionTracker, so that this code could just be "ConnectionTracker::check_for_quiescing_state_change()`. I don't have strong feelings about whether this is a member function (so you'd need to pass a ConnectionTracker object around), or a class function (which feels easier to implement).

rkd-msw:
One comment on the code.

One comment on your testing: did you quiesce with transactions in progress?

And a couple of comments on your git-fu:

- rather than reverting e59d28a, I'd have reset the branch head to 12a44e8 (`git reset --hard 		12a44e8`, I think)
- rather than leaving a commit with the message "Messy interim commit stash", I'd run `git rebase -i release-94-fixes` and squash the two commits into one commit with a message like 'Fix possible deadlock in quiescing'

johadalin:
Tested:
  * Quiescing Sprout with one sofphone calling another. Sprout does not quiesce until the transaction closes.
  * Quiescing Bono with a UE registered. UE correctly receives 305 Use Proxy messages, and Bono quiesces after 15 minutes (as expected. see #1296)

johadalin:
@rkd-msw Can you help me assess the risk in merging this in to the CCv9 release?
I do not believe it will lead to any instability, but would like a second opinion to take to @JimBunch

rkd-msw:
In favour of this change:

* this change is very localised - I can't see it causing issues in any case other than quiescing
* you've done some sensible mainline testing which has passed

but on the other hand:

* this deadlock has been possible since we added quiescing - September 2013 - but as far as we know we've never hit it
* if we did hit it, Monit polls would start failing and we'd kill the process after ~30 seconds - it wouldn't be permanently deadlocked

So I agree that the risk is low, but I suspect the benefit of merging it is also relatively low, and possibly too low to justify breaking a code freeze.

johadalin:
Agreed with @JimBunch not to merge this into the fixes branch. 
Closing this PR and reopening #1377

PR #1380, opened 2016-04-01
[Reviewer: Rob] Stop crashes on process stop or quiesce

johadalin:
When running `sudo service sprout stop` or `sudo service sprout quiesce`, we were seeing the process crash with a signal 11. 

This change should stop this happening.
I have tested this alongside PR #1378 and am in the process of building it to test separately to ensure it fixes the issue on its own. 
This will be a critical fix.

### DIFF
--- a/src/stack.cpp
+++ b/src/stack.cpp
@@ -711,10 +711,10 @@ pj_status_t init_stack(const std::string& system_name,
        ++ii)
   {
     pjsip_tpfactory* tcp_factory = NULL;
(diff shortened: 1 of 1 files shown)


rkd-msw:
Could you raise an issue describing the problem as well? Fix looks good.

johadalin:
Fixes #1382

johadalin:
Agreed merge with @JimBunch

PR #1383, opened 2016-04-01
[Reviewer: Ellie] Add port to the default Sproutlet URIs

rkd-msw:
This matches the old behaviour, where the default included a port (https://github.com/Metaswitch/sprout/blob/937b339c7949415102fe1332cb9371d8c9c10c43/sprout-base.root/etc/init.d/sprout#L265).

I've run make to confirm it compiles, and I'll test immediately on staging-aio.


eleanor-merry:
LGTM

PR #1364, opened 2016-03-22
[Reviewer: Rob] Add new sproutlet configuration options

eleanor-merry:
Rob, can you review this change to add new sproutlet configuration options.

Tested live (checking that I can set the different configuration options, and using the live tests to check that calls still work)

PR #1387, opened 2016-04-04
Sproutlet subscribe routing

Neil-Bruce:
Route subscribe messages over external port rather than sproutlet interface to allow subscription module to process them.

Fixes https://github.com/Metaswitch/sprout/issues/1389

### DIFF
--- a/mk/sprout.mk
+++ b/mk/sprout.mk
@@ -19,6 +19,9 @@ sprout_test:
 sprout_full_test:
 	${MAKE} -C ${SPROUT_DIR} full_test
 
+sprout_runtest:
+	${MAKE} -C ${SPROUT_DIR} run_test
+
 sprout_clean:
 	${MAKE} -C ${SPROUT_DIR} clean
 
--- a/mk/sprout.mk
+++ b/mk/sprout.mk
@@ -19,9 +19,6 @@ sprout_test:
 sprout_full_test:
 	${MAKE} -C ${SPROUT_DIR} full_test
 
-sprout_runtest:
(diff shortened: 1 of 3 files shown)


rkd-msw:
Hi Neil,

I see that this is based on https://github.com/ClearwaterCore/sprout/pull/44, but it’s not entirely clear what the issue you’ve hit is – could you raise an issue (https://github.com/Metaswitch/sprout/issues/new) describing the symptoms, and with a SAS trace if possible?

In general it's good to have an issue raised for every bug that gets fixed (so that we know the symptoms and can judge how severe the impact is, and so we can communicate to users that this is fixed) - and in this particular case, we’ve recently done some quite substantial work in a related area (https://github.com/Metaswitch/sprout/pull/1364), so if we know the flow that's a problem, we can check if it repros in the latest release.

Neil-Bruce:
I've raised Issue https://github.com/Metaswitch/sprout/issues/1389 for this bug.  

Andy and Ellie have already taken a look at the code changes but I am waiting for Ellie to answer a question since the code has changed to remove some similar code recently.

mk/sprout.mk
```diff
@@ -19,6 +19,9 @@ sprout_test:
 sprout_full_test:
 	${MAKE} -C ${SPROUT_DIR} full_test
 
+sprout_runtest:
```
eleanor-merry: run_test doesn't exist as a target

Neil-Bruce: Looks like the makefiles have been restructured, I'll remove this change it was just for convenience anyway.

eleanor-merry:
LGTM

PR #1391, opened 2016-04-08
[Reviewer: ANDY] Fix up Subscribe routing

eleanor-merry:
This fixes up the Gemini Subscribe live tests

The change has been to make sure that the selected sproutlet is only rejected if the subscribe would be absorbed by the subscription module. Tested in UTs and the live tests

### DIFF
--- a/include/subscription.h
+++ b/include/subscription.h
@@ -56,6 +56,8 @@ extern pj_status_t init_subscription(SubscriberDataManager* sdm,
                                      AnalyticsLogger* analytics_logger,
                                      int cfg_max_expires);
 
+extern pj_bool_t request_acceptable_to_subscription_module(pjsip_msg* msg,
+                                                           SAS::TrailId trail);
 extern void destroy_subscription();
 
 #endif
--- a/include/subscription.h
+++ b/include/subscription.h
@@ -49,15 +49,15 @@ extern "C" {
 
 extern pjsip_module mod_subscription;
 
-extern pj_status_t init_subscription(SubscriberDataManager* sdm,
-                                     SubscriberDataManager* remote_sdm,
-                                     HSSConnection* hss_connection,
-                                     ACRFactory* rfacr_factory,
-                                     AnalyticsLogger* analytics_logger,
-                                     int cfg_max_expires);
+pj_status_t init_subscription(SubscriberDataManager* sdm,
+                              SubscriberDataManager* remote_sdm,
+                              HSSConnection* hss_connection,
+                              ACRFactory* rfacr_factory,
+                              AnalyticsLogger* analytics_logger,
+                              int cfg_max_expires);
 
-extern pj_bool_t request_acceptable_to_subscription_module(pjsip_msg* msg,
-                                                           SAS::TrailId trail);
-extern void destroy_subscription();
+pj_bool_t request_acceptable_to_subscription_module(pjsip_msg* msg,
+                                                    SAS::TrailId trail);
+void destroy_subscription();
 
 #endif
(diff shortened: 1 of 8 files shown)


include/subscription.h
```diff
@@ -56,6 +56,8 @@ extern pj_status_t init_subscription(SubscriberDataManager* sdm,
                                      AnalyticsLogger* analytics_logger,
                                      int cfg_max_expires);
 
+extern pj_bool_t request_acceptable_to_subscription_module(pjsip_msg* msg,
```
bossmc: Why is this `extern`?

include/subscription.h
```diff
@@ -56,6 +56,8 @@ extern pj_status_t init_subscription(SubscriberDataManager* sdm,
...
+                                                           SAS::TrailId trail);
 extern void destroy_subscription();
```
bossmc: And this?

include/subscription.h
bossmc: According to Wikipedia:

> The `extern` keyword applied to a function prototype does absolutely nothing

src/sproutletproxy.cpp
```diff
@@ -520,9 +526,15 @@ pj_status_t SproutletProxy::UASTsx::init(pjsip_rx_data* rdata)
...
         // There is a top Route header in the request, which by definition
         // caused the request to be routed to this node, so remove it and
-        // allow the request to be forwarded.
-        TRC_INFO("Remove top Route header and forward request");
-        pj_list_erase(route);
+        // allow the request to be forwarded. The exception here is if we
+        // rejected the sproutlet selection - which means that we want to
+        // route to the same location but route externally to get the
+        // registrar/subscription modules
+        if (!sproutlet_rejected)
```
bossmc: I'd have preferred `force_external_routing` to make it clear why the variable exists.

src/subscription.cpp
```diff
@@ -704,6 +697,7 @@ pj_bool_t subscription_on_rx_request(pjsip_rx_data *rdata)
       char accept_hdr_str[256];
       memset(accept_hdr_str, 0, 256);
       pjsip_hdr_print_on(accept, accept_hdr_str, 255);
+
       SAS::Event event(trail, SASEvent::SUBSCRIBE_FAILED_EARLY_ACCEPT, 0);
```
bossmc: Does this SAS event make sense in both places this is now called?  I think we had a recent issue where we'd re-used a SAS event in a new place and it was confusing in the new usage.

src/subscription.cpp
```diff
@@ -712,10 +706,34 @@ pj_bool_t subscription_on_rx_request(pjsip_rx_data *rdata)
...
+  SAS::TrailId trail = get_trail(rdata);
+
+  if (rdata->tp_info.transport->local_name.port != stack_data.scscf_port)
+  {
+    // Not an S-CSCF, so don't handle SUBSCRIBEs.
+    return PJ_FALSE; // LCOV_EXCL_LINE
+  }
+
+  pj_bool_t accept =
+          request_acceptable_to_subscription_module(rdata->msg_info.msg, trail);
```
bossmc: I like this shared check function design.  :+1:

bossmc:
Not much to say, looks functionally good, couple of maintainability/diagnosability comments.

src/subscription.cpp
eleanor-merry: The SAS logs won't be made in the sproutlet routing case

src/sproutletproxy.cpp
eleanor-merry: Sure

PR #1393, opened 2016-04-11
[Reviewer: Matt] Iss1388

### LINKED ISSUE #1388: SAS isn't clear which sproutlet it's invoking (and why)
#### Symptoms

We were investigating issues with routing in Sprout.  In particular, we had 2 Sproutlets attached to the same port, and Sprout was selecting randomly (and choosing BGCF over S-CSCF).

It was very hard to tell which Sproutlet Sprout was selecting, and why.  SAS should tell us this.

In particular, I think Sprout should log
*   when it starts searching for Sproutlets, highlighting the URI and port its selecting on
*   when invoking a specific Sproutlet, highlighting which and why
*   when selecting randomly between multiple available Sproutlets, highlighting which Sproutlet has been ignored
*   when finishing processing a Sproutlet.

For performance reasons, I don't think we should log the full SIP message at each point.

#### Impact

This makes it hard to troubleshoot routing misconfiguration.

#### Release and environment 

release-95

eleanor-merry:
Matt, can you review this change to add SAS logs for sproutlet selection, and to prioritise the S-CSCF sproutlet in ambiguous cases

Fixes #1388 

Tested using the live tests and checking SAS


src/sproutletproxy.cpp
```diff
@@ -186,45 +193,73 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
...
            it != _sproutlets.end();
            ++it)
       {
         if ((*it)->port() == port)
         {
-          sproutlet = *it;
-          alias = (*it)->service_name();
-          break;
+          if ((!sproutlet_set) ||
+              ((*it)->service_name() == _scscf_name))
```
mirw: Can't we just sort the S-CSCF sproutlet to the front of the list?  That would mean that we don't need to pollute all the code with knowing about the S-CSCF name explicitly.

src/sproutletproxy.cpp
```diff
@@ -186,45 +193,73 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
...
   // route the message to the Subscription module instead
   if ((sproutlet) &&
-      (sproutlet->service_name() == "scscf") &&
+      (sproutlet->service_name() == _scscf_name) &&
       (request_acceptable_to_subscription_module(req, 0)))
   {
-    // This is a subscribe being routed back to the S-CSCF sproutlet but to 
-    // get this handled correctly we need route it externally to make sure it 
+    // This is a subscribe being routed back to the S-CSCF sproutlet but to
+    // get this handled correctly we need route it externally to make sure it
```
mirw: Typo: "we need **to** route it".

mirw:
Looks good.  A couple of comments.

src/sproutletproxy.cpp
eleanor-merry: I need it later on anyway when I check if the scscf sproutlet was selected though

mirw: Agreed, but I think it's tidier (and simpler code) here to use a sorted list - in particular, because the code below should be removed once we Sproutlet-ize subscription-handling.

PR #1397, opened 2016-04-13
[Reviewer: Matt] Remove misleading SAS log

rkd-msw:
Fixes https://github.com/Metaswitch/clearwater-sas-resources/issues/93.

Per my analysis there:
* we know when I-CSCF processing is happening because of logs like https://github.com/Metaswitch/clearwater-sas-resources/blob/master/clearwater_sas_resource_bundle.yaml#L1077
* we know that I-CSCFs don't authenticate messages
* so the "Authentication is not needed as the request wasn't received on the S-CSCF port" log isn't adding any value
* so we should just delete it


mirw:
Sounds reasonable.  Should we tidy up clearwater-sas-resources.yaml too?

rkd-msw:
Done in https://github.com/Metaswitch/clearwater-sas-resources/pull/96.

mirw:
Fine.

Issue #1394, opened 2016-04-12
calls fail with mixed versions of sprout


dtheobald:
#### Symptoms

During an upgrade from release-82 to release-94 when in a configuration where one sprout was running release-82 and the other sprout was running release-94, I get spurious call failures (480 temporarily unavailable) even though the UE has registered.

#### Impact

Calls fail.

#### Release and environment 

Mix of release-82 and release-94 sprout nodes

#### Steps to reproduce
1) Add ```impi_store_mode=av-impi``` to shared_config and distribute.
2) Upgrade homestead nodes to release-94
3) Upgrade one sprout node to release-94
4) Register a couple of UEs and make calls between them. Invariably a call will fail with 480.

Diags are available (sprout-1 is running release-94)

mirw:
@dtheobald, I can't see a SAS trace for the failed call.  Your 480call.sas file actually contains a REGISTER.  Please can you check and point me to the correct diagnostics?

johadalin:
I've put a couple of traces up with the diags from earlier failures Doug sent to me.

mirw:
I've looked at the SAS trace.  The failure appears to be that Homestead sends a Location-Information-Request and gets back a valid but empty Location-Information-Answer.  This suggests that the HSS believes that the subscriber is unregistered.

This looks like a very similar problem to https://github.com/Metaswitch/homestead/issues/325 - in that case, a UAR was rejected because the subscriber was unregistered, despite us believing the subscriber to have been successfully registered.

Either the HSS is randomly dropping registrations, or something else is requesting it to clear them out.  Getting **all** the SAS traces for this subscriber around the time of the failure would show whether something is requesting to clear them out, but an alternative would be to run a test with network capture on the HSS, filtering on TCP port 3868.

@dtheobald, please can you gather these diagnostics?

mirw:
Marking as critical on the basis that this might be a Clearwater bug.  If it's an HSS bug, it's clearly not.

dtheobald:
I've updated 480call.sas to contain the correct trace. I've left MY SAS running, so you can also see the complete set of traces by opening the link.

mirw:
OK, so we now have the full picture, and there are two issues here.

*   The UE sends a REGISTER to Sprout.  The uplevel sprout creates a challenge, stories both IMPI and AV entries, and sends it back.  It also sets a Chronos timeout to timeout the authentication.
[...]

PR #1398, opened 2016-04-13
[Reviewer: Alex] Fix (de)serialization of nonce_count and branch

mirw:
Alex,

Please can you review this fix to #1394?  I've UT-ed and run clearwater-live-test against this.

I'm going to check with @JimBunch to confirm he wants this merged into release-94-fixes, but I think he will.

Thanks,

Matt

### DIFF
--- a/src/impistore.cpp
+++ b/src/impistore.cpp
@@ -218,7 +218,8 @@ std::string ImpiStore::AuthChallenge::to_json_av()
 
 void ImpiStore::AuthChallenge::write_json_av(rapidjson::Writer<rapidjson::StringBuffer>* writer)
 {
-  // Write the outer base AuthChallenge fields to JSON, in the AV format.  This is just the correlator.
+  // Write the outer base AuthChallenge fields to JSON, in the AV format.  This
+  // is just the correlator.
   if (correlator != "")
   {
     writer->String(JSON_AV_CORRELATOR); writer->String(correlator.c_str());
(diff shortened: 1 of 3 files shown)


src/impistore.cpp
```diff
@@ -218,9 +218,7 @@ std::string ImpiStore::AuthChallenge::to_json_av()
 
 void ImpiStore::AuthChallenge::write_json_av(rapidjson::Writer<rapidjson::StringBuffer>* writer)
 {
-  // Write all the base AuthChallenge fields to JSON, in the AV format.
-  writer->String(JSON_AV_NONCE_COUNT); writer->Uint(nonce_count);
-  writer->String(JSON_AV_EXPIRES); writer->Int(expires);
+  // Write the outer base AuthChallenge fields to JSON, in the AV format.  This is just the correlator.
```
AlexHockey: And the tombstone too, right?

src/impistore.cpp
```diff
@@ -237,6 +235,25 @@ void ImpiStore::AuthChallenge::write_json_av(rapidjson::Writer<rapidjson::String
...
 }
 
+void ImpiStore::AuthChallenge::write_json_av_inner(rapidjson::Writer<rapidjson::StringBuffer>* writer)
+{
+  // Write the inner base AuthChallenge fields to JSON, in the AV format.
+  // Nonce count is a bit complicated.  If the tombstone flag is set (and it
+  // will be if nonce count is greater than the initial value), we need to
+  // store the last-used nonce count, rather than the next acceptable one.  We
+  // calculate this by decrementing the nonce count.  We undo this when we
+  // read the AV back in from_json_av.
```
AlexHockey: OK. I think the situation that this is trying to solve is:

* Uplevel node is in av-impi mode, generates a challenge and writes an AV with a nonce count of 1.
* Downlevel node processes the authentication response, and tombstones the AV but does not increment the nonce count (because the downlevel node does not care about nonce counts).
* If an uplevel node reads this AV back it knows the nonce count should have been incremented but wasn't, so increments the value it reads back before storing in the `AuthChallenge` object.
* Because of this processing, an uplevel node also needs to decrement the nonce count when writing AVs if the nonce count has ever been incremented. 

That took me a while to work through - could we have a comment saying as much? (feel free to steal this text obviously :-)

AlexHockey:
Looks good, my main comment is it would be good to explain the nonce count increment/decrement a bit more. 

You ran clearwater-live-test. Did you also check SAS trace to see if it fixes the bug we were seeing?

src/impistore.cpp
mirw: I don't see that as a base field (it's not part of the object at all), and it's covered at length in the following comment block, so I didn't include it - OK?

mirw:
Live test doesn't cover software upgrade, so I wasn't able to check the SAS, but I'll ask Doug to.

PR #1402, opened 2016-04-18
[Reviewer: Rob] Change max_tokens default to 100

johadalin:
Changing the default value of max_tokens in Sprout. 
Changes for the other repos are in the PRs below. Only the Chronos and docs changes are at all different, and I just want you to double check the Chronos one:

* Ralf: https://github.com/Metaswitch/ralf/pull/218
* Homestead: https://github.com/Metaswitch/homestead/pull/331
* Memento: https://github.com/Metaswitch/memento/pull/130
* Chronos: https://github.com/Metaswitch/chronos/pull/252
* Docs: https://github.com/Metaswitch/clearwater-readthedocs/pull/167

Do you think we also want to increase the value used at https://github.com/Metaswitch/cpp-common-test/blob/master/src/ut/load_monitor_test.cpp#L62 to reflect this change?


rkd-msw:
Looks good. I don't think we want to change the UT value -  that's about testing that the load monitor code works, not that particular input values are good. (It's already got a different initial token rate, for example.)

rkd-msw:
That said, should this be against release-94-fixes?

johadalin:
Closing and moving to a PR against release-94 fixes

PR #1410, opened 2016-04-21
[Reviewer: Matt] Remove restriction

eleanor-merry:
Don't enforce that a Sprout node must be a P-CSCF/I-CSCF/S-CSCF


mirw:
Looks good.

Issue #1408, opened 2016-04-21
Sprout (and memento) crash on stop


eleanor-merry:
#### Symptoms
Stopping Sprout and Memento causes them to crash (signal 11). Sadly, I don't get any stack trace in the logs; all I get is:

```
Signal 11 caught

Basic stack dump:
```

Looking at the core file for memento I get:

```
#0  0x00007f895c863688 in ?? () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#1  0x00007f895c8646f8 in _Unwind_Backtrace () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#2  0x00007f895c599e26 in __GI___backtrace (array=<optimized out>, size=32) at ../sysdeps/x86_64/backtrace.c:109
#3  0x00000000004313cb in Logger::backtrace (this=0x7b4b10, data=<optimized out>) at ../modules/cpp-common/src/logger.cpp:303
#4  0x00000000004371cf in Log::backtrace (fmt=fmt@entry=0x4a9c04 "Signal %d caught") at ../modules/cpp-common/src/log.cpp:185
#5  0x000000000041a890 in signal_handler (sig=11) at main.cpp:427
#6  <signal handler called>
#7  0x0000000000000007 in ?? ()
#8  0x00000000004158af in main (argc=<optimized out>, argv=<optimized out>) at main.cpp:713
```

eleanor-merry:
This looks like a double delete (but I'm not sure why this isn't more obvious in the logs). 

The communication monitor is passed an alarm, and it then deletes that alarm in its destructor (and doesn't set the alarm to null) - https://github.com/Metaswitch/cpp-common/blob/master/src/communicationmonitor.cpp#L59.

In memento/sprout we delete the communication monitor, then delete the alarms, causing the crash.

eleanor-merry:
This isn't in release-94-fixes (for Sprout)

PR #1412, opened 2016-04-22
[Reviewer: Alex] Don't free alarms

eleanor-merry:
Don't double delete alarms

Part of the fix for https://github.com/Metaswitch/sprout/issues/1408


AlexHockey:
Looks good. Should we also have a comment on the `CommunicationMonitor` flagging that it takes ownership of the alarm object?

PR #1418, opened 2016-04-27
Reviewer EM: Apply markups after evil Ellie's review comments

graemerobertson:
View the diff in split mode rather than unified mode (otherwise it's impossible to review). Sorry if that's totally obvious.


eleanor-merry:
LGTM

PR #1420, opened 2016-04-28
[Reviewer: Rob] Update alias_list, change port defaults, and pass through server name

eleanor-merry:
Matching PR to https://github.com/Metaswitch/homestead/pull/339. 

This PR populates the alias_list by default, changes the default ports for the BGCF/MMTel sproutlet, and passes the S-CSCF name to Homestead

Tested with the live-tests. I've also done some manual upgrade tests with a HSS


src/stack.cpp
```diff
@@ -785,19 +786,49 @@ pj_status_t init_stack(const std::string& system_name,
...
+
+  for (std::unordered_set<std::string>::iterator it = stack_data.aliases.begin();
+       it != stack_data.aliases.end();
+       ++it)
+  {
+    pj_strdup2(stack_data.pool, &stack_data.name[stack_data.name_cnt], it->c_str());
+    stack_data.name_cnt++;
+  }
+
+  /*for (std::vector<std::string>::iterator it = sproutlet_uris.begin();
```
rkd-msw: Commented-out code?

src/scscfsproutlet.cpp
```diff
@@ -435,6 +438,8 @@ void SCSCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
                                                            NULL);
       _impi = PJUtils::extract_username(proxy_auth_hdr,
                                         PJUtils::orig_served_user(req));
+      _server_name = PJUtils::uri_to_string(PJSIP_URI_IN_REQ_URI,
```
rkd-msw: Comment to say why we're using the Request-URI (as opposed to the configured S-CSCF URI, or indeed the top Route header) for this?

rkd-msw: (also in the registrar)

sprout-base.root/etc/init.d/sprout
```diff
@@ -97,9 +97,14 @@ get_settings()
         # Set up defaults and then pull in the settings for this node.
         sas_server=0.0.0.0
         sprout_rr_level="pcscf"
+
+        # These Sproutlets are enabled by default. They listen on different
+        # ports to help ensure that requests have a unambiguous Sproutlet
+        # to route to.
+        bgcf=5053
```
rkd-msw: Should we be checking in the code that the Sproutlet ports are unique? Suggestion only.

src/stack.cpp
```diff
@@ -785,19 +786,49 @@ pj_status_t init_stack(const std::string& system_name,
...
+       it != sproutlet_uris.end();
+       ++it)
+  {
+    pjsip_sip_uri* sproutlet_uri = (pjsip_sip_uri*)PJUtils::uri_from_string(
+                                                       *it,
+                                                       stack_data.pool,
+                                                       false);
+    if (sproutlet_uri)
+    {
+      stack_data.aliases.insert(PJUtils::pj_str_to_string(&sproutlet_uri->host));
```
rkd-msw: This adds the Sproutlet URIs as aliases, but I don't think it adds sprout_hostname, does it?

rkd-msw:
Just to check: have you explicitly tested that, if you register a subscriber, then change Sprout's S-CSCF URI, the old Server-Name continues to be used on SARs?

Otherwise good.

sprout-base.root/etc/init.d/sprout
eleanor-merry: Nope - they don't have to be unique (and making these defaults unique is to help the upgrade case)

src/stack.cpp
eleanor-merry: Nope, I'll add that too

eleanor-merry:
Yup, I've done that test

PR #1425, opened 2016-05-03
[Reviewer: Adam] Make reg tests independent

eleanor-merry:
This splits up the registrar tests so that they don't share the fake hssconnection results. This stops the later tests being reliant on previous tests running first.


johadalin:
Looks good

PR #1430, opened 2016-05-04
Reviewer MIRW: Change stack_data name array to a vector

### LINKED ISSUE #1427: stack_data's name array is not necessarily big enough
#### Symptoms
The array was originally of size 16, and we were writing more than 16 names to it which was causing us to scribble on other fields.

#### Impact
Random failures due to fields being set to gibberish.

#### Release and environment 
Release 95

#### Steps to reproduce
Add loads of aliases?

graemerobertson:
Fixes #1427.

Tested by running the UTs cleanly and running the live tests against a system running Gemini and Memento that I happened to have lying around.

I've increased the array size to 30 in release-94-fixes and I'm not intending to merge this PR to that branch.

### DIFF
--- a/include/stack.h
+++ b/include/stack.h
@@ -86,8 +86,7 @@ struct stack_data_struct
 
   int                  addr_family;
 
-  unsigned             name_cnt;
-  pj_str_t             name[30];
+  std::vector<pj_str_t> name;
   LastValueCache *     stats_aggregator;
 
   bool record_route_on_every_hop;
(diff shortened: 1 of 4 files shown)


mirw:
Looks good.  Have you checked that valgrind tests still pass - it's possible that we'll leak memory.

graemerobertson:
I have, although I'm pretty sure we never run through the relevant code in UTs so that's not very conclusive. I checked the pj_strdup2 logic with @bossmc and he gave it the all clear, so between our 3 great minds I'm feeling confident.

PR #1432, opened 2016-05-05
Reviewer EM: Only create a remote SDM if we have remote servers

### LINKED ISSUE #1428: Sprout always tries to write to the remote memcached store, even when there isn't one
#### Symptoms
Sprout starts with a remote_cluster_settings file with servers set to blank even if you don't create one yourself, and as a result of this, Sprout thinks there is a remote site and always tries to write to the remote memcached site.

#### Impact
Not a great deal of impact, we try and do unnecessary work and we get lots of rogue error logs.

#### Release and environment 
Release-95. We used to check the has_servers function, but we don't any more, which is effectively what has caused this issue.

#### Steps to reproduce
Run a call through Sprout.

graemerobertson:
This is better? We always iterate over the remote_sdms vector to use them so empty vector is superb.

Fixes #1428


eleanor-merry:
Looks good. What testing have you done?

graemerobertson:
The UTs run cleanly

eleanor-merry:
That's super good for your change to main.cpp!

PR #1433, opened 2016-05-05
[Reviewer: Graeme] Fix server name mangling

eleanor-merry:
Fixes up the server name on SARs/MARs

### DIFF
--- a/include/pjutils.h
+++ b/include/pjutils.h
@@ -289,6 +289,13 @@ bool should_update_np_data(URIClass old_uri_class,
                            std::string& new_uri_str,
                            bool should_override_npdi,
                            SAS::TrailId trail);
+
+// Get a string representation of the top routing header (or the
+// request URI if there's no route headers). This can return
+// an empty string (if the header isn't a valid URI), so callers
+// should validate the result
+std::string get_next_routing_header(pjsip_msg* msg);
+
 } // namespace PJUtils
 
 #endif
--- a/include/pjutils.h
+++ b/include/pjutils.h
@@ -293,7 +293,7 @@ bool should_update_np_data(URIClass old_uri_class,
 // Get a string representation of the top routing header (or the
 // request URI if there's no route headers). This can return
 // an empty string (if the header isn't a valid URI), so callers
-// should validate the result
+// should validate the result.
 std::string get_next_routing_header(pjsip_msg* msg);
 
 } // namespace PJUtils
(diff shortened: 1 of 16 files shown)


include/pjutils.h
```diff
@@ -289,6 +289,13 @@ bool should_update_np_data(URIClass old_uri_class,
                            std::string& new_uri_str,
                            bool should_override_npdi,
                            SAS::TrailId trail);
+
+// Get a string representation of the top routing header (or the
+// request URI if there's no route headers). This can return
+// an empty string (if the header isn't a valid URI), so callers
+// should validate the result
```
graemerobertson: Some of your comments end in full stops and some don't - maybe we could make this consistent? I'd prefer if they all did. Pls fix up across the codebase.

src/authentication.cpp
```diff
@@ -692,14 +692,21 @@ static pj_bool_t needs_authentication(pjsip_rx_data* rdata, SAS::TrailId trail)
...
-        // (Values of tls-pending or ip-assoc-pending indicate the challenge
-        // should be checked.)
-        TRC_INFO("SIP Digest authenticated request integrity protected by edge proxy");
+        // Values of tls-pending or ip-assoc-pending indicate the challenge
+        // should be checked.
 
-        SAS::Event event(trail, SASEvent::AUTHENTICATION_NOT_NEEDED_INTEGRITY_PROTECTED, 0);
-        SAS::report_event(event);
+        // We should still challenge though if we find that the request wasn't
+        // sent to this S-CSCF
```
graemerobertson: Can we have a more explicit comment of why this is the case?

src/ut/authentication_test.cpp
```diff
@@ -595,6 +598,31 @@ TEST_F(AuthenticationTest, IntegrityProtected)
   EXPECT_EQ(0,((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_AUTHENTICATION_STATS_TABLES.ims_aka_auth_tbl)->_attempts);
 }
 
+TEST_F(AuthenticationTest, IntegrityProtectedChangedSCSCF)
+{
+  // Test that the authentication module lets through REGISTER requests
```
graemerobertson: This comment is wrong.

graemerobertson:
I've reviewed, think I had a couple of comments, can't really remember. The code looked basically all fine.

You're reasoning for why this is necessary and why this should work sounded reasonable to me, but I am a little bit worried that we may have overlooked something. Hopefully you've tested thoroughly!

PR #1437, opened 2016-05-12
[Reviewer Alex] Add ENUM success/failure SAS logs to file-based ENUM

steelyvoid:
This change adds SAS logs to the file-based ENUM code, using the existing "ENUM Complete" and "ENUM Incomplete" logs that would be generated if DNS-based ENUM were in use.

I deliberately refrained from creating a separate SAS log for the "Failed to translate number with regex" case in the interests of simplicity.  Note that an Error log will be made to sprout_current.txt in this case, so the underlying cause can be determined if the operator needs to know.

### DIFF
--- a/src/enumservice.cpp
+++ b/src/enumservice.cpp
@@ -229,7 +229,10 @@ std::string JSONEnumService::lookup_uri_from_user(const std::string &user, SAS::
 
   if (pfix == NULL)
   {
-    TRC_INFO("No matching number range %s from ENUM lookup", user.c_str());
+    TRC_WARNING("No matching number range %s from ENUM lookup", user.c_str());
+    SAS::Event event(trail, SASEvent::ENUM_INCOMPLETE, 0);
+    event.add_var_param(user);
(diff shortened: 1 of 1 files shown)


AlexHockey:
I think it's fine to use the same log for the "Failed to translate number with regex" case. It might be a good idea to pass in different values for the instance ID. This is the last parameter passed into the `SAS::Event` constructor and can be used to distinguish different instances of the same log.

I think you should also generate the `ENUM_STARTED` log as well (in the same way `DNSEnumService::lookup_uri_from_user` does).

steelyvoid:
Thanks Alex

Just to get back to you on the ENUM_STARTED log: is there really any need for this?  Presumably its value in DNS-based ENUM is that it tells you how long the ENUM lookups are taking, but file-based lookups are going to be pretty much instantaneous.  Wouldn't this log just add to clutter?

Happy to put it in if you disagree.

AlexHockey:
Yeah I'm convinced on the ENUM_STARTED log.

Issue #1429, opened 2016-05-04
scscf_test.cpp accidentally uses the SP::Message object defined in bono_test.cpp


graemerobertson:
Both scscf_test.cpp and bono_test.cpp define and use an SP::Message class. Unfortunately, scscf_test actually uses bono_test's version of SP::Message. This is presumably because at link time the bono_test version of the object gets picked (effectively randomly, although actually probably actually based on some heuristic like alphabetical order or order in a Makefile or something). It's a little bit surprising that at some point the build doesn't fail because we've got multiple versions of the same object, but perhaps that isn't a thing.

We have a lot of test fixtures that use very similar versions of the Message class, and they're inconsistently defined (some are defined in an SP namespace, some are defined in the test fixture class). Ideally we should almost certain have a single all-purpose common Message class across the UTs, but that is beyond the scope of this issue.

Incidentally, what even is the namespace SP supposed to be?

graemerobertson:
@Tom-Rees is fixing this I believe

graemerobertson:
@Tom-Rees @AlexHockey Let's close this issue since it explicitly states that trying to commonsing the Message class is beyond the scope of this issue (and we're already tracking it elsewhere) - agreed?

Tom-Rees:
Sounds good to me - Alex, feel free to re-open if you disagree.

PR #1440, opened 2016-05-16
[Reviewer: Alex] Rename bono_test Message namespace to be unique

Tom-Rees:
Hi Alex,

This is a temporary fix to https://github.com/Metaswitch/sprout/issues/1429. 

I've renamed the namespace so we no longer get the symbol collision. As it says in the issue, this it a temporary solution. It would be nice to have a common message class, but that's outside the scope of this fix.

Tested that this builds correctly, nothing more.


AlexHockey:
LGTM.

PR #1444, opened 2016-05-18
[Reviewer: Rob] Always make ENT log

### LINKED ISSUE #1226: ENT log not issued when sas_server not set
For test case, ENT-CL_SPROUT_INVALID_SAS_OPTION, the sas_server option is removed from the shared_config file, the file is uploaded, and sprout is restarted.  Sprout no longer issues the ENT log beginning with, "The sas_server option in /etc/clearwater/config is invalid or not configured...".  However a Warning is issued in the sprout log.  Paradoxically Homestead does issue the ENT log when the sas_server option is removed.  The issue was detected during V8 regression testing.  The issue can be reproduced by removing the sas_server line in shared_config, uploading the shared_config file, and then restarting Sprout.

eleanor-merry:
This fixes #1226. I couldn't repro the exact problem, but I've made it so that we always log if the sas server option is missing/invalid.

Tested live

I'll make the same changes to the other SAS'd components as well if this PR is OK.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -667,11 +667,6 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
           TRC_INFO("SAS set to %s", options->sas_server.c_str());
           TRC_INFO("System name is set to %s", options->sas_system_name.c_str());
         }
-        else
-        {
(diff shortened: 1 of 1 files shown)


rkd-msw:
LGTM.

PR #1446, opened 2016-05-19
[Reviewer Alex] Only session_expires with record_route

Tom-Rees:
This is a fix to some slightly confusing behavior where we occasionally add a session expires header without record-routing ourselves into the dialog. Doing so, while not 'wrong', led to some confusing issues in the field.

I've added some UTs for the correct behavior which now pass. No live testing performed yet.

### DIFF
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -424,8 +424,6 @@ void SCSCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
 
   pjsip_status_code status_code = PJSIP_SC_OK;
 
-  _se_helper.process_request(req, get_pool(req), trail());
-
   // Work out if we should be auto-registering the user based on this
   // request and if we are, also work out the IMPI to register them with.
   const pjsip_route_hdr* top_route = route_hdr();
@@ -514,6 +512,8 @@ void SCSCFSproutletTsx::on_rx_in_dialog_request(pjsip_msg* req)
 {
   TRC_INFO("S-CSCF received in-dialog request");
 
+  // We must be record-routed to be in the dialog
+  _record_routed = true;
   _se_helper.process_request(req, get_pool(req), trail());
 
   // Create an ACR for this request and pass the request to it.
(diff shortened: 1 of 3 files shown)


src/ut/scscf_test.cpp
```diff
@@ -8270,3 +8270,130 @@ TEST_F(SCSCFTest, TestSessionExpires)
...
+  list<HeaderMatcher> hdrs;
+  hdrs.push_back(HeaderMatcher("Record-Route", "Record-Route:.*"));
+  hdrs.push_back(HeaderMatcher("Session-Expires", "Session-Expires:.*"));
+
+  list<HeaderMatcher> rsp_hdrs;
+  rsp_hdrs.push_back(HeaderMatcher("Session-Expires"));
+  rsp_hdrs.push_back(HeaderMatcher("Record-Route"));
+}
+
+TEST_F(SCSCFTest, TestSessionExpiresAStoAS)
```
AlexHockey: I'd prefer if this was name `TestSessionExpiresWhenNoRecordRoute`, since the's the record routing that's important here (the AS->AS is just a way of forcing sprout to not rr).

I'll stop being pedantic now.

src/scscfsproutlet.cpp
```diff
@@ -1683,14 +1686,15 @@ bool SCSCFSproutletTsx::lookup_ifcs(std::string public_id, Ifcs& ifcs)
...
-                                         ACR::NodeRole billing_role)
+void SCSCFSproutletTsx::add_to_dialog(pjsip_msg* msg,
+                                      bool billing_rr,
+                                      ACR::NodeRole billing_role)
 {
   pj_pool_t* pool = get_pool(msg);
+  _se_helper.process_request(msg, pool, trail());
 
   pjsip_route_hdr* rr = NULL;
   if (!_record_routed)
```
AlexHockey: The invariant on line 1725 is only true if `add_to_dialog` is only called on initial requests. It it is called on a subsequent request it may crash (since `rr` may end up being null). So you should definitely confirm that the function is only ever called on initial requests. It is also worth seeing if we can make this function more defensive (e.g. having it do nothing if called on an in-dialog request).

AlexHockey:
Looks good. I had one main comment - you've changed one of the invariants related to record-routing and we need to make sure that hasn't broken anything or made the code more fragile.

Tom-Rees:
I've tested that adding an 'add_to_dialog' call in in_dialog requests causes the UT to core dump on the failed assert, so I think we should be as safe as before.

PR #1452, opened 2016-06-03
[Reviewer: Graeme] Always present I-CSCF stats

rkd-msw:
This change means that we always expose I-CSCF transaction stats, even if I-CSCF functionality is disabled. This gives a more consistent experience when polling differently-configured Sprout nodes for stats.

If this looks OK I'll make the same change for S-CSCF and BGCF, but not MMTel (as we never expect that to be enabled in production).

### DIFF
--- a/include/sproutlet.h
+++ b/include/sproutlet.h
@@ -562,11 +562,7 @@ class Sproutlet
 {
 public:
   /// Virtual destructor.
-  virtual ~Sproutlet()
-  {
-    delete _incoming_sip_transactions_tbl;
-    delete _outgoing_sip_transactions_tbl;
-  }
+  virtual ~Sproutlet() {}
 
   SNMP::SuccessFailCountByRequestTypeTable* _incoming_sip_transactions_tbl;
   SNMP::SuccessFailCountByRequestTypeTable* _outgoing_sip_transactions_tbl;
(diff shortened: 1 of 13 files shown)


include/sproutlet.h
```diff
@@ -601,7 +605,11 @@ class Sproutlet
   /// Constructor.
   Sproutlet(const std::string& service_name,
             int port,
-            const std::string& service_host="") :
+            const std::string& service_host="",
+            SNMP::SuccessFailCountByRequestTypeTable* incoming_sip_transactions_tbl = NULL,
+            SNMP::SuccessFailCountByRequestTypeTable* outgoing_sip_transactions_tbl = NULL):
```
graemerobertson: Should `):` be `) :`?

src/icscfplugin.cpp
```diff
@@ -81,6 +81,17 @@ bool ICSCFPlugin::load(struct options& opt, std::list<Sproutlet*>& sproutlets)
 {
   bool plugin_loaded = true;
 
+  // Create the SNMP tables here - they should exist based on whether the
+  // plugin is loaded, not whether the Sproutlet is enabled, in order to
+  // simplify SNMP polling of multiple differently-configured Sprout nodes.
+  //
+  // We'll leak these objects if the Sproutlet isn't enabled - this is a small,
+  // fixed-size leak so we don't need to worry about it.
```
graemerobertson: When do these objects currently get deleted? Can't we delete them in the unload function?

graemerobertson:
Looks legit, excited for the test plan :tulip:

src/icscfplugin.cpp
rkd-msw: Good idea - I'll do that.

rkd-msw:
I'll test by:

* creating a Sprout node
* making a call and checking that the I- and S-CSCF stats are nonzero
* disabling I-CSCF function
* making a call and checking that the S-CSCF stats are nonzero, and the I-CSCF stats are present but zero

src/icscfplugin.cpp
graemerobertson: Cool, remember to delete this comment.

rkd-msw:
[sto1905]

Issue #1441, opened 2016-05-16
WebSocket ConnectionTracker::connection_active fails CHECK_PJ_TRANSPORT_THREAD


mirw:
#### Symptoms

Connect over WebSockets.  Bono crashes with the following stack.

```
Thread 5 (Thread 0x7f8fcffff700 (LWP 7062)):
#0  0x00007f8ff1ba5b99 in __libc_waitpid (pid=8544, stat_loc=stat_loc@entry=0x7f8fcfff8a60, options=options@entry=0) at ../sysdeps/unix/sysv/linux/waitpid.c:40
#1  0x00007f8ff1b2b2e2 in do_system (line=<optimized out>) at ../sysdeps/posix/system.c:148
#2  0x00000000004e2e93 in Logger::backtrace (this=0xcd9120, data=<optimized out>) at ../modules/cpp-common/src/logger.cpp:318
#3  0x000000000059b40d in Log::backtrace (fmt=fmt@entry=0x73e0d8 "Signal %d caught") at ../modules/cpp-common/src/log.cpp:185
#4  0x00000000005f5afc in signal_handler (sig=11) at main.cpp:1131
#5  <signal handler called>
#6  0x00007f8ff1b30a03 in _IO_vfprintf_internal (s=s@entry=0x7f8fcfffba90, format=<optimized out>, format@entry=0x731d58 "Function expected to be called on PJSIP transport thread (%s) - has been called on different thread (%s)", ap=ap@entry=0x7f8fcfffdcd8) at vfprintf.c:1661
#7  0x00007f8ff1bee235 in ___vsnprintf_chk (s=s@entry=0x7f8fcfffbcb2 "Function expected to be called on PJSIP transport thread (websockets) - has been called on different thread ()\n", maxlen=<optimized out>, maxlen@entry=8156, flags=flags@entry=1, slen=slen@entry=18446744073709551615, format=format@entry=0x731d58 "Function expected to be called on PJSIP transport thread (%s) - has been called on different thread (%s)", args=args@entry=0x7f8fcfffdcd8) at vsnprintf_chk.c:63
[...]

PR #1454, opened 2016-06-06
[Reviewer: Rob] Fix log-spam from CHECK_PJ_TRANSPORT_THREAD macro in Bono

johadalin:
To avoid spam logs in Bono, a check is made in connections_active to see if it is being called on a websockets thread before running the check to see if it is the pjsip transport thread or not. 

To avoid races, the lock has been replaced in this function only. The macro in place in other functions should alert us if we need to replace the lock in any other locations. 

As the websockets thread should behave like a pjsip transport thread, deadlock should not be possible.

Tested on a live deployment, running live tests and connecting/disconnecting a webRTC client. No spurious log outputs were seen, and Sprout/Bono ran without crashes.

### DIFF
--- a/include/connection_tracker.h
+++ b/include/connection_tracker.h
@@ -84,7 +84,7 @@ public:
   void unquiesce();
 
 private:
-  // This must be held when accessing connection_active, to avoid contention
+  // This must be held when accessing _connection_listeners, to avoid contention
   // between the transport thread and websocket threads.
   pthread_mutex_t _lock;
 
--- a/include/connection_tracker.h
+++ b/include/connection_tracker.h
@@ -84,6 +84,10 @@ public:
   void unquiesce();
 
 private:
+  // This must be held when accessing connection_active, to avoid contention
+  // between the transport thread and websocket threads.
+  pthread_mutex_t _lock;
+
   // A map of all the connections known to the connection manager, and their
   // state listeners.  This is a set of pjsip transports, but only includes
   // connection-based transports (not datagram transports).
--- a/src/connection_tracker.cpp
+++ b/src/connection_tracker.cpp
@@ -49,6 +49,11 @@ ConnectionTracker::ConnectionTracker(
   _quiescing(PJ_FALSE),
   _on_quiesced_handler(on_quiesced_handler)
 {
+  pthread_mutexattr_t attrs;
+  pthread_mutexattr_init(&attrs);
+  pthread_mutexattr_settype(&attrs, PTHREAD_MUTEX_RECURSIVE);
+  pthread_mutex_init(&_lock, &attrs);
+  pthread_mutexattr_destroy(&attrs);
 }
 
 
@@ -64,6 +69,7 @@ ConnectionTracker::~ConnectionTracker()
                                           it->second,
                                           (void *)this);
   }
+  pthread_mutex_destroy(&_lock);
 }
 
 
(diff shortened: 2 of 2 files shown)


src/connection_tracker.cpp
```diff
@@ -49,6 +49,11 @@ ConnectionTracker::ConnectionTracker(
   _quiescing(PJ_FALSE),
   _on_quiesced_handler(on_quiesced_handler)
 {
+  pthread_mutexattr_t attrs;
+  pthread_mutexattr_init(&attrs);
+  pthread_mutexattr_settype(&attrs, PTHREAD_MUTEX_RECURSIVE);
```
rkd-msw: Why recursive?

src/connection_tracker.cpp
```diff
@@ -49,6 +49,11 @@ ConnectionTracker::ConnectionTracker(
   _quiescing(PJ_FALSE),
   _on_quiesced_handler(on_quiesced_handler)
 {
+  pthread_mutexattr_t attrs;
```
rkd-msw: Simpler to just use `PTHREAD_RECURSIVE_MUTEX_INITIALIZER`?

src/connection_tracker.cpp
```diff
@@ -111,10 +117,17 @@ void ConnectionTracker::connection_active(pjsip_transport *tp)
...
-    // this is not the case.
-    CHECK_PJ_TRANSPORT_THREAD();
+    pthread_mutex_lock(&_lock);
+
+    // We expect to be called by only websocket transport threads, or the PJSIP
+    // transport thread. First check if we are on a websocket thread, then run
+    // the standard CHECK_PJ_TRANSPORT_THREAD macro if we aren't.
+    // Race/locking safety is based on this assumption. Raise an error log if
+    // the above is not the case.
+    if ((strcmp(pj_thread_get_name(pj_thread_this()),"websockets")) != 0)
```
rkd-msw: Should have a space after the comma.

rkd-msw:
Looks good - couple of comments about simplifying mutex initialization.

src/connection_tracker.cpp
```diff
@@ -111,10 +117,17 @@ void ConnectionTracker::connection_active(pjsip_transport *tp)
...
   if ((tp->flag & PJSIP_TRANSPORT_DATAGRAM) == 0)
   {
-    // We expect to only be called on the PJSIP transport thread, and our data
-    // race/locking safety is based on this assumption. Raise an error log if
-    // this is not the case.
-    CHECK_PJ_TRANSPORT_THREAD();
+    pthread_mutex_lock(&_lock);
+
+    // We expect to be called by only websocket transport threads, or the PJSIP
+    // transport thread. First check if we are on a websocket thread, then run
```
rkd-msw: Maybe worth explicitly saying "not the PJSiP worker threads".

rkd-msw: I think the sentence starting "First check if..." doesn't say anything the code doesn't say. If you're worried that the strcmp isn't clear enough, how about:

```
bool is_websockets_thread = strcmp(pj_thread_get_name(pj_thread_this()),"websockets") == 0;
if (!is_websockets_thread)
...
```

include/connection_tracker.h
```diff
@@ -84,6 +84,10 @@ class ConnectionTracker
   void unquiesce();
 
 private:
+  // This must be held when accessing connection_active, to avoid contention
```
rkd-msw: `connection_active` is a function, whereas we usually talk about locks in terms of controlling access to data. What's this really protecting? (If it's `_connection_listeners`, for example, we should be using the lock more widely, so that the websockets thread isn't modifying that map while the transport thread is trying to read it and getting corrupt data).

rkd-msw:
...and a couple more about what the lock is really for.

rkd-msw:
:+1:

johadalin:
"Fixes" #1441

PR #1456, opened 2016-06-13
[Reviewer: Rob] Use the new AlarmManager

eleanor-merry:
Example of using the new AlarmManager in https://github.com/Metaswitch/cpp-common/pull/526


rkd-msw:
Looks good.

PR #1457, opened 2016-06-14
Switch from Python version of bracket_ipv6_address to C++ version

mirw:
Same as https://github.com/Metaswitch/clearwater-infrastructure/pull/332, but for this repo.


rkd-msw:
:+1:

PR #1458, opened 2016-06-15
[Reviewer: rob] Extend the AsCommunicationTracker to include the failure reason in ENT logs.

AlexHockey:
At present we have an `AsCommunicationTracker` class that is notified of successful and failed AS communication attempts and raises alarms and ENT logs appropriately. This PR extends this class to allow the user to pass a textual description of the cause of the failure so it can be output in ENT logs, and changes the S-CSCF to supply this description. 

Tested in UT.


rkd-msw:
Looks good. As discussed, though, we should UT that the log only contains the first failure reason.

PR #1459, opened 2016-06-16
[Reviewer AJH] Only set the SAS trail if UAS tsx create works

steelyvoid:
The set_trail call in authenticate_rx_request can only be made if the call to pjsip_tsx_create_uas2 succeeds, as the tsx pointer is still null otherwise

### DIFF
--- a/src/authentication.cpp
+++ b/src/authentication.cpp
@@ -1181,6 +1181,7 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
   status = pjsip_tsx_create_uas2(NULL, rdata, NULL, &tsx);
   if (status != PJ_SUCCESS)
   {
+    set_trail(tdata, trail);
     // LCOV_EXCL_START - defensive code not hit in UT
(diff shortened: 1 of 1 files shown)


AlexHockey:
If the call fails I think we should be calling `set_trail` on the `tdata` instead of the `tsx` (the function is overloaded to accept different types of argument). Otherwise the response will not appear in SAS.

steelyvoid:
Like so?  I've verified that the failing UT that was causing me to hit a SEGV here no longer crashes

AlexHockey:
Looks good. As discussed, we used to have a UT that exposed this bug, but @eleanor-merry fixed the UT to not expose it. I wonder if we should undo the UT fix (or add a new UT that does not have the fix) to prevent your new code from being regressed. Thoughts?

steelyvoid:
The UT that accidentally hit the bad code path is AuthenticationTest.IntegrityProtected.  I assume that's testing something important so I wouldn't want to undo the fix to it.  It does make sense for a UT to be written to exercise the error path

PR #1464, opened 2016-06-30
[Reviewer: Graeme] Map HSS responses to SIP failure codes correctly

### LINKED ISSUE #1453: Sprout returns 504 SERVER_TIMEOUT error when service profile doesn't contain a public identity
[This branch of code](https://github.com/Metaswitch/sprout/blob/092299363e18bb0be15b18c25ecacba1001ea64d/src/registrar.cpp#L630) is very confused. We enter it if update_registration_state doesn't return `HTTP_OK`. This can happen either because we've failed to contact Homestead/the HSS or because we've received some invalid IFCs (and possibly other reasons). However, we set the terribly named `st_code` variable to `PJSIP_SC_SERVER_TIMEOUT` regardless, and then we drop a completely contradictory log - `"rejecting register request with invalid public/private identity"`.

The impact of this is that a REGISTER request can fail due to a `504 SERVER TIMEOUT` when in fact no servers have timed out, we've just got some bad state in Homestead/the HSS. This is pretty confusing. Similarly, I assume that if a request to Homestead/the HSS genuinely times out then we'll still drop the log above, which would also be confusing.

Seen whilst running the live tests on an AIO node with Sprout at current dev, i.e. `092299363e18bb0be15b18c25ecacba1001ea64d`.

richardwhiuk:
Treat a server failure (either caused by a failure by the HSS or by a failure
to decode the response) as a SIP 500 Internal Server Error.

Log an appropriate message instead of assuming all failures are caused by
incorrect registration details, including the return code and HSS response
where appropriate.

This fixes #1453 

I've checked that all the UTs pass, and that the two new UTs fail without this code.

### DIFF
--- a/include/hss_sip_mapping.h
+++ b/include/hss_sip_mapping.h
@@ -1,0 +1,55 @@
+/**
+ * @file hss_sip_mapping.h Map HSS responses to SIP responses
+ *
+ * Project Clearwater - IMS in the Cloud
+ * Copyright (C) 2016  Metaswitch Networks Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version, along with the "Special Exception" for use of
+ * the program along with SSL, set forth below. This program is distributed
+ * in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details. You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ *
+ * The author can be reached by email at clearwater@metaswitch.com or by
+ * post at Metaswitch Networks Ltd, 100 Church St, Enfield EN2 6BQ, UK
+ *
+ * Special Exception
+ * Metaswitch Networks Ltd  grants you permission to copy, modify,
+ * propagate, and distribute a work formed by combining OpenSSL with The
+ * Software, or a work derivative of such a combination, even if such
+ * copying, modification, propagation, or distribution would otherwise
+ * violate the terms of the GPL. You must comply with the GPL in all
+ * respects for all of the code used other than OpenSSL.
+ * "OpenSSL" means OpenSSL toolkit software distributed by the OpenSSL
+ * Project and licensed under the OpenSSL Licenses, or a work based on such
+ * software and licensed under the OpenSSL Licenses.
+ * "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+ * under which the OpenSSL Project distributes the OpenSSL toolkit software,
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+
+
+#ifndef HSS_SIP_MAPPING_H__
+#define HSS_SIP_MAPPING_H__
+
+extern "C" {
+#include <pjsip.h>
+}
+
+#include "acr.h"
+
+bool process_hss_sip_failure(HTTPCode http_code,
+                             std::string& reg_state,
+                             pjsip_rx_data* rdata,
+                             struct stack_data_struct& stack_data,
+                             ACR* acr,
+                             const char* type);
+
+#endif
+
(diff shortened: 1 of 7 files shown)


include/hss_sip_mapping.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef HSS_SIP_MAPPING_H__
+#define HSS_SIP_MAPPING_H__
+
+extern "C" {
+#include <pjsip.h>
+}
+
+#include "acr.h"
+
+bool process_hss_sip_failure(HTTPCode http_code,
```
graemerobertson: Can you write a comment for this function, either here or in the CPP file? In particular, calling PJUtils::respond_stateless is a fairly major action so it would be good if this was clear from looking at the function definition.

(We're generally fairly inconsistent about a) writing function comments and b) where we write function comments, particularly in Sprout, which isn't great)

src/hss_sip_mapping.cpp
```diff
@@ -0,0 +1,112 @@
...
+  int st_code = PJSIP_SC_OK;
+
+  if ((http_code != HTTP_OK) || (regstate != HSSConnection::STATE_REGISTERED))
+  {
+    // We failed to register this subscriber at the HSS. This may be because
+    // the HSS is unavilable, the public identity doesn't exist, the public
+    // identity doesn't belong to the private identity, or there was an error
+    // communicating with the HSS.
+
+    if (http_code == HTTP_OK)
```
graemerobertson: Should this be a switch statement?

src/hss_sip_mapping.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+  int st_code = PJSIP_SC_OK;
+
+  if ((http_code != HTTP_OK) || (regstate != HSSConnection::STATE_REGISTERED))
+  {
+    // We failed to register this subscriber at the HSS. This may be because
+    // the HSS is unavilable, the public identity doesn't exist, the public
+    // identity doesn't belong to the private identity, or there was an error
+    // communicating with the HSS.
+
```
graemerobertson: I don't think we'd normally put a blank line here. Also both new files have blank lines at the end of them which is my absolute least favourite thing, the one thing I stand against in fact.

Obviously this isn't exactly 'must address' feedback.

src/hss_sip_mapping.cpp
```diff
@@ -0,0 +1,112 @@
...
+    if (http_code == HTTP_OK)
+    {
+      TRC_ERROR("Rejecting %s request following failure to register on the HSS: %s",
+                type, regstate.c_str());
```
graemerobertson: Can we rename `type` to `sip_msg_type` or something that makes it clear what it is? I'd probably actually call it `sip_msg_type_log_str` but that's fairly long-winded.

src/hss_sip_mapping.cpp
```diff
@@ -0,0 +1,112 @@
...
+      // response
+      TRC_ERROR("Rejecting %s request following error communicating with the HSS",
+                type);
+
+      st_code = PJSIP_SC_INTERNAL_SERVER_ERROR;
+    }
+    else
+    {
+      TRC_ERROR("Rejecting %s request following response %d from HSS",
+                type, http_code);
```
graemerobertson: Shouldn't we set `st_code` to something other than PJSIP_SC_OK here?

graemerobertson:
I've reviewed, looks basically all fine.

Two further bits of feedback...
* This presumably breaks Sprout coverage, and you presumably already know that because I saw you asking about how to run the UT coverage. I don't much care whether you fix this by adding some more boring/identical UTs or just LCOV_EXCL offending branches (assuming they're essentially no-ops).
* Your github username is absurd.

include/hss_sip_mapping.h
richardwhiuk: Yes, done.

src/hss_sip_mapping.cpp
richardwhiuk: Done.

src/hss_sip_mapping.cpp
richardwhiuk: Fixed

src/hss_sip_mapping.cpp
richardwhiuk: I've renamed it, and clarified it's usage in the function comment.

src/hss_sip_mapping.cpp
richardwhiuk: Yes, I was dithering what the correct answer was. I've defaulted to just returning the same as before on the assumption it represents some unknown error that will disappear if the subscriber retries later.

Issue #965, opened 2015-03-06
Sprout's stdout goes to /dev/null


mirw:
Issue #963 covers a crash in which glibc detects memory corruption and dumps details to stdout.  Unfortunately, sprout's stdout is redirected to /dev/null, so there's no way to see these details.

We should redirect this output elsewhere and gather it as part of diagnostics dumps.

richardwhiuk:
Sprout's error and info logs are redirected by Utils::daemonize in cpp-common.

So we need to provide a path for stderr and stdout in src/main.cpp, and then add parameters to cpp-common to use them.

mirw:
Just assigned the pull requests back to you.

PR #1466, opened 2016-06-30
[Reviewer: Matt W] Log stdout and stderr to file when a daemon

richardwhiuk:
Matt, this is the sprout half of the fix to #965.

We use the new function in https://github.com/Metaswitch/cpp-common/pull/531 to add a new logfile for standard out and standard error, and add a logrotate configuration for both files.

### DIFF
--- a/debian/sprout-base.logrotate
+++ b/debian/sprout-base.logrotate
@@ -1,0 +1,12 @@
+/var/log/sprout/sprout_out.log /var/log/sprout/sprout_err.log {
+        rotate 4
+        weekly
+        minsize 1M
+        missingok
+        create 640 root adm
+        notifempty
+        compress
+        delaycompress
+        copytruncate
+}
+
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -1349,7 +1352,8 @@ int main(int argc, char* argv[])
 
   if (opt.daemon)
   {
-    int errnum = Utils::daemonize();
+    int errnum = Utils::daemonize(opt.log_directory + "/" + STANDARD_OUT_FILE,
+                                  opt.log_directory + "/" + STANDARD_ERR_FILE);
```
mirw: What about bono?  This includes the same main.cpp.  A few lines below (see `prog_name`), we parse `argv[0]` to find which executable we're running, and name other log files using this.  Maybe we should do the same here?

debian/sprout-base.logrotate
```diff
@@ -0,0 +1,12 @@
+/var/log/sprout/sprout_out.log /var/log/sprout/sprout_err.log {
```
mirw: Probably need a similar entry for bono (in debian/bono.logrotate).

mirw:
Looks good.  Comments are just that we use the same binary for bono, and we should cater for that too.

Issue #1462, opened 2016-06-28
Allow configuration of FQDN in Route header used when communicating with ASs


mirw:
#### Symptoms

As discussed in http://lists.projectclearwater.org/pipermail/clearwater_lists.projectclearwater.org/2016-June/003003.html, Clearwater must connect directly to any ASs, not be behind a NAT.  Specifically, the issue is that Clearwater provides its local IP address in its Route header, which the AS can't connect to.

#### Impact

It's not possible to deploy Clearwater behind a NAT relative to the AS.

#### Release and environment 

release-100

#### Steps to reproduce

Deploy Clearwater behind a NAT with respect to the AS.

PR #1469, opened 2016-06-30
[Reviewer: Matt W] Allow configuration of S-CSCF Node URI

richardwhiuk:
Matt,

#1462 was raised to cover allowing the host used in the Route header when communicating with ASs to be configurable. The rationale here was that Sprout might be behind a NAT relative to the AS, and thus it needs a public address on which to communicate with Sprout.

Following discussions with @bossmc, we believe that the node's public hostname already performs this role, and we should just be advertising the SCSF as on that URL. This should then work for all situations currently (as the public_host will resolve to the local_host when there is no NAT) and will work for the suggested configuration, as the public_host address will be, according to the AS, the no NATed address. This will cause additional DNS resolutions to be performed, but these will be cached.

I've also taken the liberty of removing the -c / --scsf-uri documented option as this is no longer implemented - it was removed in https://github.com/Metaswitch/sprout/pull/1364


richardwhiuk:
Matt,

Following our discussion regarding split horizon DNS, I've changed to adding a new --scsf-node-uri option (piped through from /etc/clearwater/config via init), which will allow the S-CSCF node URI to be configured on a per node basis by placing scscf_node_uri = sip://domain:port in /etc/clearwater/local_config.

If unspecified, or specified as blank., this will default to the previous behaviour

mirw:
Looks good.  No comments.

PR #1470, opened 2016-07-01
[Reviewer: EM] Disable ability to configure chronos

### LINKED ISSUE #1463: Disable the ability to configure the chronos connection
#### Symptoms
I can disable Chronos by setting chronos_hostname to "" in the config. This means that the chronos_connection object is NULL. However, the authentication module and the subscriber data manager use the ChronosConnection without checking it's NULL.

richardwhiuk:
This configuration option was undocumented, and isn't supported, so we just hardcode localhost:7253

This fixes #1463

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1854,7 +1854,7 @@ int main(int argc, char* argv[])
       chronos_callback_host = "[::1]:" + port_str;
     }
 
-    std::string chronos_service = "localhost:7253";
+    std::string chronos_service = "127.0.0.1:7253";
     TRC_STATUS("Creating connection to Chronos %s using %s as the callback URI",
                chronos_service.c_str(),
                chronos_callback_host.c_str());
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -1873,6 +1843,26 @@ int main(int argc, char* argv[])
...
+    std::string chronos_callback_host = "127.0.0.1:" + port_str;
+
+    // We want Chronos to call back to its local sprout instance so that we can
+    // handle Sprouts failing without missing timers.
+    if (is_ipv6(opt.http_address))
+    {
+      chronos_callback_host = "[::1]:" + port_str;
+    }
+
+    std::string chronos_service = "localhost:7253";
```
eleanor-merry: Actually can this be 127.0.0.1? The default value was localhost, but we did override this in the init.d file to 127.0.0.1

eleanor-merry:
LGTM

PR #1471, opened 2016-07-04
[Reviewer: MIRW] Commonalize logging and daemonization

richardwhiuk:
Use the common logging and daemonization function in Utils.

Only slightly controversial change is to move init_pjsip_logging down below creating the other loggers, but this looks fine.


mirw:
Looks good.  I agree that `init_pjsip_logging` should be fine there.

PR #1474, opened 2016-07-05
[Reviewer: MIRW] Incremental build with installed submodules

richardwhiuk:
'Install' submodules to a separate directory and then rsync the changed ones
into place only caring about differing checksums.

This means that all of the sprout source files will build incrementally correctly when built using the top level Makefile.

The remaining two long operations are libmemcached which always rebuilds, and cassandra, which is always rebuilt.


Makefile
```diff
@@ -30,16 +30,36 @@ DEB_NAMES += clearwater-sip-stress clearwater-sip-stress-stats clearwater-sip-pe
...
+#
+# To work around this, we install them elsewhere, and then synchronize them to allow
+# incremental builds to work
+PRE_ROOT := ${ROOT}/build/module-install
+PRE_PREFIX := ${PRE_ROOT}/usr
+PRE_INSTALL_DIR := ${PRE_PREFIX}
+PRE_INCLUDE_DIR := ${PRE_INSTALL_DIR}/include
+PRE_LIB_DIR := ${PRE_INSTALL_DIR}/lib
+
+sync_install:
```
bossmc: .PHONY?

mirw:
Looks plausible.  Presumably you've tested it?

High-level question is whethere we're right to put this all in Makefile, or should we be putting this in the `.mk` files - the latter feels cleaner, if only because it means that we can move these over to other projects more easily.

richardwhiuk:
Yes - I've tested it.

I've found a fix that works to make the cassandra build incremental, and on your suggestion, I've moved the work to make the incrementalism of the sprout source files based on submodule changes into clearwater-build-infrastructure.

I've also fixed the fact that sprout shouldn't link to cassandra/thrift, only sprout tests.

With both these fixes we now can run a full sprout `make` in ~25 second and sprout `make test` in ~30 seconds, with no changed files.

mirw:
Great stuff on reducing the build time.

As just discussed, can we do this in a less-disruptive way by building everything to the right place, but having a build target to save off all the files beforehand and restoring them again on success?  (Need to watch our for submodule build failures when doing this.)

In particular, I'm currently worried that it's possible to do `make sas-client` followed by `make sprout` in the src subdirectory, and not pick up the new version of the SAS client.

richardwhiuk:
Even better, we can save off the new files afterwards, and then 'correct' the timestamps of the old files.

This makes make sas-client work correctly, and turns make sync_install into a magic 'make it faster' command.

PR #1478, opened 2016-07-12
Bgcf 8

ChrisRees:
Changes to Sprout to allow the new BGCF to route based off DN's using exact matching.
Added a method to PJSip which allows us to check if a DN is present in a URI and extract it.
Refactored the URIClassifier::classify_uri() to pull the logic for checking for a DN out into a separate method. The existing logic for classify_uri is unaffected.

UT's passing locally, including the new UT for our PJUtils method.


ChrisRees:
Matt, I think these are supposed to come through you, please can we have feedback latest COB 13/07/16 as we need to get on with ST this week.

mirw:
@AlexHockey, please can you review?

AlexHockey:
Hi Chris, I've got a few high level comments on the new functions you've added. 

* `get_dn`: I assumed from its name that if it returned true then it would output a DN. But if I call it with `sip:bob@example.com` it returns true with `routing_value` set to `"bob"`, which isn't a DN. This function isn't doing much more than `user_from_uri` so maybe it would be better to remove the length check and just call it `user_as_string_from_uri`?

* `contains_dn_as_user`: I really didn't have any idea what this did from its signature. I looked at the code and I'm still confused. It looks like it tries to classify a URI assuming it contains a DN (and returns `UNKNOWN` if not). But it looks like calling it with `sip:definitely-not-a-phone-number@home.domain.com` may cause it to classify the URI as a global or local number, which definitely feels wrong. Have I missed something?

I think you need to decide what you actually want these functions to do, then rename and/or add comments to their header files explaining what they are trying to do. Once you've tightened this up I'll do a more detailed review.

ChrisRees:
I'm happy enough to rename the get_dn method.
It's assumed that when we're calling it we've done all the relevant checks to be happy it's a DN.

For contains_dn_as_user, it's just taking the existing logic which for SIP URI's checks for the "user=phone" parameter then classifies it as either a LOCAL or GLOBAL phone number. It might be better to rename it to something like: classify_number, especially as it returns a URIClass.

ChrisRees:
I'll put together the rest of the review job then send it back, missed the last statement.

ChrisRees:
I've updated the various bits and pieces as suggested. Let me know if it's still unclear.

src/pjutils.cpp
```diff
@@ -2171,6 +2171,14 @@ bool PJUtils::get_rn(pjsip_uri* uri, std::string& routing_value)
   return rn_set;
 }
 
+bool PJUtils::user_as_string_from_uri(pjsip_uri* uri,
+                                      std::string& routing_value)
+{
+  pj_str_t user = PJUtils::user_from_uri(uri);
+  routing_value = pj_str_to_string(&user);
+  return (routing_value.size() > 0);
```
AlexHockey: I think I'd rather you just return the string here, for symmetry with `user_from_uri`.

src/uri_classifier.cpp
```diff
@@ -138,7 +139,48 @@ URIClass URIClassifier::classify_uri(const pjsip_uri* uri, bool prefer_sip)
     }
   }
   // No number portability data
-  else if (PJSIP_URI_SCHEME_IS_TEL(uri))
+  else if (user_type != URIClass::UNKNOWN)
+  {
+    // we have a phone number
+    ret = user_type;
+  }
+  else if(PJSIP_URI_SCHEME_IS_SIP(uri))
```
AlexHockey: Need a space after the `if`.

src/uri_classifier.cpp
```diff
@@ -112,6 +112,7 @@ URIClass URIClassifier::classify_uri(const pjsip_uri* uri, bool prefer_sip)
   // First, check to see if this URI has number portability data - this takes priority
   bool has_rn = false;
   bool has_npdi = false;
+  URIClass user_type = URIClassifier::classify_user_type(uri, prefer_sip);
```
AlexHockey: You call this method then sometimes don't use the result. Could you only call it when you need to?

include/sproutsasevent.h
```diff
@@ -64,6 +64,8 @@ namespace SASEvent
   const int BGCF_NO_ROUTE_DOMAIN = SPROUT_BASE + 0x000012;
   const int BGCF_FOUND_ROUTE_NUMBER = SPROUT_BASE + 0x000013;
   const int BGCF_NO_ROUTE_NUMBER = SPROUT_BASE + 0x000014;
+  const int BGCF_FOUND_DIRECTORY_NUMBER = SPROUT_BASE + 0x000015;
+  const int BGCF_NO_DIRECTORY_NUMBER = SPROUT_BASE + 0x000016;
```
AlexHockey: You've added these constants, but I don't see any code that uses them in this PR, and I haven't received a PR for any SAS resource bundle changes. Should they be in this PR?

src/ut/pjutils_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+
+
+ TEST_F(PJUtilsTest, GetDNTest)
+ {
+    std::string uri_str = "sip:2012030005@domain.com";
+    pjsip_uri* uri = PJUtils::uri_from_string(uri_str, pool);
+    std::string routing_value;
+
+    EXPECT_EQ(true, PJUtils::get_dn(uri, routing_value));
```
AlexHockey: This function no longer exists. Do these tests still work??

src/ut/pjutils_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "pjutils.h"
+
+ class PJUtilsTest : public BaseTest
+ {
+ public:
+   static pj_caching_pool caching_pool;
+   static pj_pool_t* pool;
+   static pjsip_endpoint* endpt;
+
+  static void SetUpTestCase()
```
AlexHockey: Alignment look wrong.

src/ut/pjutils_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+ * "OpenSSL" means OpenSSL toolkit software distributed by the OpenSSL
+ * Project and licensed under the OpenSSL Licenses, or a work based on such
+ * software and licensed under the OpenSSL Licenses.
+ * "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+ * under which the OpenSSL Project distributes the OpenSSL toolkit software,
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+
+///
+///----------------------------------------------------------------------------
```
AlexHockey: This comment seems slightly random.

include/pjutils.h
```diff
@@ -272,6 +272,7 @@ std::string remove_visual_separators(const pj_str_t& number);
 
 bool get_npdi(pjsip_uri* uri);
 bool get_rn(pjsip_uri* uri, std::string& routing_value);
+bool user_as_string_from_uri(pjsip_uri* uri, std::string& routing_value);
```
AlexHockey: Surely the 2nd parameter should be called `user` not `routing_value`?

src/uri_classifier.cpp
```diff
@@ -138,7 +139,48 @@ URIClass URIClassifier::classify_uri(const pjsip_uri* uri, bool prefer_sip)
...
+      ret = OFFNET_SIP_URI;
+    }
+  }
+
+  TRC_DEBUG("Classified URI as %d", (int)ret);
+  return ret;
+}
+
+URIClass URIClassifier::classify_user_type(const pjsip_uri* uri,
+                                           bool prefer_sip)
```
AlexHockey: It would be useful to have a comment (either here or in the header explaining what `prefer_sip` does.

include/uri_classifier.h
```diff
@@ -57,6 +57,10 @@ namespace URIClassifier
 {
   URIClass classify_uri(const pjsip_uri* uri, bool prefer_sip = true);
 
+  // Classifies the user type, either returns LOCAL_PHONE_NUMBER,
+  // GLOBAL_PHONE_NUMBER or UNKNOWN.
+  URIClass classify_user_type(const pjsip_uri* uri, bool prefer_sip = true);
```
AlexHockey: This method name is still not very clear. I think what it does it classify the URI if it contains a DN (albeit with an existing bug), otherwise if returns unknown. So how about `classify_uri_if_dn`?

src/uri_classifier.cpp
```diff
@@ -138,7 +139,48 @@ URIClass URIClassifier::classify_uri(const pjsip_uri* uri, bool prefer_sip)
...
+    pjsip_sip_uri* sip_uri = (pjsip_sip_uri*)uri;
+    pj_str_t host = sip_uri->host;
+    bool home_domain = is_home_domain(host);
+    bool local_to_node = is_local_name(host);
+
+    TRC_DEBUG("home domain: %s, local_to_node: %s, enforce_user_phone: %s, prefer_sip: %s",
+              home_domain ? "true" : "false",
+              local_to_node ? "true" : "false",
+              enforce_user_phone ? "true" : "false",
+              prefer_sip ? "true" : "false");
```
AlexHockey: This function doesn't care about all of these parameters so you shouldn't log all of them.

AlexHockey:
I've reviewed - comments inline. I'll admit I don't understand why these changes were needed but I've sent you a separate email about that. I'll hold onto the PR until we've resolved that.

ChrisRees:
As discussed we now don't need these changes in sprout, closing the pull request and removing the branch.

PR #1480, opened 2016-07-12
[Reviewer: Rob] Fix up rr uris

### LINKED ISSUE #1479: There's always a port in the sproutlet record-route headers
We build up the RR headers in the sproutletproxy - https://github.com/Metaswitch/sprout/blob/dev/src/sproutletproxy.cpp#L93. 

However, we always include a port - this is wrong as it breaks our SRV support (as anything that routes to Sprout using those headers should route using the port and therefore not use any SRV records)

eleanor-merry:
This PR plumbs through the actual sproutlet URIs to the sproutletproxy (rather than constructing them in the sproutletproxy).

Tested by checking the RR headers change if I change the sproutlet configuration

Fixes #1479


include/sproutlet.h
```diff
@@ -584,6 +584,9 @@ class Sproutlet
   /// Returns the name of this service.
   const std::string service_name() const { return _service_name; }
 
+  /// Returns the uri of this service (as a string)
```
rkd-msw: Can you uppercase it as URI, for consistency with the comment at line 627?

src/sproutletproxy.cpp
```diff
@@ -88,16 +88,12 @@ SproutletProxy::SproutletProxy(pjsip_endpoint* endpt,
...
-                             std::to_string((*it)->port()) +
-                             ";transport=tcp";
-    TRC_DEBUG("Record-Route URI = %s", rr_uri_str.c_str());
     pjsip_sip_uri* root_uri = (pjsip_sip_uri*)PJUtils::uri_from_string(
-                                                       rr_uri_str,
+                                                       (*it)->uri_as_str(),
                                                        stack_data.pool,
                                                        false);
-    _root_uris.insert(std::pair<std::string, pjsip_sip_uri*>((*it)->service_name(), root_uri));
+    _root_uris.insert(std::pair<std::string, pjsip_sip_uri*>(
```
rkd-msw: Feels like an odd pattern of line breaks, but I'm not too fussed. (If line length is a concern, is `_root_uris.insert(std::make_pair((*it)->service_name(), root_uri));` the way to go?)

rkd-msw:
Cosmetic comments only. @sathiyan-sivathas, you'll want to take this fix, probably by upgrading the VMware Sprout tomorrow.

rkd-msw:
Actually, @eleanor-merry, given that I only had cosmetic comments, I'm going to merge this so that it's in the `latest` repo first thing tomorrow morning.

PR #1483, opened 2016-07-14
[Reviewer: Rob] Debuggable plugin disabled

chris-elford-metaswitch:
More info in the commit messages.

### DIFF
--- a/src/bgcfplugin.cpp
+++ b/src/bgcfplugin.cpp
@@ -98,7 +98,7 @@ bool BGCFPlugin::load(struct options& opt, std::list<Sproutlet*>& sproutlets)
                                                                                     "1.2.826.0.1.1578918.9.3.23");
   if (opt.enabled_bgcf)
   {
-    TRC_INFO("BGCF plugin enabled.");
+    TRC_STATUS("BGCF plugin enabled");
 
     // Create BGCF service required for the BGCF Sproutlet.
     _bgcf_service = new BgcfService();
--- a/src/bgcfplugin.cpp
+++ b/src/bgcfplugin.cpp
@@ -44,6 +44,7 @@
 #include "sproutletplugin.h"
 #include "bgcfservice.h"
 #include "bgcfsproutlet.h"
+#include "log.h"
 
 class BGCFPlugin : public SproutletPlugin
 {
@@ -97,6 +98,8 @@ bool BGCFPlugin::load(struct options& opt, std::list<Sproutlet*>& sproutlets)
                                                                                     "1.2.826.0.1.1578918.9.3.23");
   if (opt.enabled_bgcf)
   {
+    TRC_INFO("BGCF plugin enabled.");
+
     // Create BGCF service required for the BGCF Sproutlet.
     _bgcf_service = new BgcfService();
 
--- a/src/cdivasplugin.cpp
+++ b/src/cdivasplugin.cpp
@@ -43,6 +43,7 @@
 #include "sproutletplugin.h"
 #include "sproutletappserver.h"
 #include "mmtel.h"
+#include "log.h"
 
 class CDivASPlugin : public SproutletPlugin
 {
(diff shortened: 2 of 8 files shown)


src/bgcfplugin.cpp
```diff
@@ -97,6 +98,8 @@ bool BGCFPlugin::load(struct options& opt, std::list<Sproutlet*>& sproutlets)
                                                                                     "1.2.826.0.1.1578918.9.3.23");
   if (opt.enabled_bgcf)
   {
+    TRC_INFO("BGCF plugin enabled.");
```
rkd-msw: I wouldn't normally expect these log messages to end with punctuation - c.f. https://github.com/Metaswitch/sprout/blob/dev/src/acr.cpp#L216

rkd-msw: Generally, our "I have created XYZ based on a config option" logs at startup are TRC_STATUS - e.g. https://github.com/Metaswitch/sprout/blob/dev/src/main.cpp#L1750. I *think* that only STATUS, WARNING and ERROR logs appear by default, so these logs won't appear unless log_level is explicitly set to a higher value - whereas your commit message says you want it in the mainline case.

rkd-msw:
Comments inline, but at a higher level - shouldn't we be doing this in pluginloader.cpp?

[This line](https://github.com/Metaswitch/sprout/blob/03ea18f6135269ec8090c6602eca86f7a995a267/src/pluginloader.cpp#L93) already says when we're trying to load a Sproutlet, and I think if you made [this line](https://github.com/Metaswitch/sprout/blob/03ea18f6135269ec8090c6602eca86f7a995a267/src/pluginloader.cpp#L120) a TRC_STATUS instead of TRC_DEBUG, you'd be able to see what Sproutlets thought they were enabled and which ones thought they were disabled, and you'd get it consistently instead of having to remember to add a log for each Sproutlet.

rkd-msw:
(Actually, you could probably be even more explicit by having pluginloader.cpp check the length of the plugin_sproutlets list after each call to `load()`, and logging out "Plug-in bgcfsproutlet.so has loaded 0 Sproutlets" or "Plug-in bgcfsproutlet.so has loaded 1 Sproutlet")

chris-elford-metaswitch:
I'd like to do it the way you suggest, but if I was just worried about whether these options were set or not, I could read other logs printed during startup. I'm trying to print whether the Sproutlet plugin thinks it's enabled or not, because it may come to a different conclusion to the plugin loader.

The issue is that the ABI between Sprout and its Sproutlet plugins includes the options struct passed to the Sproutlets as opt. If Sprout and the Sproutlet plugin are compiled with a different version of [sproutlet_options.h](https://github.com/Metaswitch/sprout/blob/dev/include/sproutlet_options.h), then the Sproutlet will look in the wrong place in the struct, and get the wrong idea.

That's exactly what happened to a mailing list user - Memento wasn't being loaded because it was an old version.

Of course, there is a wider issue here. I'm going to talk to Huw Smithson, who is trying to tidy up this interface, and see whether he can resolve it.

benritchie:
note - I wasn't aware Huw was tidying up this interface.  we are writing a new closed source BGCF (actually based on the existing code), but at least for now, we are expecting to leave the existing open source bgcf as is - for open source use.

rkd-msw:
I'm not suggesting that you check whether the options are set (I agree that that is unhelpful) - I'm suggesting that you check whether the plugin has loaded a Sproutlet. Taking the particular case of [the Memento plugin](https://github.com/Metaswitch/sprout/blob/498d97f484195eb25189c429585711f3eb8b9bc5/src/mementoasplugin.cpp#L89):

* if the pluginloader reports that it has loaded one Sproutlet, you know that it thinks it's enabled and the TTL/call list length options are correct
* if the pluginloader reports that it has loaded zero Sproutlets, and you have the "Can't have an unlimited maximum call length and a unlimited TTL for the call list store - disabling Memento" log, you know why that's happened
* if the pluginloader reports that it has loaded zero Sproutlets, and you don't have the "Can't have an unlimited maximum call length and a unlimited TTL for the call list store - disabling Memento" log, you know that it thinks `opt.enabled_memento` is false

I think logging in pluginloader.cpp is more generically useful because it doesn't require every Sproutlet to have this log, and it also catches cases where the plugin is enabled by config, but something else goes wrong and causes the Sproutlet to not be loaded (what if we didn't have that TRC_ERROR in the Memento case, for example?). I don't object to having "XYZ plugin enabled" logs in our Sproutlets as well for clarity, but I think we should have a solution in the common Sproutlet infrastructure.

chris-elford-metaswitch:
I'm not sure when I'm going to get to adding that generic infrastructure. This was just meant to be a throwaway addition of some logs to allow us to debug a problem in the future, that I could sneak in in a little down time.

rkd-msw:
I've done that under https://github.com/Metaswitch/sprout/pull/1484, so fine for you just to apply my punctuation/log level markups and merge.

chris-elford-metaswitch:
I've raised issue #1485 to cover the fact that, even with these changes, Sprout isn't debuggable enough here.

PR #1487, opened 2016-07-15
[Reviewer: AJH] Allow an external Chronos cluster to be configured

richardwhiuk:
Alex,

This is the only change required to sprout onto of the split-clusters work to allow Sprout to use an external storage node.


src/main.cpp
```diff
@@ -422,6 +424,8 @@ static void usage(void)
        "                            The URI of this S-CSCF used by other servers, including AS, to contact\n"
        "                            this specific node. Defaults to \"sip:<localhost>:<port_scscf>\"."
        "     --pidfile=<filename>   Write pidfile\n"
+       "     --chronos-hostname <hostname>\n"
+       "                            Specify the hostname of a remote Chronos cluster\n"
```
AlexHockey: Say it defaults to localhost?

AlexHockey:
Looks good. Just one comment inline.

richardwhiuk:
@eleanor-merry - I think you've separately implemented this on dev and merged that into split clusters, so this can be closed?

eleanor-merry:
Yup

PR #1491, opened 2016-07-19
[Review AJH] Retire BGCF routing in S-CSCF if not originating or terminating

steelyvoid:
Hi Alex

This is the change to initial request routing we discussed back in April/May, taking out the default initial route to BGCF if you're unable to create an application chain and thus can't do originating or terminating process.  We agreed that there's no justification for this from 24.229, and BGCF should not be used as a "last resort".

Tested by firing an INVITE with garbage P-Served-User and P-Asserted-Identity headers at the S-CSCF port.  SAS PR to follow


AlexHockey:
Looks good. Passing back to co-ordinate checkin with https://github.com/Metaswitch/clearwater-sas-resources/pull/101

PR #1492, opened 2016-07-19
[Reviewer: RKD] Add ENT logs for UDP transport.

sathiyan-sivathas:
Make sure all failure paths have ENT logs for UDP transport creation.  (I added one for TCP as well because it was missing).

### DIFF
--- a/src/stack.cpp
+++ b/src/stack.cpp
@@ -341,7 +341,7 @@ pj_status_t create_tcp_listener_transport(int port, pj_str_t& host, pjsip_tpfact
   {
     CL_SPROUT_SIP_TCP_START_FAIL.log(port,
                                      PJUtils::pj_status_to_string(status).c_str());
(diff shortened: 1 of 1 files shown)


src/stack.cpp
```diff
@@ -336,6 +339,12 @@ pj_status_t create_tcp_listener_transport(int port, pj_str_t& host, pjsip_tpfact
   status = fill_transport_details(port, &addr, host, &published_name);
   if (status != PJ_SUCCESS)
   {
+    CL_SPROUT_SIP_TCP_START_FAIL.log(port,
+                                     PJUtils::pj_status_to_string(status).c_str());
+    TRC_ERROR("Failed to fill in TCP transport for port %d  (%s)",
```
rkd-msw: There are two spaces after the %d - should just be one.

rkd-msw:
One ridiculously trivial comment.

Issue #1404, opened 2016-04-18
Sprout will challenge registrations repeatedly if memcached is down, rather than returning a 5xx error


rkd-msw:
#### Symptoms

A mailing list user provided logs relating to registration failure, and I spotted this:

```
15-04-2016 11:00:01.578 UTC Debug memcachedstore.cpp:410: Read for impi\\2010000006@xflowresearch.com on replica 0 returned SUCCESS
15-04-2016 11:00:01.578 UTC Debug memcachedstore.cpp:453: Read 23976 bytes from table impi key 2010000006@xflowresearch.com, CAS = 2102
15-04-2016 11:00:01.578 UTC Debug impistore.cpp:754: Retrieved IMPI for 2010000006@xflowresearch.com
[...]

PR #1496, opened 2016-07-27
[Reviewer: RKD] Add UT for memcached SET failure

sathiyan-sivathas:
Adds UT for #1404 to get 100% code coverage.  The test:
- Sends in a register.
- Forces the memcached SET to fail.
- Waits for 500 Server Internal Error.

cpp-common pull request to follow.


rkd-msw:
Looks good.

PR #1497, opened 2016-07-29
[Reviewer: RKD] Make TCP uplift configurable

sathiyan-sivathas:
Make TCP uplift configurable via the _disable_tcp_switch_ option in local/shared config.

Tested live that the TRC_STATUS in main.cpp is hit when the new option is set.  Tested in UT that that we can set the PJSIP option at runtime.

PJSIP pull request to follow.


rkd-msw:
Looks good. Can you document disable_tcp_switch in https://github.com/Metaswitch/clearwater-readthedocs/tree/master/docs (both in the overall config options list, and in the UDP-specific bit of the app server guide)?

Issue #1495, opened 2016-07-26
Bono fails to send requests to an (alive) Sprout


rkd-msw:
#### Symptoms

When running the live tests, four of the 87 tests failed with a 408 error on registration.

SAS shows that we sent a REGISTER to Sprout's I-CSCF port, and that was followed by a 'Request failed due to transport error' log 10ms later.

There's no evidence in the logs that Sprout restarted.

#### Impact

I've only seen unregister failures, but there could be call failures.

The impact is worse because of https://github.com/Metaswitch/sprout/issues/1050 (which means we don't retry).

#### Release and environment 

Jigglypuff, release-103, EC2.

#### Steps to reproduce

I can't reliably reproduce it.

rkd-msw:
Bono logs at this time only show:

```
26-07-2016 15:48:14.682 UTC Status load_monitor.cpp:285: Maximum incoming request rate/second unchanged - only handled 20 requests in last 2736ms, minimum threshold for a change is 136.800003
26-07-2016 15:48:19.090 UTC Status load_monitor.cpp:285: Maximum incoming request rate/second unchanged - only handled 20 requests in last 4412ms, minimum threshold for a change is 220.600006
```

However, we did previously get these logs:

```
26-07-2016 15:42:17.209 UTC Error pjsip: tcpc0x7f98bc01 TCP connect() error: Connection refused [code=120111]
26-07-2016 15:42:17.209 UTC Error pjsip: tcpc0x7f98bc04 TCP connect() error: Connection refused [code=120111]
26-07-2016 15:42:17.209 UTC Error pjsip: tcpc0x7f98bc02 TCP connect() error: Connection refused [code=120111]
26-07-2016 15:42:17.209 UTC Error pjsip: tcpc0x7f98bc05 TCP connect() error: Connection refused [code=120111]
26-07-2016 15:42:17.209 UTC Error pjsip: tcpc0x7f98bc05 TCP connect() error: Connection refused [code=120111]
26-07-2016 15:42:18.213 UTC Error pjsip: tcpc0x7f98bc01 TCP connect() error: Connection refused [code=120111]
```

Could the thread still be using a connection from the pool which isn't actually connected?

rkd-msw:
Could we add some better diags at https://github.com/Metaswitch/sprout/blob/dev/src/bono.cpp#L3008?

Issue #1050, opened 2015-05-21
Bono doesn't retry transactions if one target fails


mirw:
Set up very light stress (1 registration per second) against a 1-bono, 2-sprout deployment.  All runs smoothly.

Kill one of the sprouts.  Bono should switch over and use the other (and not fail any transactions) but it rejects one or two with a 408 Request Timeout response.

This appears to be because Bono uses the upstream connection pool, which overrides any retry behavior.

PR #1498, opened 2016-07-29
[Reviewer: Matt] Remove the connection pool - just rely on standard SIP DNS resolution

rkd-msw:
We discussed the other day that the connection pool probably wasn't as useful as when it was originally implemented, because our DNS resolution function allows us to spread load across a cluster rather than routing all traffic to one peer.

I've therefore tried just cutting it out. I've verified that this does mean Bono retries requests when a Sprout is down, fixing #1050 and (maybe) #1495 - see SAS screenshot below.

2016-07-29_18h06_28

As discussed, this probably has slightly worse behaviour when under so much load that you have TCP congestion (as we only have one connection per Sprout, not 50 connections total), but i think that's the right tradeoff - given that Bono is more for demos/getting started than production, it's better to be resilient (and capable of demoing Clearwater's fault-tolerance) than optimising for behaviour under TCP congestion.


src/bono.cpp
```diff
@@ -3268,24 +3252,6 @@ pj_status_t init_stateful_proxy(SubscriberDataManager* reg_sdm,
...
   dialog_tracker = new DialogTracker(flow_table);
 
-  // Create a connection pool to the upstream proxy.
-  if (upstream_proxy_connections > 0)
-  {
-    pjsip_host_port pool_target;
-    pool_target.host = pj_strdup3(stack_data.pool, upstream_proxy_arg.c_str());
-    pool_target.port = upstream_proxy_port;
-    sprout_ip_tbl = SNMP::IPCountTable::create("bono_connected_sprouts",
-                                               ".1.2.826.0.1.1578918.9.2.3.1");
```
mirw: What are we doing about this statistic?  Do we need to maintain it?  (Probably not.)  If not, though, we should tidy up the MIB definition, and any Cacti scripts that use it (as well as anything else).

mirw:
Looks OK - one comment about statistics inline.  However, I'd also like to confirm that this isn't going to break our stress runs - how do we do that?

rkd-msw:
I'll test that we can drive Sprout nodes to 60% CPU both with and without this change - is there any kind of breakage you can think of which that won't catch?

mirw:
I guess overload is probably the interesting one - do we recover OK from TCP congestion?  (Maybe this isn't critical given that it's Bono.)

rkd-msw:
Given that this is Bono-only, I can't convince myself it's worth fixing up the merge conflicts or doing a lot of testing here - closing.

PR #1500, opened 2016-08-01
[Reviewer: Ellie] Remove existing Route headers before the BGCF adds them

rkd-msw:
In my testing, I hit a case where a message arrived at the BGCF with Route headers still attached. This meant that when the BGCF added Route headers, the message still followed the pre-existing ones.

I don't think that's sensible behaviour, so I've added code to strip all existing Route headers when we start adding BGCF routes. I've tested that this works.

I'll talk to you tomorrow about the exact flow that triggered this, for background.


eleanor-merry:
Change looks fine, but I'd like to see a UT for it

PR #1501, opened 2016-08-02
[Reviewer: AJL] Split storage debian packaging for Sprout/Bono

sathiyan-sivathas:
(1/6)

Several more pull requests to follow, but I'll put a summary of all of them in here:
- For Sprout, Ralf and Homestead, remove the database dependencies from the old packages and put them in the new "-node" packages.
- For Bono, Ellis, Homer and Homestead-prov, create new "-node" metapackages that are exactly the same as the existing packages so users can install packages with consistent names whether they want split storage or not.
- For the new nodes (which I've called Atlas and Janus for now since we've still not had any ideas) create packages that contain the database and Ralf/Homestead packages respectively.


sathiyan-sivathas:
I forgot to mention testing.

I've run `make deb` for all of the repos and everything built successfully.  All the packages are on my repo server so you can get them from there for your chef work.

johadalin:
Changes look fine. 

As this is the main PR, i'll put this comment here:
I think it is worth re-working the package description for the -node packages, so that they aren't identical to the packages they pull in. I doubt anyone would really notice, but it would be good to clarify that these are packages to install the full node, not just the system process

PR #1506, opened 2016-08-10
[Reviewer: Ellie] Make Sproutlet matching code more concise

rkd-msw:
@chris-elford-metaswitch flagged that the does_uri_match_sproutlet function was quite long could be simplified (after he looked at it for a mailing list issue).

I've pulled the function to extract service names out into a separate function, and deleted the checking of the "services" parameter (based on the comment "LCOV_EXCL_START - not currently supported").

### DIFF
--- a/src/ut/sproutletproxy_test.cpp
+++ b/src/ut/sproutletproxy_test.cpp
@@ -692,6 +692,12 @@ public:
     pjsip_tsx_layer_instance()->start();
   }
 
+  std::list<std::string> extract_services(std::string uri)
+  {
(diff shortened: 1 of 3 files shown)


src/ut/sproutletproxy_test.cpp
```diff
@@ -2299,3 +2305,25 @@ TEST_F(SproutletProxyTest, LocalNonSubscribe)
 
   delete tp;
 }
+
+// Tests standard routing of a subscription request to ensure it is
```
eleanor-merry: Test description is wrong

eleanor-merry:
Looks fine. I don't think you can drop support for the services parameter though (I think the comment was saying that we don't support services of the form `memento&something`, not that we don't support the services parameter at all).

PR #1507, opened 2016-08-11
[Review AJH] Allow for minimal Accept-Contact header

steelyvoid:
Fixes a problem whereby the (valid) header "Accept-Contact: *" was being rejected by parse_hdr_accept_contact.

Tested live using an OPTIONS message with the above header.  Custom header UTs pass.  I decided not to write a new UT to cover this case - let me know if you think thats the wrong call.

### DIFF
--- a/src/custom_headers.cpp
+++ b/src/custom_headers.cpp
@@ -1276,7 +1276,13 @@ pjsip_hdr* parse_hdr_accept_contact(pjsip_parse_ctx* ctx)
   pj_str_t header_value;
   pj_scan_get(scanner, &pc->pjsip_TOKEN_SPEC, &header_value);
 
-  for (;;)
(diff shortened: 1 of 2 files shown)


AlexHockey:
Looks good to me. Would a unit test be hard to add? Regression tests are generally a good thing.

steelyvoid:
OK.  I wasn't sure whether it was worth the increase in UT execution time, but I'll add one if you prefer

AlexHockey:
Thanks Steve. FYI we generally don't worry too much about execution time of the UTs (unless the UT is unusual in some way, which this isn't).

rkd-msw:
I think I'd say that we don't worry about the **aggregate** execution time of the UTs - an individual UT that took a second to run would be frowned upon, I think (but this UT will be milliseconds at most),

PR #1508, opened 2016-08-12
Alarms, comm links all always enabled

GAB-MS:
Removal of a few 'if' statements so that the alarms and comm links are always enabled, instead of only being enabled if I/S-CSCF functionality is enabled.

Please note - **this is not yet ready for review by the Clearwater team.** This code will first be internally reviewed by JAG2's team.

Furthermore, the changes shown below initially look horrible - the only changes are the removal of three 'if' statements, their associated braces, and reducing the indentation of the code they contained. Unfortunately, it would appear that these changes are causing problems for the comparison.

The code has been tested, and all Sprout UTs and the full suite of live tests all pass.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1625,63 +1625,60 @@ int main(int argc, char* argv[])
                                                       ".1.2.826.0.1.1578918.9.3.31");
   }
 
-  if (opt.enabled_icscf || opt.enabled_scscf)
-  {
-    // Create Sprout's alarm objects.
-    alarm_manager = new AlarmManager();
+  // Create Sprout's alarm objects.
+  alarm_manager = new AlarmManager();
 
-    chronos_comm_monitor = new CommunicationMonitor(new Alarm(alarm_manager,
+  chronos_comm_monitor = new CommunicationMonitor(new Alarm(alarm_manager,
+                                                            "sprout",
+                                                            AlarmDef::SPROUT_CHRONOS_COMM_ERROR,
+                                                            AlarmDef::MAJOR),
+                                                  "Sprout",
+                                                  "Chronos");
+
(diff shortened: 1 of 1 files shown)


rkd-msw:
You can append ?w=1 to the URL (e.g. https://github.com/Metaswitch/sprout/pull/1508/files?w=1) to get diffs excluding whitespace - you can't comment in this mode, but it's useful to get a better view of PRs like this.

huws:
I think you also need to remove the 'if' test around the code that starts the ENUM service (~line 1801).  At the moment it checks for S-CSCF or I-CSCF, but BGCF uses it, too.

mirw:
Should we be doing a review of all sproutlet-specific code in core sprout?  If this is more than you can take on, please raise an issue and we'll take a look.

huws:
Matt - I think we'll raise an issue for the sproutlet-specific code review.  We'll have some input to provide, as there are some things that we already know about (e.g. the options structure), but I think it would be better for you guys to own this.  That said, we may end up fixing further individual examples of this, as and when they become a problem for us.

huws:
Looks good - thanks.  Good catch on the chronos_connection - I'd missed that.

GAB-MS:
Rob, I'm assigning this to you for review, as you've been involved in the discussion of these code changes. Thank you

rkd-msw:
Looks good - presumably you've tested, not just that Sprout is still OK, but that this change does what you want it to? Fine to merge if so.

rkd-msw:
(although we're currently cutting the release, so don't merge until @AME2 emails to give the OK)

benritchie:
George - they must have finished cutting the release by now.  Can you check this is ready to merge, and if so merge please?

Thanks

PR #1515, opened 2016-08-22
[Reviewer: Ellie] Update to use the new clearwater-etcd-plugins submodule structure

johadalin:
I will be doing a full install using the packages built with this before merging anything, ensuring that the plugins still get put in the right locations. 
Same for the other components undergoing similar changes


.gitmodules
```diff
@@ -61,3 +61,6 @@
 [submodule "modules/memento-as"]
 	path = modules/memento-as
 	url = git@github.com:Metaswitch/memento-as.git
+[submodule "clearwater-etcd-plugins"]
```
eleanor-merry: Again, please put in modules

eleanor-merry:
Sure. Can you move your new submodule into the modules folder (creating it if necessary) for each PR though.

PR #1516, opened 2016-08-23
[Reviewer: EM] Rename SIP Connection Pool class

TJW2:
Rename the ConnectionPool class in include/connection_pool.h to SipConnectionPool to avoid a conflict with the template ConnectionPool class in cpp-common introduced in pull request [#546](https://github.com/Metaswitch/cpp-common/pull/546).

### DIFF
--- a/docs/Bono Sprout Architecture.md
+++ b/docs/Bono Sprout Architecture.md
@@ -120,7 +120,7 @@ When the stateful proxy module is running as a Bono node, it performs various ed
 
 When running as a Bono node all routing of requests is performed based on route headers and request URI.  If a request cannot be routed based on route headers or the request URI then Bono will always route it to a Sprout node - it does not access the registration store directly itself.  This means that a Bono node will never fork a SIP request.
 
-Each Bono node uses the SipConnectionPool class to manage a pool of TCP connections to the Sprout nodes for this purpose.  The SipConnectionPool class periodically recycles these connections to keep the load evenly spread across the Sprout nodes.
+Each Bono node uses the SIPConnectionPool class to manage a pool of TCP connections to the Sprout nodes for this purpose.  The SIPConnectionPool class periodically recycles these connections to keep the load evenly spread across the Sprout nodes.
 
 Bono tracks incoming client connections in the FlowTable class, including recording when the connection has been authenticated by the client correctly responding to a challenge from a Sprout node.  Each client connection is identified by a flow token which is used in Path and Route headers as per [RFC5626](http://tools.ietf.org/html/rfc5626) for ensuring that SIP flows for the client use the appropriate connection (both for security and NAT traversal purposes).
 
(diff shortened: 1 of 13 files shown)


modules/cpp-common
```diff
@@ -1 +1 @@
-Subproject commit e806b651a4ed6e24828849850235a526b6796d03
+Subproject commit 8a2e0b9b4f25b4cf5206e97e1df8785da5708507
```
TJW2: Is this change something to be worried about? I'm not sure what it's doing.

include/sipconnectionpool.h
```diff
@@ -1,5 +1,5 @@
 /**
- * @file connection_pool.h Class for maintaining a SIP connection pool.
+ * @file sipconnectionpool.h Class for maintaining a SIP connection pool.
```
eleanor-merry: Can this be sip_connection_pool rather than sipconnectionpool?

src/sipconnectionpool.cpp
```diff
@@ -1,5 +1,5 @@
 /**
- * @file connection_pool.cpp
+ * @file sipconnectionpool.cpp
```
eleanor-merry: Likewise, can this be sip_connection_pool as well?

docs/Bono Sprout Architecture.md
```diff
@@ -120,7 +120,7 @@ When the stateful proxy module is running as a Bono node, it performs various ed
 
 When running as a Bono node all routing of requests is performed based on route headers and request URI.  If a request cannot be routed based on route headers or the request URI then Bono will always route it to a Sprout node - it does not access the registration store directly itself.  This means that a Bono node will never fork a SIP request.
 
-Each Bono node uses the ConnectionPool class to manage a pool of TCP connections to the Sprout nodes for this purpose.  The ConnectionPool class periodically recycles these connections to keep the load evenly spread across the Sprout nodes.
+Each Bono node uses the SipConnectionPool class to manage a pool of TCP connections to the Sprout nodes for this purpose.  The SipConnectionPool class periodically recycles these connections to keep the load evenly spread across the Sprout nodes.
```
eleanor-merry: I think I'd rather SIPConnectionPool rather than SipConnectionPool, but I don't really care that much

eleanor-merry:
LGTM

docs/Bono Sprout Architecture.md
TJW2: Here I'm using the style used by HttpConnection - is there a convention for this?

include/sipconnectionpool.h
TJW2: sipconnectionpool is consistent with sipresolver, although the rest of the files in sprout/src are named with a mixture of both styles. Should I make this change anyway?

eleanor-merry: I think we've settled on using _ now (and new files follow this pattern), so can you make the change?

PR #1517, opened 2016-08-23
[Reviewer: Matt] Minor perf-test fixes

rkd-msw:
I ran clearwater-sip-perf on an all-in-one node - here are the fixes I needed to make that work.

* clearwater-sip-perf was marked as conflicting with sprout/bono/restund, but actually it runs fine.
* the home domain (example.com) doesn't resolve on an AIO node, so I gave sip-perf an optional argument which is the IP to run stress on.
* if a 100 Trying came in after the INVITE, we'd use its Record-Route headers on subsequent messages - but as 100s aren't end-to-end this meant in-dialog messages got misrouted. We already had some stuff to make this work for Via headers which I've just extended.

### DIFF
--- a/clearwater-sip-perf.root/usr/share/clearwater/bin/sip-perf
+++ b/clearwater-sip-perf.root/usr/share/clearwater/bin/sip-perf
@@ -9,6 +9,13 @@ ulimit -Sn 100000
 # Read in config.
 . /etc/clearwater/config
 
+stress_target=$home_domain:5060
+
+if [ -n "$1" ]
+then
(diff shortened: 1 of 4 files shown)


mirw:
Looks good.  Should any of the other fixes be applied to clearwater-sip-stress too?

rkd-msw:
No - the 100 Trying one makes sense, but I don't think it's useful to run clearwater-sip-stress on an all-in-one node.

mirw:
Fine.

Issue #1509, opened 2016-08-12
Restund can't be started by monit


eleanor-merry:
#### Symptoms
Noticed on the staging AIO node that Restund wasn't running. Monit.log is full of
```
[UTC Aug 12 18:12:03] error    : 'restund_process' process is not running
[UTC Aug 12 18:12:03] info     : 'restund_process' trying to restart
[UTC Aug 12 18:12:03] info     : 'restund_process' restart: /etc/init.d/restund
[UTC Aug 12 18:12:33] error    : 'restund_process' failed to restart (exit status 15) -- /etc/init.d/restund:  * Restarting Restund STUN/TURN server restund
```
Starting it with `service restund start` works just fine.
This is similar to https://github.com/Metaswitch/clearwater-monit/issues/41, but seen in the dev code 

#### Impact
Restund isn't running on Bono nodes

#### Release and environment 
From at least Ivysaur

richardwhiuk:
I dug into this more, as I didn't realise there was an open issue.

Running `sudo service restund restart` failed as did `sudo /etc/init.d/restund restart`.

Running `sudo bash /etc/init.d/restund restart` worked but `sudo bash` and then `/etc/init.d/restund/restart` in the subshell failed.

Doing the second showed me that bash was saying that it's process was being terminated, though this line doesn't appear if you run sudo /etc/init.d/restund restart.

Adding debug flags to the bash script led me to a conclusion.

When we call restart, we call do_stop and then do_start.

do_stop calls `start-stop-daemon --stop --quiet --retry=TERM/30/KILL/5 --name restund`, we kill the init script, so we never get round to calling do_start.

This is because the process name essentialy is 'restund' so we kill it.

The fix is probably just to add --exec $DAEMON to the start-stop args.

PR #1518, opened 2016-08-24
[Reviewer: RKD] Stop restund properly

richardwhiuk:
If we don't specify the executable, start-stop-daemon will kill the init script.

This fixes issue #1509.

I've tested this live and verifies that this fixes this issue. I've tested both with and without a running restund, and checked that the PID of the restund process changed, and that there wasn't one left behind.


rkd-msw:
:+1:

PR #1519, opened 2016-08-24
[Reviewer: Adam] Track the time from INVITE to 180 Ringing

rkd-msw:
I'm going to live test by:

- making a single call
- checking that the stats are updated, and that the recorded latency approximately matches the latency in SAS
- running the live tests and checking that this stat still looks sensible

I'll load up the MIB and use snmptable to check this, so I check that the MIB and the stat match.

### DIFF
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -432,7 +432,9 @@ void SCSCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
   pjsip_status_code status_code = PJSIP_SC_OK;
 
   // Store off the time we received this request, for statistics purposes.
-  _tsx_start_time = Utils::get_time();
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC_COARSE, &ts);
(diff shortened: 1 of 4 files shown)


rkd-msw:
Testing went fine.

src/scscfsproutlet.cpp
```diff
@@ -665,6 +676,18 @@ void SCSCFSproutletTsx::on_rx_response(pjsip_msg* rsp, int fork_id)
...
   {
+    // If this is a 180 Ringing response being sent to a caller (not to an app
+    // server or originating S-CSCF), calculate the time it's taken to get to
+    // the ringing point and update our statistics.
+    if (_session_case->is_originating() &&
+        _as_chain_link.complete() &&
+        _req_type == PJSIP_INVITE_METHOD &&
+        st_code == PJSIP_SC_RINGING)
+    {
+      uint64_t ringing_us = (Utils::get_time() - _tsx_start_time) * 1000;
```
johadalin: You suck, make this really microseconds!!!

johadalin:
LGTM :+1:

Issue #1514, opened 2016-08-22
Privacy header options delimiter is incorrect


richardwhiuk:
#### Symptoms

Privacy header options are parsed with a comma separated delimiter instead of a semi-colon, as defined by RFCC 3323 - https://www.ietf.org/rfc/rfc3323.txt

#### Impact

Privacy header options are stripped by Clearwater when several occur in a single header, when using MMTel.

#### Release and environment 

Reported on release-99 - Farfetch’d. Reported by Aravid Ashok Shirabur -http://lists.projectclearwater.org/pipermail/clearwater_lists.projectclearwater.org/2016-August/003152.html

### Analysis

Privacy headers are parsed by custom_headers.cpp which parses them into a pjsip_generic_array_hdr. When this is parsed, it is separated by commas, instead of semi colons.

richardwhiuk:
@mirw and I discussed this.

The Privacy header RFC is inconsistent with other headers, in that in uses `,` instead of `;`.

Currently, we use PJSIP's parser to parse the Privacy header in to a pjsip_generic_array_hdr. This parse phase, and the subsequent print phase, uses `,` for pjsip_generic_array_hdr.

The important files here in Sprout are https://github.com/Metaswitch/sprout/blob/dev/src/custom_headers.cpp and https://github.com/Metaswitch/sprout/blob/dev/src/mmtel.cpp#L502

We probably have two options:

1. We can add a pjsip_privacy_hdr type which splits/combines on `;`. This is the most natural fit when compared to Accept headers. This can either live in Sprout or in PJSIP. We'd probably want to tweak the parse_generic_array_hdr function to allows us to reuse as much of that code as possible.

2. Instead of parsing to a pjsip_generic_array_hdr, we parse to a generic string header, and then split the Privacy header inside the MMTel AS. This would involve less change to the upstream PJSIP, which doesn't support for the Privacy header, but would be less in keeping with the general design.

richardwhiuk:
This is fixed in https://github.com/Metaswitch/sprout/pull/1521

PR #1521, opened 2016-08-25
[Reviewer: RJW2] Privacy header now splits on semicolons rather than commas.

oa-metaswitch:
Modified the privacy header code to use a semicolon delimiter rather than commas, see issue #1514. The changes are:

1) Added parse and print functions for a privacy header to custom_headers.cpp and custom_headers.h
2) Modified the mmtel code to call the privacy header constructor rather than construct a generic array header.
This commit must be used in conjunction with the corresponding privacy_header branch of pjsip, for which a corresponding pull request will be made

### DIFF
--- a/include/custom_headers.h
+++ b/include/custom_headers.h
@@ -44,21 +44,11 @@
 
 extern "C" {
 #include <pjsip.h>
+#include <pjsip/print_util.h>
 }
 
 // Main entry point
 pj_status_t register_custom_headers();
-
-// Utility macro from sip_parser.c
-#define copy_advance(buf,str)                   \
-  do {                                          \
-    if ((str).slen >= (endbuf-buf)) return -1;  \
-    pj_memcpy(buf, (str).ptr, (str).slen);      \
-    buf += (str).slen;                          \
-  } while (0)
-
-// Function pointer from pjsip for overloading header print and clone methods.
-typedef int (*pjsip_hdr_print_fptr)(void *hdr, char *buf, pj_size_t len);
 
 /// Custom header structures.
 
--- a/include/custom_headers.h
+++ b/include/custom_headers.h
@@ -57,6 +57,9 @@ pj_status_t register_custom_headers();
     buf += (str).slen;                          \
   } while (0)
 
+// Function pointer from pjsip for overloading header print and clone methods.
+typedef int (*pjsip_hdr_print_fptr)(void *hdr, char *buf, pj_size_t len);
+
 /// Custom header structures.
 
 enum session_refresher_t
@@ -111,6 +114,7 @@ typedef struct pjsip_reject_contact_hdr {
 /// Utility functions (parse, create, init, clone, print_on)
 
 // Privacy
+pjsip_generic_array_hdr* pjsip_privacy_hdr_create( pj_pool_t *pool, const pj_str_t *hnames);
 pjsip_hdr* parse_hdr_privacy(pjsip_parse_ctx* ctx);
 
 // Assocciated URI
--- a/src/custom_headers.cpp
+++ b/src/custom_headers.cpp
@@ -53,34 +53,25 @@ extern "C" {
 /// generic array header, however it splits on commas. We therefore override
 /// the default parser and print function in the generic array header with a
 /// custom constructor.
-
-static int pjsip_privacy_hdr_print( pjsip_generic_array_hdr *hdr,
+static int pjsip_privacy_hdr_print(pjsip_generic_array_hdr *hdr,
                       char *buf, pj_size_t size)
 {
-    return pjsip_generic_array_hdr_delimited_print(hdr, buf, size, "; ",2);
+  return pjsip_generic_array_hdr_delimited_print(hdr, buf, size, "; ", 2);
 }
 
-
-pjsip_generic_array_hdr* pjsip_privacy_hdr_init( pj_pool_t *pool,
-                                   void *mem,
-                                   const pj_str_t *hnames)
-{
-    pjsip_generic_array_hdr *hdr = pjsip_generic_array_hdr_init(pool, mem,hnames);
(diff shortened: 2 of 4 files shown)


richardwhiuk:
This is a good change – it’d be excellent if we can get it in before we cut the release – that way we can advertise it’ll be in the release. I think that’s probably going to be at some point on Wednesday.

This is mostly fine – most of my comments are either stylistic – e.g. some tweaks regarding indentation,  or regarding ongoing maintenance.

One high level thought – PJSIP and Clearwater have different coding styles, so it’s best to try and match what is around you. 

Oh, and don’t forget to update the sub module reference :)

include/custom_headers.h
```diff
@@ -57,6 +57,9 @@ pj_status_t register_custom_headers();
     buf += (str).slen;                          \
   } while (0)
 
+// Function pointer from pjsip for overloading header print and clone methods.
+typedef int (*pjsip_hdr_print_fptr)(void *hdr, char *buf, pj_size_t len);
```
richardwhiuk: This is a redefinition of something in pjsip. That’s not a particularly good idea, because one of the definitions could change and you won’t get any warning that the definitions conflict.

As such, instead, you want to include the relevant header file in pjsip. I think that if you add `#include <print_util.h>` in the extern C block above, it’ll pull in the right definition.

src/custom_headers.cpp
```diff
@@ -46,16 +46,44 @@ extern "C" {
...
-/// the built-in PJSIP parser code.
+/// The privacy header is delimited by semicolons. We want to use pjsip's
+/// generic array header, however it splits on commas. We therefore override
+/// the default parser and print function in the generic array header with a
+/// custom constructor.
+
+static int pjsip_privacy_hdr_print( pjsip_generic_array_hdr *hdr,
+                      char *buf, pj_size_t size)
+{
+    return pjsip_generic_array_hdr_delimited_print(hdr, buf, size, "; ",2);
```
richardwhiuk: All of the code in this file you’ve added is incorrectly indented – it should be two space instead of four – if you haven’t been introduced to our coding conventions they are at http://clearwater.readthedocs.io/en/stable/Clearwater_CPP_Coding_Guidelines.html

src/custom_headers.cpp
```diff
@@ -46,16 +46,44 @@ extern "C" {
...
 /// in register_custom_headers().
 ///
-/// The Privacy header is a simple comma-separated list of values so we use
-/// the built-in PJSIP parser code.
+/// The privacy header is delimited by semicolons. We want to use pjsip's
+/// generic array header, however it splits on commas. We therefore override
+/// the default parser and print function in the generic array header with a
+/// custom constructor.
+
+static int pjsip_privacy_hdr_print( pjsip_generic_array_hdr *hdr,
```
richardwhiuk: There shouldn’t be a space after the opening bracket

src/custom_headers.cpp
```diff
@@ -46,16 +46,44 @@ extern "C" {
...
+                      char *buf, pj_size_t size)
+{
+    return pjsip_generic_array_hdr_delimited_print(hdr, buf, size, "; ",2);
+}
+
+
+pjsip_generic_array_hdr* pjsip_privacy_hdr_init( pj_pool_t *pool,
```
richardwhiuk: There shouldn’t be a space after the opening bracket

src/custom_headers.cpp
```diff
@@ -46,16 +46,44 @@ extern "C" {
...
+                                   void *mem,
+                                   const pj_str_t *hnames)
+{
+    pjsip_generic_array_hdr *hdr = pjsip_generic_array_hdr_init(pool, mem,hnames);
+    hdr->vptr->print_on = (pjsip_hdr_print_fptr) &pjsip_privacy_hdr_print;
+    return hdr;
+}
+
+pjsip_generic_array_hdr* pjsip_privacy_hdr_create( pj_pool_t *pool,
```
richardwhiuk: There shouldn’t be a space after the opening bracket

src/custom_headers.cpp
richardwhiuk: There is a missing space after the , and before the 2.

src/custom_headers.cpp
```diff
@@ -46,16 +46,44 @@ extern "C" {
...
+/// generic array header, however it splits on commas. We therefore override
+/// the default parser and print function in the generic array header with a
+/// custom constructor.
+
+static int pjsip_privacy_hdr_print( pjsip_generic_array_hdr *hdr,
+                      char *buf, pj_size_t size)
+{
+    return pjsip_generic_array_hdr_delimited_print(hdr, buf, size, "; ",2);
+}
+
```
richardwhiuk: Extra blank line

src/custom_headers.cpp
```diff
@@ -46,16 +46,44 @@ extern "C" {
...
+    void *mem = pj_pool_alloc(pool, sizeof(pjsip_generic_array_hdr));
+    return pjsip_privacy_hdr_init(pool, mem, hnames);
+}
+
 // LCOV_EXCL_START
 pjsip_hdr* parse_hdr_privacy(pjsip_parse_ctx* ctx)
 {
-  pjsip_generic_array_hdr *privacy = pjsip_generic_array_hdr_create(ctx->pool, &STR_PRIVACY);
-  pjsip_parse_generic_array_hdr_imp(privacy, ctx->scanner);
+  const pjsip_parser_const_t* pconst=pjsip_parser_const();
```
richardwhiuk: The equals sign wants spaces around it

src/custom_headers.cpp
richardwhiuk: I think you can declare this static – or alternatively inline inside the create function.

richardwhiuk:
Looks good.

oa-metaswitch:
I've updated the dependency on pjsip. Merging the branch now.

PR #1523, opened 2016-08-26
Build submodules with "make" rather than "make build"

oa-metaswitch:
The root makefile builds submodules recursively by calling make build in each plugin directory. However, including cpp.mk provides a make all, but not a make build-target. To make these slightly more consistent, I propose simply calling make in the plugin directory for the build, allowing the plugin itself to specify the default target.


rkd-msw:
:+1:

PR #1524, opened 2016-08-26
[Reviewer: Richard] Don't allow sprout_hostname to be an IP address

### LINKED ISSUE #1029: IPv6 is generally broken for the S-CSCF sproutlet

eleanor-merry:
We can't have sprout_hostname be an IP address when it's used to build up a URI - bail out immediately in this case

Fixes #1029 

(Not tested at all yet)


richardwhiuk:
I think this is the right check to make, if we are happy we don't allow ports in the sprout_hostname. Given the docs say: `a hostname that resolves by DNS round-robin to all Sprout nodes in the cluster`, I think I'm happy that's correct.

Surely we should do the check in main.cpp instead though - not every time we parse a sproutlet? https://github.com/Metaswitch/sprout/blob/6ebe053af31dbe9dc6a33803b7beeebcc8ed1951/src/main.cpp#L1120 looks like a sensible place.

PR #1525, opened 2016-08-30
[Reviewer AJH] Registration elements in notify for all impus in irs

steelyvoid:
Modifies the XML in NOTIFYs so that registration elements are duplicated for each IMPU in the Implicit Registration Set.  Prior to this fix, if a REGISTER was performed for an IMPU that is not the primary for the IRS, any reg-event NOTIFYs emitted would list the contact details for the registration under an element for the primary IMPU - not the IMPU the registration was made for - thereby confusing SUBSCRIBEd entities.

Tested live as well as through the modified UTs.

One detail worth highlighting in case its not what you expected: I've added calls to get_registration_data in a couple of places (get_reg_data in handlers.cpp and remove_bindings in registration_utils.cpp) in order to get hold of the IRS list in those places where we don't already have it.  I've adopted the approach that, if the call fails, or returns a blank list for whatever reason, I log and dummy up a default list consisting of the one IMPU we do have to minimise the impact on SUBSCRIBEd entities.


include/subscriber_data_manager.h
```diff
@@ -538,19 +544,35 @@ class SubscriberDataManager
...
                                      bool& all_bindings_expired = unused_bool,
                                      pjsip_rx_data* extra_message_rdata = NULL,
                                      pjsip_tx_data* extra_message_tdata = NULL);
 
+  /// Test-specific version of the above function that assumes that the AoR is
+  /// the only IMPU in the Implicit Registration Set.
+  ///
+  /// Production code should not use this entry point as it should be determining
+  /// this list by querying homestead (assuming it doesn't have the list in hand
+  /// already).
```
AlexHockey: Why do you need this method for the UTs?

src/handlers.cpp
```diff
@@ -351,6 +377,8 @@ void AoRTimeoutTask::handle_response()
   report_sip_all_register_marker(trail(), _aor_id);
 }
 
+
```
AlexHockey: Unnecessary blank lines.

src/handlers.cpp
```diff
@@ -160,6 +160,32 @@ static bool sdm_access_common(SubscriberDataManager::AoRPair** aor_pair,
   return true;
 }
 
+static bool get_reg_data(HSSConnection* hss,
+                         std::string aor_id,
+                         std::vector<std::string>& irs_impus,
+                         std::map<std::string, Ifcs> ifc_map,
```
AlexHockey: Do you ever use this parameter?

AlexHockey: OK, I see that you do. However the parameter doesn't work as it's not passed by reference.

AlexHockey: Is there any extra UT or live testing you should be doing to check if there are any other issues like this?

src/handlers.cpp
```diff
@@ -625,12 +666,9 @@ SubscriberDataManager::AoRPair* DeregistrationTask::deregister_bindings(
...
     // Deregister with any application servers
-    std::vector<std::string> uris;
-    std::map<std::string, Ifcs> ifc_map;
-    std::string state;
     TRC_INFO("ID %s", aor_id.c_str());
 
-    if (_cfg->_hss->get_registration_data(aor_id, state, ifc_map, uris, trail()) == HTTP_OK)
+    if (got_ifcs)
     {
       RegistrationUtils::deregister_with_application_servers(ifc_map[aor_id],
```
AlexHockey: As noted above, `ifc_map` will be empty here which means that 3rd party deregistration won't work.

src/registrar.cpp
```diff
@@ -217,6 +217,7 @@ bool get_private_id(pjsip_rx_data* rdata, std::string& id)
 SubscriberDataManager::AoRPair* write_to_store(
                    SubscriberDataManager* primary_sdm,         ///<store to write to
                    std::string aor,                            ///<address of record to write to
+                   std::vector<std::string> irs_impus,         ///(IMPUs in Implicit Registration Set
```
AlexHockey: You have a `(` rather than a `<`.

src/registration_utils.cpp
```diff
@@ -475,19 +476,37 @@ void RegistrationUtils::remove_bindings(SubscriberDataManager* sdm,
...
+
+  if ((http_code != HTTP_OK) || irs_impus.empty())
+  {
+    // We were unable to determine the set of IMPUs for this AoR.  Push the AoR
+    // we have into the IRS list so that we have at least one IMPU we can issue
+    // NOTIFYs for.
+    TRC_WARNING("Unable to get Implicit Registration Set for %s: %d", aor.c_str(), http_code);
+    irs_impus.clear();
+    irs_impus.push_back(aor);
+  }
```
AlexHockey: I'm a little surprised you need to re-read the registration data here. I would have thought you would already have the IMPUs in hand from a previous HSS dip.

src/ut/handlers_test.cpp
```diff
@@ -177,17 +177,33 @@ TEST_F(AoRTimeoutTasksTest, MainlineTest)
...
+      EXPECT_CALL(*mock_hss, get_registration_data(_, _, _, _, _))
+           .WillOnce(DoAll(SetArgReferee<3>(std::vector<std::string>(irs_impus)), //IMPUs in IRS
+                           Return(HTTP_OK)));
       EXPECT_CALL(*remote_store1, get_aor_data(aor_id, _)).WillOnce(Return(remote_aor1));
-      EXPECT_CALL(*remote_store1, set_aor_data(aor_id, remote_aor1, _, _, _, _)).WillOnce(Return(Store::OK));
+      EXPECT_CALL(*remote_store1, set_aor_data(aor_id, irs_impus, remote_aor1, _, _, _, _)).WillOnce(Return(Store::OK));
       EXPECT_CALL(*remote_store2, has_servers()).WillOnce(Return(true));
+      EXPECT_CALL(*mock_hss, get_registration_data(_, _, _, _, _))
+           .WillOnce(DoAll(SetArgReferee<3>(std::vector<std::string>(irs_impus)), //IMPUs in IRS
+                           Return(HTTP_OK)));
```
AlexHockey: This worries me slightly. We've had to do a homestead dip for each GR site, which doesn't scale very well. Can't you do one dip and reuse it for each write to the store?

src/ut/mock_hss_connection.h
```diff
@@ -63,6 +63,13 @@ class MockHSSConnection : public HSSConnection
                                   const std::string& private_user_identity,
                                   const std::string& type,
                                   SAS::TrailId trail));
+
+  MOCK_METHOD5(get_registration_data, HTTPCode(
+                                  const std::string& public_user_identity,
+                                  std::string& regstate,
+                                  std::map<std::string, Ifcs >& ifcs_map,
+                                  std::vector<std::string>& associated_uris,
+                                  SAS::TrailId trail));
```
AlexHockey: Alignment looks odd (admittedly it looks odd in the method above as well).

AlexHockey:
Thanks Steve. I've reviewed. A couple of issues inline, but hopefully nothing too major. 

One additional thought. The example XML from TS 24.229 is 

```xml
<?xml version="1.0"?>
<reginfo xmlns="urn:ietf:params:xml:ns:reginfo"
  xmlns:cp="urn:ietf:params:xml:ns:common-policy"
  xmlns:eri="urn:3gpp:ns:extRegInfo:1.0"
  version="0" state="full">
  <registration aor="sip:user1_public1@home1.net" id="as9"
    state="active">
    <contact id="76" state="active" event="registered"> 
      <uri>sip:[5555::aaa:bbb:ccc:ddd]</uri>
      <unknown-param name="audio"/>
    </contact>
  </registration>
  <registration aor="sip:user1_public2@home1.net" id="as10"
    state="active">
    <contact id="86" state="active" event="created"> 
      <uri>sip:[5555::aaa:bbb:ccc:ddd]</uri>
      <unknown-param name="audio"/>
    </contact>
    <cp:actions>
      <eri:rph ns="wps" val="1"/>
      <eri:privSender/>
    </cp:actions>
  </registration>
</reginfo>
```

Assuming that the contact ID is wrong (and is supposed to be 76 in both `contact` elements), these elements *do* distinguish between the event (one is registered and one is created). Presumably this is because the UE registered for `sip:user1_public1@home1.net` abd `sip:user1_public2@home1.net was implicitly registered. Does your code distinguish between these two cases? If not should it? It might be too  hard to implement for not much benefit, but I wanted to get your thoughts.

include/subscriber_data_manager.h
steelyvoid: There were a significant number of existing calls to set_aor_data in the UTs (about 30 or so), and they don't care about the value of the impus vector, so it seemed less of a disruptive change to leave them as is and just provide a UT-specific entry point.  Let me know if you want me to change those lines to pass a dummy impus vector.

src/handlers.cpp
steelyvoid: Thanks for the spot.  Bit odd that there wasn't an existing UT that caught this - I'll add one.

src/registration_utils.cpp
steelyvoid: I don't think I can source them in all cases.  For instance, third_party_register_failed calls remove_bindings and that is called from send_register_cb.  The plumbing seems to me to be getting a bit intrusive.  Let me know if you think this is worth pursuing.

steelyvoid:
No, it won't distinguish between those cases, and this was one of the reasons why I wasn't sure we'd understood the behaviour intended by the spec. Rob's view was that, while the fix that says "just duplicate the registration elements for all IMPUs in the IRS" is probably not what was intended, it is at least better behaviour than what we have at the moment (where the NOTIFY can come back with a registration element for an IMPU that the SUBSCRIBEr doesn't recognise).

It seems to me that to fix this properly, we'd need to put more information in the RegStore (which was my original intention, but you vetoed this on the grounds of migration pain - see our email thread "RE: Sorting NOTIFYs for implicit reg sets" and your response on 3/8/16).  Let me know if you can see a way round this.

steelyvoid:
Actually, I'm not convinced that "registered" means "explicitly registered" and "created" means "implicitly registered" as you suggest above.  I think (from the spec - don't really understand the code, but I could believe its doing this) that "registered" is what you use when you have just created a binding for the first time, and "created" is what you use when you're just reporting an existing binding that you have previously reported.

include/subscriber_data_manager.h
AlexHockey: It would be nice to have one method (mt read is that it's basically wrong to call this method without the IRS IMPUs). How hard would it be migrate the UTs?

AlexHockey:
> Actually, I'm not convinced that "registered" means "explicitly registered" and "created" means "implicitly registered" as you suggest above.

To quote from [RFC 3860](https://tools.ietf.org/html/rfc3680)

>   There are two ways in which [a binding] it can become active.  One is
>   through an actual SIP REGISTER request (corresponding to the
>   registered event), and the other is when the contact is created
>   administratively, or through some non-SIP means (the created event).

My interpretation of this is that registration of an implicit registration would be treated as a contact that was created administratively, and so should use the created event. Do you have a spec reference that contradicts this.

src/registration_utils.cpp
AlexHockey: The important factors are when do we end up doing an extra homestead dip and how frequent are the extra dips. If they are infrequent then the extra dips are fine. Can you figure this out and let me know?

include/subscriber_data_manager.h
steelyvoid: Not hard - just tedious.  I'll change those lines to call a common method so that any future changes to parameters are easier to make

steelyvoid:
I'm not sure I would call the implicit registration of all IMPUs in the IRS as "an administrative creation".  "created administratively" suggests to me a binding that has been set up by the systems administrator rather than through the action of a UE (though I confess I've no idea why anyone would do that).

I was looking at the stuff in 5.4.2.1.2 4.e.III and IV (TS24229), which suggests that what is important is whether this is a registration or a re-registration, though I confess I find the whole area somewhat unclear.

Can you discuss with Rob and decide whether created/registered depends on whether the REGISTER was for this IMPU or another one in the IRS, since, if so, I need to change the data model and potentially reimplement the PR.  We have already agreed that this fix might not address all the problems in this area, but does get us nearer to something that is right, so, if its unclear, I would suggest that we hold off making significant changes of this type until we get further evidence of what real world SUBSCRIBErs are actually expecting.

include/subscriber_data_manager.h
AlexHockey: Sounds great. Thanks!

AlexHockey:
My thoughts:

* My reading of the spec is that "created" *is* for implicit registration and "registered" is for explicit registration. From section 5.4.2.1.2 4.e.IV of TS 24.229

>IV)	has been automatically registered or registered by the S-CSCF, and has not been previously automatically registered: ...
>-	set the event attribute within the <contact> element to "created"; and

* However 100% spec compliance requires us to track how each binding/IMPU pair was registered. This goes even further than just storing the list of associated IMPUs sprouts regstore, and makes me even more nervous. 
* This PR definitely makes things better, so we should merge it even if it doesn't give us 100% spec compliance.

AlexHockey:
The reason I'm asking the questions now is I want to understand where this PR gets us in relation to 100% spec compliance, and what would be left to do.

steelyvoid:
RFC 5628 ("Registration Event Package Extension for Session Initiation Protocol (SIP) Globally Routable User Agent URIs (GRUUs)") seems to be the most complete consideration of IRSs and NOTIFY bodies (see 8.2 Example: Implicit Registration), and seems to suggest that the associated registrations have the following differences in the <contact> subelements of their <registration> elements:
-  Different contact IDs
-  Different GRUUs - it looks as though GRUUs are specific to an IMPU rather than the IRS
-  Different event values.  As you suggest above, it looks as though, for the initial registration the intent is that the IMPU used for the registration is "registered" and the associated IMPUs are "created".  Its not clear to me how this changes on a reregistration.

That seems to imply that, not only do we need to record which IMPU was used for the initial registration, we need to track different GRUUs implicitly created for each of of the IMPUs in the IRS.

I'm lost now.  I think we need someone who understands the terminology here (interactions of GRUUs and bindings) to work out what the intended model is.  Suggest we talk to RKD when he's free

src/registration_utils.cpp
AlexHockey: I'm happy that deregistration caused by action at the HSS, a failed 3rd party registration, and a flow failed are all non-mainline enough to not worry about. On the timeout case: with VoLTE going out of signal for 30 minutes and having your registration time out seems like a fairly common scenario. But that just ends up replacing a re-registration with a de-registration, so won't introduce any extra load. 

> When deregistering the bindings as part of a "DeregisterTask". We'll end up doing one dip per binding in this case.

I was a bit worried by this though. One dip per *binding*?! That could introduce a lot of extra load.

AlexHockey: Cool. In that case I'm happy.

src/handlers.cpp
steelyvoid: Fixed and a UT added for 3rd party deregistration

steelyvoid:
Hi Alex.  All comments addressed apart from the issue of how much further we need to go to match the spec.  Do you need this question resolved before we merge?

AlexHockey:
> Do you need this question resolved before we merge?

No I don't think we do. I understand where the gaps are, and I'm confident this is moving us in the right direction. Merge when ready.

PR #1529, opened 2016-09-05
Fix bug in script stopping sipp-stress

TJW2:
The clearwater-sip-stress script does not correctly kill the sip-stress process in the case that it is not caught when killing processes in the relevant pidfile. This pull request covers fixing this problem.


TJW2:
I've tested this change by starting the SIPp stress service, deleting its pidfile, then stopping the service and checking that the relevant processes are killed.

AlexHockey:
:+1:

PR #1532, opened 2016-09-13
[Reviewer: RKD] Fix header printing

richardwhiuk:
Instead of changing the vptr table for all generic array headers,
instead provide a different vptr table for privacy headers.

This fixes a bug introduced in https://github.com/Metaswitch/sprout/pull/1521 whereby we will incorrectly print headers (e.g. Supported) if we've created a Privacy header - we end up printing: `Supported: replaces; norefersub; extended-refer; timer; X-cisco-serviceuri` instead of `Supported: replaces, norefersub, extended-refer, timer, X-cisco-serviceuri`

This requires the separately assigned https://github.com/Metaswitch/pjsip-upstream/pull/50 change.

I've tested by running the tests - need to investigate why this doesn't always cause failures.

### DIFF
--- a/src/custom_headers.cpp
+++ b/src/custom_headers.cpp
@@ -60,25 +60,30 @@ static int pjsip_privacy_hdr_print(pjsip_generic_array_hdr *hdr,
   return pjsip_delimited_array_hdr_print(hdr, buf, size, &semicolon_delimiter);
 }
 
+static pjsip_hdr_vptr privacy_hdr_vptr =
+{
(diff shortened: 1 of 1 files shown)


rkd-msw:
Looks good. Speculation on why it doesn't always cause failures:

* log level being different, so stuff doesn't get printed
* sending stuff out over the network (e.g. to app servers or separated S-CSCF) - the print functions are used to serialise it for sending over the network

PR #1534, opened 2016-09-16
[Reviewer: Richard] Downgrade spurious TRC_ERROR to TRC_DEBUG

### LINKED ISSUE #1531: 'error' log for dull signaling flow
#### Symptoms
This log:

    09-09-2016 01:28:03.706 UTC Error basicproxy.cpp:402: Rejecting CANCEL request with 481 status code

#### Impact
On our server we have set up alerts for error logs from sprout, in the expectation that error logs indicate things that we might care about.  

This looks as though it's some fairly normal window condition in the signaling that I really don't need to be alerted about.

Suggest that this log should be a warning at most, probably only info.

#### Release and environment 
Recent sprout code.

#### Steps to reproduce
I don't have trace but, presumably, just send in a CANCEL for a call that doesn't exist any more.

mirw:
This fixes #1531.  I've run UT but not repro-ed the exact scenario.  However, I really can't see how I'd have got this wrong.

### DIFF
--- a/src/basicproxy.cpp
+++ b/src/basicproxy.cpp
@@ -396,7 +396,7 @@ void BasicProxy::reject_request(pjsip_rx_data* rdata, int status_code)
 {
   if (rdata->msg_info.msg->line.req.method.id != PJSIP_ACK_METHOD)
   {
-    TRC_ERROR("Rejecting %.*s request with %d status code",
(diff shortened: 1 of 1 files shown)


richardwhiuk:
As discussed, we think the flows here are covered well enough by other debugging, and as discussed this definitely isn't an error condition. As such I'm happy.

PR #1535, opened 2016-09-19
Don't generate response to ACK requests [reviewer RKD]

### LINKED ISSUE #1504: Exception hit while processing ACK causes crash
#### Symptoms

While testing under stress, I hit https://github.com/Metaswitch/sprout/issues/1488 and then Sprout crashed.  Exceptions like the above should be handled, but I then hit the following trace:

```
08-08-2016 16:28:49.707 UTC Error thread_dispatcher.cpp:159: Exception SAS Trail: 507134 (maybe)
08-08-2016 16:28:49.707 UTC Error thread_dispatcher.cpp:164: Exception Call-Id: 2010008542-2.000000///20729-10322@192.168.0.18
08-08-2016 16:28:49.707 UTC Error thread_dispatcher.cpp:171: Exception CSeq: 16 ACK
08-08-2016 16:28:49.746 UTC Error pjsip: Assert failed: ../src/pjsip/sip_util.c:505 req_msg->line.req.method.id != PJSIP_ACK_METHOD
```

The code tries to handle but then tries to send a response, and this hits a follow-on exception, which isn't handled.  We shouldn't try to send a response if the failed request was an ACK.

#### Impact

Bono crashed when it shouldn't.  Temporary Bono outage while it restarted.

#### Release and environment 

Custom build from https://github.com/Metaswitch/sprout/tree/7800335bf98bec7e88eef0f3798977c8fbf19660, running on OpenStack on Aarch64

#### Steps to reproduce

Hit an exception while processing an ACK, most easily by running a multi-core Bono.

pdavies:
Fixes #1504

### DIFF
--- a/src/common_sip_processing.cpp
+++ b/src/common_sip_processing.cpp
@@ -354,7 +354,7 @@ static pj_bool_t process_on_rx_msg(pjsip_rx_data* rdata)
 
     if (rdata->msg_info.msg->type == PJSIP_REQUEST_MSG)
     {
-      if(rdata->msg_info.msg->line.req.method.id == PJSIP_ACK_METHOD)
+      if (rdata->msg_info.msg->line.req.method.id == PJSIP_ACK_METHOD)
       {
         TRC_WARNING("Dropping malformed ACK request");
       }
--- a/src/common_sip_processing.cpp
+++ b/src/common_sip_processing.cpp
@@ -330,8 +330,8 @@ static pj_bool_t process_on_rx_msg(pjsip_rx_data* rdata)
     return PJ_TRUE;
   }
 
-  // If a message has parse errors, reject it (if it's a request) or
-  // drop it (if it's a response).
+  // If a message has parse errors, reject it (if it's a request other than ACK)
+  // or drop it (if it's a response or an ACK request).
   if (!pj_list_empty((pj_list_type*)&rdata->msg_info.parse_err))
   {
     SAS::TrailId trail = get_trail(rdata);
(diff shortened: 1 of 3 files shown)


src/common_sip_processing.cpp
```diff
@@ -354,13 +354,20 @@ static pj_bool_t process_on_rx_msg(pjsip_rx_data* rdata)
...
     if (rdata->msg_info.msg->type == PJSIP_REQUEST_MSG)
     {
-      TRC_WARNING("Rejecting malformed request with a 400 error");
-      PJUtils::respond_stateless(stack_data.endpt,
-                                 rdata,
-                                 PJSIP_SC_BAD_REQUEST,
-                                 NULL,
-                                 NULL,
-                                 NULL);
+      if(rdata->msg_info.msg->line.req.method.id == PJSIP_ACK_METHOD)
```
rkd-msw: Can you add a space between "if" and "(", both here and below?

src/thread_dispatcher.cpp
```diff
@@ -172,15 +172,21 @@ static int worker_thread(void* p)
         }
 
         // Make a 500 response to the rdata with a retry-after header of
-        // 10 mins
-        pjsip_retry_after_hdr* retry_after =
+        // 10 mins unless the request was an ACK
+
+        if(!((rdata->msg_info.msg->type == PJSIP_REQUEST_MSG) &&
```
rkd-msw: Hmm - it looks like we'll pass this check on a response (where `rdata->msg_info.msg->type != PJSIP_REQUEST_MSG`) and attempt to return 500. Do you agree?

pdavies:
These tweaks are done now @rkd-msw

src/thread_dispatcher.cpp
```diff
@@ -172,15 +172,21 @@ static int worker_thread(void* p)
         }
 
         // Make a 500 response to the rdata with a retry-after header of
-        // 10 mins
-        pjsip_retry_after_hdr* retry_after =
+        // 10 mins unless the request was an ACK
+
+        if ((rdata->msg_info.msg->type == PJSIP_REQUEST_MSG) &&
+           !(rdata->msg_info.msg->line.req.method.id == PJSIP_ACK_METHOD))
```
rkd-msw: You can do `(x != y)` - that's clearer than `!(x == y)`.

rkd-msw:
One last comment, then it's fine to merge this.

PR #1538, opened 2016-09-22
[Reviewer: RKD] Add memento -node split-storage packages

sathiyan-sivathas:
Add memento -node packages.

Tested in a split-storage deployment.  There will be a further pull request with function changes.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -129,19 +129,6 @@ Priority: extra
 Depends: gemini-as (= ${binary:Version})
 Description: Debugging symbols for gemini, the Mobile Twinning Application Server plug-in
 
-Package: memento-as-node
-Architecture: any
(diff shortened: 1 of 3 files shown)


debian/control
```diff
@@ -129,9 +129,22 @@ Priority: extra
 Depends: gemini-as (= ${binary:Version})
 Description: Debugging symbols for gemini, the Mobile Twinning Application Server plug-in
 
+Package: memento-as-node
+Architecture: any
+Depends: sprout-base, memento-cassandra
```
rkd-msw: memento-as as well?

sathiyan-sivathas: Yes. I realise I've made the same mistake elsewhere too.

PR #1539, opened 2016-09-22
[Reviewer: RKD] Add cassandra hostname option for memento plugin

sathiyan-sivathas:
If we're using memento and `cassandra_hostname` is set, add it to memento's plugin options.  When the memento plugin connects to cassandra, use `cassandra_hostname` (if it is set) in preference to `localhost`.

Tested that the plugin can connect to a remote cassandra using the live tests.

### DIFF
--- a/src/mementoasplugin.cpp
+++ b/src/mementoasplugin.cpp
@@ -107,8 +107,7 @@ bool MementoPlugin::load(struct options& opt, std::list<Sproutlet*>& sproutlets)
                                                     "Memcached");
 
       // If the memento cassandra hostname option is set, use that instead of "localhost".
-      if ((opt.plugin_options.count("memento")) &&
-          (opt.plugin_options.find("memento")->second.count("cassandra")))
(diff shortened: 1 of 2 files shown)


src/mementoasplugin.cpp
```diff
@@ -105,8 +106,15 @@ bool MementoPlugin::load(struct options& opt, std::list<Sproutlet*>& sproutlets)
                                                     "Memento",
                                                     "Memcached");
 
+      // If the memento cassandra hostname option is set, use that instead of "localhost".
+      if ((opt.plugin_options.count("memento")) &&
+          (opt.plugin_options.find("memento")->second.count("cassandra")))
+      {
+        cassandra = opt.plugin_options.find("memento")->second.find("cassandra")->second;
```
rkd-msw: `opt.plugin_options["memento"]["cassandra"]` should be the empty string if it's not set, I think, which might be a cleaner check. (["x"] default-constructs a value if "x" is not found, and both std::map and std::string have default constructors.)

Issue #1008, opened 2015-04-17
Bono doesn't SAS log rejected requests


mirw:
When Bono rejects a request (e.g. because it's not authenticated and it's not a REGISTER), it sends SAS events for it, but doesn't send any markers and so the call can't be found.

Normally, the SAS markers would be applied in `UASTransaction::log_on_tsx_start`.  However, Bono can reject requests before this point.  In this case, `reject_request` in stateful_proxy.cpp is called instead, and that doesn't apply SAS markers.

AlexHockey:
Crucially, it's the start and end time markers that don't exist - the start time marker is crucial for SAS to be able to display the trace.

rkd-msw:
Fixed by the above PR.

PR #1541, opened 2016-09-22
[Reviewer: Graeme] Add start and end markers when bono statelessly rejects a request

AlexHockey:
When bono statelessly rejects a request it does not add start and end time markersm, which means the trace cannot be displayed in SAS. 

I've tested this by hacking clearwater-live-test to make a call without registering the caller (which causes bono to statelessly reject with a 403). Before the change the call did not appear in SAS results but after the change it did.

### DIFF
--- a/src/bono.cpp
+++ b/src/bono.cpp
@@ -740,8 +740,8 @@ static int proxy_verify_request(pjsip_rx_data *rdata)
 /// Rejects a request statelessly.
 static void reject_request(pjsip_rx_data* rdata, int status_code)
 {
-  // Log start and end markers. These are need for the failed request to appear
-  // in SAS.
+  // Log start and end markers. These are needed for the failed request to
+  // appear in SAS.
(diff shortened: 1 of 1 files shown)


AlexHockey:
Randomly victimizing @graemerobertson for code review.

AlexHockey:
Fixes #1008

src/bono.cpp
```diff
@@ -740,6 +740,13 @@ static int proxy_verify_request(pjsip_rx_data *rdata)
 /// Rejects a request statelessly.
 static void reject_request(pjsip_rx_data* rdata, int status_code)
 {
+  // Log start and end markers. These are need for the failed request to appear
```
graemerobertson: need**ed**

AlexHockey:
It took a bit of digging to figure out it was these markers that were missing, so it wasn't completely trivial ;-)

PR #1542, opened 2016-09-23
[Reviewer: RKD] BGCF-23 PJUtils updates 

jack-atack:
Added method to extract parameter values from either TEL request URIs, or the userinfo section of SIP request URI's.

### DIFF
--- a/src/pjutils.cpp
+++ b/src/pjutils.cpp
@@ -2171,22 +2171,22 @@ bool PJUtils::get_rn(pjsip_uri* uri, std::string& routing_value)
   return rn_set;
 }
 
-pjsip_param* PJUtils::get_tel_uri_param(pjsip_uri* uri, pj_str_t STR_PARAM)
+pjsip_param* PJUtils::get_userpart_param(pjsip_uri* uri, pj_str_t param)
 {
-  pjsip_param* param_ptr = NULL;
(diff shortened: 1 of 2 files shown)


src/pjutils.cpp
```diff
@@ -2171,6 +2171,25 @@ bool PJUtils::get_rn(pjsip_uri* uri, std::string& routing_value)
   return rn_set;
 }
 
+pjsip_param* PJUtils::get_tel_uri_param(pjsip_uri* uri, pj_str_t STR_PARAM)
```
rkd-msw: STR_PARAM isn't a constant so it shouldn't be in caps - maybe just `param`?

src/pjutils.cpp
```diff
@@ -2171,6 +2171,25 @@ bool PJUtils::get_rn(pjsip_uri* uri, std::string& routing_value)
...
+{
+  pjsip_param* param_ptr = NULL;
```
rkd-msw: Does this need to be called `_ptr`? We know it's a pointer from the definition, and we don't normally keep type information in the name. I'm not too worried, though.

rkd-msw:
Looks good - couple of comments about naming.

jack-atack:
Amendments have been made.  Will merge this in once the code has been unfrozen

PR #1545, opened 2016-09-26
[Reviewer: Seb] Build memento with the cassndra connection pool

AlexHockey:
This fixes the following error when trying to start sprout using commit [94ad54](https://github.com/Metaswitch/sprout/commit/94ad54da9be24b94e3f39d8bb32ae63aa9300c51)

26-09-2016 14:05:53.785 UTC Error pluginloader.cpp:148: Error loading Sproutlet plug-in /usr/share/clearwater/sprout/plugins/memento-as.so - /usr/share/clearwater/sprout/lib/libcassandra.so: undefined symbol: _ZN6apache6thrift12GlobalOutputE

The error happens because memento-as.so loads libcassandra.so, which depends on _ZN6apache6thrift12GlobalOutputE that should be provided by libthrift.so. Thrift used to be loaded by memento-as.so, but this stopped happening when we introduced the cassandra connection pooling stuff. 

Fix is to build memento-as.so with cassandra_connection_pool.so


sebrexmetaswitch:
Thanks, LGTM

PR #1551, opened 2016-09-29
[Reviewer AJH] 'Fetch binding' REGISTERs

steelyvoid:
This PR adds support for "Fetch binding" REGISTERs - i.e. REGISTERs that don't include Contact headers - for which the registrar is supposed to respond (after authentication) with the current binding list _without_ actually performing a  REGISTER.  Support for these is specifically mentioned in TS24.229 (in chapter 5.1.1.4.1, it states:
"The UE can fetch bindings as defined in RFC 3261 [26] at any time after the initial registration has been completed. The procedure for fetching bindings is the same as for a reregistration except that the REGISTER request does not contain a Contact header field.")

The fix adds support for Contact-less REGISTERs.  In addition, while adding UTs to cover the (previously uncovered) existing code between https://github.com/Metaswitch/sprout/blob/dev/src/registrar.cpp#L266 and https://github.com/Metaswitch/sprout/blob/dev/src/registrar.cpp#L333 (which I was cribbing for my own code and therefore needed to cover anyway), I found a bug in the code that copies AoR bindings from the backup AoRPair and the target AoRPair (the later deletion of the former caused storage pointed to by the latter to be release).  I fixed this and refactored the code to ensure that it got covered by all UTs.

I've also tested the fix live, ensuring that bindings are returned as expected both for IRSes with bindings and for IRSes without bindings


src/registrar.cpp
```diff
@@ -479,6 +530,346 @@ SubscriberDataManager::AoRPair* write_to_store(
...
+
+  // Add a PCFA header.
+  PJUtils::add_pcfa_header(tdata->msg, tdata->pool, ccfs, ecfs, true);
+
+  return;
+}
+
+// Process a "fetch bindings" REGISTER and send the response.  Don't increment stats for
+// this call as its not a real state-affecting register - just a query.
+void process_fetch_bindings(pjsip_rx_data* rdata, int now, SAS::TrailId trail)
```
AlexHockey: I'm not sure about is whether it is right to split out the main processing into a separate function. This has a reasonable amount of duplicate code (for policing the URI scheme, SAS logging, adding certain headers, etc). Also some code seems to be missing (for example I'd expect these registrations to generate ACRs), but that would just lead to more duplication. Can you explain you decision for me a bit please?

steelyvoid: The reason why I split the command out at a high level was that, to me, these fetch bindings REGISTERs are a complete different command from a real REGISTER (the only significant piece of function they share is the reporting of existing bindings, which is why I cracked out the response construction) and its more natural in a code structure sense.  I found the process_register routine hard enough to read as it is, and felt that I'd be making the code even more unreadable if I tried to put in all the fetch-bindings specific processing inline in the same place.

On the specific point about ACRs, my concern here is that TS32.299 seems to assume that REGISTERs are dialog stateful (see e.g. 7.2.35 Cause-Code AVP, code -3).  If we report these fetch binding REGISTERs via AVPs, isn't there a risk that subscribers might start being charged for them as for a real REGISTER?

steelyvoid: If I were to commonalise the common error branches (into common checks before making the HSS and store calls, and common checks afterwards), would that address your concern?  I avoided this because I didn't want to be passing the various error codes by reference and thought it would make the code structure harder to follow (for instance, we'd have to skip the statistic updates for fetch binding REGISTERs), but perhaps this would be a compromise that meets your desire to minimise the amount of similar looking code?

AlexHockey: I felt that a REGISTER with no contacts was just a normal REGISTER that *didn't* do something (namely, modify the bindings). I think that is why SIP allows you to fetch the bindings using a REGISTER, rather than by using some other mechanism. I'm not sure this is the right way of looking at things, just the way I did instinctively. 

My main concern is that in future we are going to add some new functionality and just not add it to the "fetch bindings" case because we forget about it, and nothing is going to catch this. I wonder whether the right thing is to carve out the bits that are different in the "no bindings" and "bindings present cases" but keep most of the processing in the `process_register_request`. I'd like to think about this over the weekend. Lets chat Monday is that's OK?

rkd-msw: I agree with Alex - a REGISTER updates N bindings (one per Contact header), then returns all the existing bindings on the response. A "fetch bindings" register just describes the behaviour when N=0. (I've tweaked the code a bit, and the existing code does essentially handle this, if you add some checks for `emergency_bindings >0` and don't call `set_aor_data` if `num_bindings == 0`).

I'm also not convinced that we shouldn't be updating statistics here - if there are real-world use-cases for this, then our stats should track bugs in it. Maybe we should increment the re-register statistics (on the principle that "The procedure for fetching bindings is the same as for a reregistration except that the REGISTER request does not contain a Contact header field"), or maybe we should have a new table.

I do agree that `process_register_routine` is quite labyrinthine, though, and I wonder if the right refactoring is to split it into phases:

- look at the REGISTER and pull out information (number of Contact headers, private ID, public ID)
- setup (e.g. initial SAS logging, creating the ACR)
- initial validation (rejecting emergency deregistrations, etc.)
- the actual registration interactions of talking to Homestead and getting/updating bindings
- creating the 200 OK response and adding all the necessary headers
- teardown (sending and destroying the ACR, sending SAS end marker, updating stats)

Maybe these would be methods on a RegisterTsx, so you could still share state easily between them.

rkd-msw: See https://github.com/Metaswitch/sprout/compare/fetch_bindings_2 for my "I've tweaked the code a bit" stuff, and https://github.com/Metaswitch/sprout/blob/e3f3d79aa1433948c66a14d9c36e16bb7ba6b9b3/src/registrar.cpp for a stab at splitting `process_register_request` into phases, although I'm not quite happy with that yet.

AlexHockey: Steve, are you happy to combine the "fetch bindings" processing with the "normal" register processing as Rob and I have suggested? FWIW I would be happy with a change along the lines of Rob's tweak if you don't have the time for a bigger refactor (I would rather have the code structured the way we want it from the outset even if it ends up being more complicated, as that feels like easier technical debt to deal with in the future).

src/authentication.cpp
```diff
@@ -664,25 +664,32 @@ static pj_bool_t needs_authentication(pjsip_rx_data* rdata, SAS::TrailId trail)
...
+                                                              contact_hdr->next);
+      }
+
+      if (emergency_reg)
+      {
+        SAS::Event event(trail, SASEvent::AUTHENTICATION_NOT_NEEDED_EMERGENCY_REGISTER, 0);
+        SAS::report_event(event);
+
+        return PJ_FALSE;
+      }
```
steelyvoid: That's not what the test is doing.  Its only SAS logging and returning if ALL of the Contact headers are emergency contacts (which is why it was erroneously making the SAS log if there were no Contacts at all).  Of course, the intended function may be just wrong - let me know if you need me to change it

AlexHockey: My mistake. I would like the variable renamed to something like `all_bindings_emergency" though if that's OK.

src/ut/registrar_test.cpp
```diff
@@ -193,10 +194,103 @@ class RegistrarTest : public SipTest
...
+                                     &SNMP::FAKE_REGISTRATION_STATS_TABLES,
+                                     &SNMP::FAKE_THIRD_PARTY_REGISTRATION_STATS_TABLES);
+    ASSERT_EQ(PJ_SUCCESS, ret);
+  }
+
+  RegistrarTestRemoteSDM() : RegistrarTest()
+  {
+    _remote_data_store_no_bindings->flush_all();
+                                 // start from a clean slate on each test
+  }
```
steelyvoid: OK.  I used the same model as used by the original RegistrarTest (which does its flushes in the constructor) for simplicity - do you want me to change both to do their flushes in Setup()?

src/ut/registrar_test.cpp
```diff
@@ -335,6 +442,21 @@ TEST_F(RegistrarTest, BadScheme)
...
+TEST_F(RegistrarTest, BadSchemeFetchBindings)
+{
+  Message msg;
+  msg._scheme = "sips";
+  msg._contact = "";
+  inject_msg(msg.get());
+  ASSERT_EQ(1, txdata_count());
+  pjsip_msg* out = current_txdata()->msg;
+  out = pop_txdata()->msg;
+  EXPECT_EQ(404, out->line.status.code);
```
steelyvoid: Thats what the pre-existing BadScheme test (for regular REGISTERs) expects.  Do you need me to change the function here?

src/ut/registrar_test.cpp
```diff
@@ -1567,6 +1755,20 @@ TEST_F(RegistrarTest, AssociatedUrisNotFound)
...
 
+TEST_F(RegistrarTest, AssociatedUrisNotFoundFetchBindings)
+{
+  Message msg;
+  msg._user = "6505550232";
+  msg._contact = "";
+  inject_msg(msg.get());
+  ASSERT_EQ(1, txdata_count());
+  pjsip_msg* out = current_txdata()->msg;
+  EXPECT_EQ(403, out->line.status.code);
```
steelyvoid: Search me.  Again, its what the preexisting test (immediately above) expects.  Now that we're doing "no contacts" REGISTERs inline, I might be able to retire some of these 'fetch binding' specific tests without losing code coverage, since that's all they're there for.

src/ut/registrar_test.cpp
```diff
@@ -2461,6 +2663,13 @@ class RegistrarTestMockStore : public SipTest
     delete _sdm; _sdm = NULL;
     delete _local_data_store; _local_data_store = NULL;
     delete _chronos_connection; _chronos_connection = NULL;
+
+    ((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_REGISTRATION_STATS_TABLES.init_reg_tbl)->reset_count();
+    ((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_REGISTRATION_STATS_TABLES.re_reg_tbl)->reset_count();
+    ((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_REGISTRATION_STATS_TABLES.de_reg_tbl)->reset_count();
+    ((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_THIRD_PARTY_REGISTRATION_STATS_TABLES.init_reg_tbl)->reset_count();
+    ((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_THIRD_PARTY_REGISTRATION_STATS_TABLES.re_reg_tbl)->reset_count();
+    ((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_THIRD_PARTY_REGISTRATION_STATS_TABLES.de_reg_tbl)->reset_count();
```
steelyvoid: Nope.  Again, just copied from the RegistrarTest constructor higher up.  I'll see if I can remove them (from here and the original)

src/ut/registrar_test.cpp
```diff
@@ -2527,3 +2736,199 @@ TEST_F(RegistrarTestMockStore, SubscriberDataManagerGetsFail)
...
+  ASSERT_EQ(1, txdata_count());
+  pjsip_msg* out = current_txdata()->msg;
+  EXPECT_EQ(500, out->line.status.code);
+  free_txdata();
+}
+
+// Test multiple registrations with a local and first remote SDM that are not
+// returning bindings (so that the bindings are returned by the second remote
+// SDM)
+TEST_F(RegistrarTestRemoteSDM, MultipleRegistrations)
```
steelyvoid: Its a straight copy of the MultipleRegistrations test under the RegistrarTest group, and it does need to set up multiple registrations so that it can test returning multiple bindings when the local SDM and the first of the remote SDMs isn't returning any data.

Therefore I can't split it up into multiple tests since I don't want the store to be flushed between each test.  I could try trimming it down so that it does the minimum number of tests required to cover the code if you prefer?

src/authentication.cpp
steelyvoid: Will do

src/ut/registrar_test.cpp
steelyvoid: Can't get rid of the casts.  Given that this is a pre-existing restriction, are you OK to leave as is?

src/pjutils.cpp
```diff
@@ -647,14 +647,24 @@ int PJUtils::max_expires(pjsip_msg* msg, int default_expires)
...
     {
-      max_expires = expires;
+      int expires = (contact->expires != -1) ? contact->expires : default_expires;
+      if (expires > max_expires)
+      {
+        max_expires = expires;
+      }
+      contact = (pjsip_contact_hdr*)pjsip_msg_find_hdr(msg, PJSIP_H_CONTACT, contact->next);
     }
-    contact = (pjsip_contact_hdr*)pjsip_msg_find_hdr(msg, PJSIP_H_CONTACT, contact->next);
```
AlexHockey: Ah, OK. Is it right to return the default value when there are no contact headers, or should we return 0? Code like https://github.com/Metaswitch/sprout/blob/386ecaf14f569a8444e660f14090dc8f41c9166c/src/acr.cpp#L284 suggests that is required (since it needs the default to be returned if there are no bindings), but code like https://github.com/Metaswitch/sprout/blob/b8e455ab169efadd57f5db2ddb2190fd33947dc5/src/bono.cpp#L1684 suggests otherwise (as this will cause bono to create a flow when the bindings are being refreshed).

src/ut/registrar_test.cpp
AlexHockey: No, just something to bear in mind for the future :-)

src/ut/registrar_test.cpp
```diff
@@ -246,46 +340,59 @@ class Message
...
+  contact[0] = 0;
+  if (!_contact.empty())
+  {
+    n = snprintf(contact, sizeof(contact),
+                 "Contact: %1$s%2$s%3$s\r\n",
+                 /*  1 */ (_contact == "*") ? "*" : string("<").append(_contact).append(">").c_str(),
+                 /*  2 */ _contact_params.c_str(),
+                 /*  3 */ _contact_instance.c_str()
+                 );
+    EXPECT_LT(n, (int)sizeof(buf));
```
AlexHockey: Fine (it looks like this is because ASSERT does an early return, so won't work here because the function has a different return type).

src/ut/registrar_test.cpp
AlexHockey: Not if it's pre-existing behavior.

src/ut/registrar_test.cpp
AlexHockey: Thanks. I can't see why they're needed buy maybe I'm missing something.

src/ut/registrar_test.cpp
AlexHockey: If this is a copy of an existing test I care less about it (though if the tests are literally the same maybe the code should be moved to a common function). Though its a shame such a long test ever existed in the first place!

AlexHockey:
Hi Steve, generally I'm OK with your comments - there is just one question left about what `max_expires` should return when there are no bindings, and I'm pretty sure different callers need different things :-(

src/ut/registrar_test.cpp
AlexHockey: OK. It's a shame no-one ever took the time to add some helper methods to reset the stats because the signal to noise ratio is terrible. Not worth fixing under this PR though.

src/ut/registrar_test.cpp
steelyvoid: Agree on commonalisation - I'll do that

src/ut/registrar_test.cpp
steelyvoid: I'm going to get rid of some of these duplicate tests as there's no need for them now that we're inlined the fetch bindings function

include/subscriber_data_manager.h
```diff
@@ -116,6 +116,16 @@ class SubscriberDataManager
       pjsip_sip_uri* pub_gruu(pj_pool_t* pool) const;
       std::string pub_gruu_str(pj_pool_t* pool) const;
       std::string pub_gruu_quoted_string(pj_pool_t* pool) const;
+
+      // Binding copy function.  Needs to preserve _address_of_record so that
+      // freeing of the source doesn't release the storage pointed to by the
+      // latter.
+      static void copy_binding(Binding* dest, Binding* src)
```
steelyvoid: Rob has confirmed that there's no need for _address_of_record to be a pointer, so I'll just change it to a regular std::string and remove the need for all this elaborate copy logic.

AlexHockey: Hurrah!

src/pjutils.cpp
steelyvoid: I would argue that it should be the default, since "max_expires == 0" is generally taken to mean "this is a deregister", which is the reason why I had to change this in the first place (see https://github.com/Metaswitch/sprout/blob/dev/src/icscfsproutlet.cpp#L253).  I would argue that this is the behaviour we should go with, since we have no evidence that bono will do the right thing with a Contactless REGISTER anyway (happy to raise an issue to track this if you want)

AlexHockey: That interpretation seems un-intuitive to me. If I have a REGISTER with no contacts `max_expires` returns the default (300, say). If I add a contact with an expiry of 295, the maximum *decreases* to 295, which doesn't seem right. I'm starting to think that `max_expires` is effectively undefined if there are no contacts, and that this function should output two variables: whether the maximum is defined and, if it is, the value. Thoughts?

steelyvoid: I don't really care, to be honest (all feels a bit "angels on the head of a pin" to me) so I'm happy to go with whatever formulation you want as long as its not too much effort to implement.

include/subscriber_data_manager.h
```diff
@@ -188,6 +198,9 @@ class SubscriberDataManager
     /// corresponding subscription does nothing.
     void remove_subscription(const std::string& to_tag);
 
+    // Test-only function to remove the bindings from an AOR object
+    void test_clear_bindings();
```
steelyvoid: Will do

include/subscriber_data_manager.h
```diff
@@ -210,6 +223,11 @@ class SubscriberDataManager
     // Return the expiry time of the binding or subscription due to expire next.
     int get_next_expires();
 
+    /// Copy all bindings and subscriptions to this AoR
+    ///
+    /// @param source_aor           Source AoR for the copy
+    void copy_aor(SubscriberDataManager::AoR* source_aor);
```
steelyvoid: Yep - done

src/pjutils.cpp
AlexHockey: Fine. In that I think we should have:

* `max_expires` output the expiry as an output param, and return if this value is valid (i.e. were there any contacts). 
* Create a new utility function called `is_deregistration` that calls `max_expires` and checks the return code and the output value. This replaces the existing calls to `max_expires` except in...
* acr.cpp, which needs to do something like `max_expires(reg, -1, result) ? result : -1`.

Sorry to be precise about this, but if utility methods start doing weird things then our code becomes less maintainable.

src/registrar.cpp
```diff
@@ -210,6 +210,75 @@ bool get_private_id(pjsip_rx_data* rdata, std::string& id)
...
+    // to the store.
+    TRC_ERROR("Failed to get AoR binding for %s from store", aor.c_str());
+  }
+  else
+  {
+    // If we don't have any bindings, try the backup AoR and/or store.
+    if (aor_pair->get_current()->bindings().empty())
+    {
+      std::vector<SubscriberDataManager*>::iterator it = backup_sdms.begin();
+      SubscriberDataManager::AoRPair* local_backup_aor = NULL;
```
steelyvoid: Ditto

src/subscriber_data_manager.cpp
```diff
@@ -767,6 +780,28 @@ int SubscriberDataManager::AoR::get_next_expires()
...
+// Copy all bindings and subscriptions to this AoR
+void SubscriberDataManager::AoR::copy_aor(SubscriberDataManager::AoR* source_aor)
+{
+  for (Bindings::const_iterator i = source_aor->bindings().begin();
+       i != source_aor->bindings().end();
+       ++i)
+  {
+    Binding* src = i->second;
+    Binding* dst = get_binding(i->first);
+    Binding::copy_binding(dst, src);
```
steelyvoid: Retired this, having fixed up the Binding object so that it no longer contains a pointer to outside storage (so we can go back to straight copying of the object)

src/pjutils.cpp
AlexHockey: Ah, it actually needs the value from max_expires. Fine, maybe it needs to call max_expires directly and cope with the situation that it returns false due to there being no bindings (and by "cope with" I'm happy with the position that it maintains its existing bugged behavior). TBH as long as max_expires fails when there are no bindings, I don't really care how other bits of the code call it.

steelyvoid: Cool - I'll make it so

steelyvoid: Alex, can you confirm that the latest commit addresses your concern here?  Then I think we're good to go

include/pjutils.h
```diff
@@ -141,7 +141,9 @@ pj_bool_t msg_supports_extension(pjsip_msg* msg, const char* extension);
 
 pj_bool_t is_first_hop(pjsip_msg* msg);
 
-int max_expires(pjsip_msg* msg, int default_expires);
+pj_bool_t get_max_expires(pjsip_msg* msg, int default_expires, int* max_expires);
```
AlexHockey: It's more idiomatic in C++ to pass the past parameter as a reference rather than a pointer (`int& max_expires`)

src/bono.cpp
```diff
@@ -1681,7 +1681,8 @@ static void proxy_process_register_response(pjsip_rx_data* rdata)
         // in 5 minutes (300s).  This should never happens as it means the
         // registrar is misbehaving, but we defensively assume a short expiry
         // time as this is more secure.
-        int max_expires = PJUtils::max_expires(rdata->msg_info.msg, 300);
+        int max_expires;
```
AlexHockey: Does this code need to set `max_expires` to the default in case there are no contacts. Ah I see that `get_max_expires` uses the default as a fallback. That's fine, but I don't think we should use the fallback (unless we forget to check the return code!).

AlexHockey:
I've left a few minor comments in you "Refactor max_expires" commit but after that I'm happy.

src/bono.cpp
steelyvoid: Are you asking for an explicit change here?

include/pjutils.h
steelyvoid: OK.  FWIW, I think thats less maintainable code - you can't tell that max_expires is a returned parameter unless you notice the "&" at the top, whereas "*max_expires" makes it explicit - but I'll make that change.

src/pjutils.cpp
```diff
@@ -630,46 +630,61 @@ pj_bool_t PJUtils::is_first_hop(pjsip_msg* msg)
...
-
 /// Gets the maximum expires value from all contacts in a REGISTER message
 /// (request or response).
-int PJUtils::max_expires(pjsip_msg* msg, int default_expires)
+///
+/// Returns TRUE if the maximum expires value is meaningful (i.e. if the
+/// REGISTER includes Contact headers) and FALSE otherwise.  Set max_expires
+/// to the default value in the latter case to ensure that callers that fail
+/// to check the returncode are at least using a sensible default.
+pj_bool_t PJUtils::get_max_expires(pjsip_msg* msg, int default_expires, int* max_expires)
```
steelyvoid: OK

src/bono.cpp
AlexHockey: I was about to suggest initializing max_expires to 300, but now I realize that will have no effect, so no change needed. Sorry for the thinking out loud.

include/pjutils.h
AlexHockey: FWIW I agree with you, but being idiomatic is important too IMO

src/pjutils.cpp
steelyvoid: Does that apply to the return to "is_deregistration" also?

steelyvoid: Hi Alex

Can you cast an eye over the last commit and check that it addresses your remaining concerns about max_expires? Note that I also found and fixed a bug in the new "inline" model (contactless REGISTERS were registering at Homestead and leaving the registration up even if there were no bindings)

AlexHockey: Looks good to me.

AlexHockey:
Passing back to you to handle the merge.

PR #1553, opened 2016-10-03
[Reviewer: Ellie] [INT-54] Changed bono alias to domain, to allow for routing to ibcf by domain

GAB-MS:
Testing of a Clearwater deployment with an IBCF showed that whenever the IBCF was routed to via domain, rather than IP, the node would not remove its own route header and would hence attempt to route to itself.

@mirw suggested that this could be fixed by changing the alias argument in Bono's init.d configuration. I have changed it from the public IP to ibcf.<zone>, which is the expected IP to be used for a cluster IBCF domain entry.

I have tested this change and verified that it fixes the issue I encountered, without breaking the core Bono functionality. 

Please let me know whether you're happy with this change. Ellie, if you're not the right person for this or if you're too busy to take this right now, please pass it to the appropriate person (or let me know who to pass it to). Thank you

### DIFF
--- a/debian/bono.init.d
+++ b/debian/bono.init.d
@@ -120,6 +120,7 @@ get_settings()
         [ -r /etc/clearwater/user_settings ] && . /etc/clearwater/user_settings
 
         # Work out which features are enabled.
+        IBCF_ENABLED=Y
         if [ -d /etc/clearwater/features.d ]
         then
           for file in $(find /etc/clearwater/features.d -type f)
@@ -137,11 +138,10 @@ get_daemon_args()
         # Get the settings
         get_settings
 
-        if [ $ibcf_enabled = Y ]
+        if [ $IBCF_ENABLED = Y ]
         then
           [ -z "$trusted_peers" ] || ibcf_arg="--ibcf=$trusted_peers"
-          [ -n "$ibcf_domain" ] || ibcf_domain="ibcf.$home_domain"
-          $bono_alias_list = "$bono_alias_list,$ibcf_domain"
+          [ -z "$ibcf_domain" ] || $bono_alias_list = "$bono_alias_list,$ibcf_domain"
         fi
 
         [ -z "$ralf_hostname" ] || ralf_arg="--ralf=$ralf_hostname"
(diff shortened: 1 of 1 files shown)


mirw:
I think the alias parameter should include both the IP address and the domain.  (The alias parameter is comma-separated, I think.)  I also feel a little uneasy about Bono assuming that ibcf.<cluster name> resolves to itself - mightn't you have some bonos that are IBCFs and some not?

GAB-MS:
Is there a better way to approach this? The issue we have is that we would have to manually change this for every IBCF node that we spin up. Would Chef be able to handle this modification automatically for IBCF nodes?

mirw:
Right now, you can't set a bono alias in /etc/clearwater/shared_config or /etc/clearwater/local_config.  Sprout has an alias_list option - I suggest replicating that here - see sprout-base.root/etc/init.d/sprout for an example.

Once you've got that, then yes, you can absolutely script that in Chef.  @eleanor-merry is probably the expert on that.

GAB-MS:
That makes quite a bit more sense. I've made changes to this along similar lines to what is done in Sprout. Could you take a look and see if it was what you expected, Matt?

(Also, sorry, the commits show up as being added 13 hours ago - my dev VM's clock has a habit of going out of sync)

eleanor-merry:
It can't be the same alias_list as Sprout uses - can you add a bono_alias_list instead?

GAB-MS:
Done

GAB-MS:
I've played around with this a bit and changed it around. I've added a bono_alias_list. It now also looks for an ibcf_domain setting in user_settings; if it's there, it will get appended to the alias list.

I've tested this package, this doesn't break anything and makes it easy to add a domain alias for IBCF functionality.

Matt, Eleanor, are you happy with this change?

mirw:
Fine by me.  We should also update https://github.com/Metaswitch/clearwater-readthedocs/blob/master/docs/Clearwater_Configuration_Options_Reference.md.

eleanor-merry:
LGTM as well

PR #1556, opened 2016-10-05
[Reviewer: Ellie] Add an --allow-fallback-ifcs option to enable more wildcard use-cases

rkd-msw:
We don't support wildcard matching yet, but one thing that's quite common is to have wildcard PSIs, where we do a lookup for +1234 and get back a single IFC for +123!*!.

We don't need wildcard matching support to handle this - all we need to do is assume the IFC the HSS gave us is the one we're meant to have, even if the Identity field doesn't match.

This PR adds a command-line option to make that assumption (I haven't assumed it by default, because it could do the wrong thing in some situations, e.g. if we had multiple IFCs).

Not urgent.

### DIFF
--- a/src/hssconnection.cpp
+++ b/src/hssconnection.cpp
@@ -369,6 +369,10 @@ bool decode_homestead_xml(const std::string public_user_identity,
 
         associated_uris.push_back(uri);
         ifcs_map[uri] = ifc;
+        if (public_id == sp->first_node("PublicIdentity"))
+        {
+          ifcs_map[public_user_identity] = ifc;
+        }
 
         if (!found_aliases)
         {
--- a/src/hssconnection.cpp
+++ b/src/hssconnection.cpp
@@ -69,7 +69,8 @@ HSSConnection::HSSConnection(const std::string& server,
                              SNMP::EventAccumulatorTable* homestead_uar_latency_tbl,
                              SNMP::EventAccumulatorTable* homestead_lir_latency_tbl,
                              CommunicationMonitor* comm_monitor,
-                             std::string scscf_uri) :
+                             std::string scscf_uri,
+                             bool fallback_if_no_matching_ifc) :
   _http(new HttpConnection(server,
                            false,
                            resolver,
@@ -82,7 +83,8 @@ HSSConnection::HSSConnection(const std::string& server,
   _sar_latency_tbl(homestead_sar_latency_tbl),
   _uar_latency_tbl(homestead_uar_latency_tbl),
   _lir_latency_tbl(homestead_lir_latency_tbl),
-  _scscf_uri(scscf_uri)
+  _scscf_uri(scscf_uri),
+  _fallback_if_no_matching_ifc(fallback_if_no_matching_ifc)
 {
 }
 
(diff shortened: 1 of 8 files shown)


src/hssconnection.cpp
```diff
@@ -369,6 +372,20 @@ bool decode_homestead_xml(const std::string public_user_identity,
...
+        // where we don't do wildcard matching yet. In this case, our behaviour
+        // is determined by fallback_if_no_matching_ifc:
+        //
+        // - if it's true, we'll use the first IFC given regardless of the
+        //   identity it specifies
+        // - if it's false, we just won't invoke any application servers
+        if (fallback_if_no_matching_ifc &&
+            (public_id == sp->first_node("PublicIdentity")))
+        {
+          ifcs_map[public_user_identity] = ifc;
```
eleanor-merry: Do you need to put the public_user_identity in the associated_uris as well?

src/hssconnection.cpp
```diff
@@ -369,6 +372,20 @@ bool decode_homestead_xml(const std::string public_user_identity,
...
+        // We might get a set of IFCs where there's no Identity node that
+        // matches our URI - the main example is with wildcard identities,
+        // where we don't do wildcard matching yet. In this case, our behaviour
+        // is determined by fallback_if_no_matching_ifc:
+        //
+        // - if it's true, we'll use the first IFC given regardless of the
+        //   identity it specifies
+        // - if it's false, we just won't invoke any application servers
+        if (fallback_if_no_matching_ifc &&
+            (public_id == sp->first_node("PublicIdentity")))
```
eleanor-merry: This check is just so that you only set the value in the map once right? Can you make this clear (I spent a while trying to work out why the first Public Identity was important here).

src/ut/scscf_test.cpp
```diff
@@ -2611,7 +2687,7 @@ TEST_F(SCSCFTest, ISCRetargetWithoutCdiv)
 }
 
 
-TEST_F(SCSCFTest, URINotIncludedInUserData)
+TEST_F(SCSCFTest, DISABLED_URINotIncludedInUserData)
```
rkd-msw: Whoops - it didn't pass (by design) before I added the config option and I forgot to re-enable it.

src/hssconnection.cpp
rkd-msw: Can't hurt. I've had to move the logic around to do this correctly (avoiding the risk of putting it in there twice), so can you re-review?

src/ut/scscf_test.cpp
```diff
@@ -1401,6 +1398,82 @@ TEST_F(SCSCFTest, TestTerminatingTelURI)
...
+                                "      <Extension></Extension>\n"
+                                "    </SPT>\n"
+                                "  </TriggerPoint>\n"
+                                "  <ApplicationServer>\n"
+                                "    <ServerName>sip:1.2.3.4:56789;transport=UDP</ServerName>\n"
+                                "    <DefaultHandling>0</DefaultHandling>\n"
+                                "  </ApplicationServer>\n"
+                                "  </InitialFilterCriteria>\n"
+                                "</ServiceProfile></IMSSubscription>");
+  _hss_connection->allow_fallback_ifcs();
```
eleanor-merry: Worth having a UT where allow_fallback_ifcs is true but the identity does match the public user identity, to check that nothing goes wrong in the mainline case?

eleanor-merry:
:shipit:

PR #1557, opened 2016-10-06
[Reviewer: AJH] Only check for NP data in URI classification when this is more important than domain-based classification

steelyvoid:
This addresses a couple of problems in our handling of terminating requests that contain NP data.  

-  If a terminating request arrives at the I-CSCF with a global telephone SIP URI that contains NP data, we don't convert it to a Tel URI (and don't perform an ENUM lookup) because we are looking for a GLOBAL_PHONE_NUMBER and we get (FINAL_)NP_DATA from classify_uri

-  If a terminating request arrives at the S-CSCF with a global telephone SIP URI that contains NP data, classify_uri gives us (FINAL_)NP_DATA and we reject the request because we don't see NODE_LOCAL_SIP_URI or HOME_DOMAIN_SIP_URI.

Note that there is an alternative fix here, which is to recognize that NP data is effectively orthogonal to other types of classification and to return it as a three valued output param (NONE/NP_DATA/FINAL_NP_DATA) from classify_uri, taking the *NP_DATA values out of the returned enum.  Would that make more sense going forward?

### DIFF
--- a/src/pjutils.cpp
+++ b/src/pjutils.cpp
@@ -2110,26 +2110,15 @@ pjsip_uri* PJUtils::translate_sip_uri_to_tel_uri(const pjsip_sip_uri* sip_uri,
     tel_uri->ext_param.ptr = ext->value.ptr;
   }
 
-  // Copy across any routing number and NPDI parameters to the new URI
-  pjsip_param* rn = pjsip_param_find(&sip_uri->userinfo_param, &STR_RN);
-  bool npdi = (pjsip_param_find(&sip_uri->userinfo_param, &STR_NPDI) != NULL);
-
-  if (rn != NULL)
+  // Copy across any SIP user parameters to the new Tel URI
+  for (pjsip_param* p = sip_uri->userinfo_param.next;
+       (p != NULL) && (p != &sip_uri->userinfo_param);
+       p = p->next)
   {
-    // Add the `rn` parameter.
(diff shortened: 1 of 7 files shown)


AlexHockey:
Looks good. I had one question about whether we should be going further in translate_sip_uri_to_tel URI. 

I agree that number-portability is effectively a separate axis of classification, and what you've suggested would be a better API, but I also don't feel the API in the PR is that bad, so probably isn't worth changing.

src/pjutils.cpp
```diff
@@ -2110,6 +2110,28 @@ pjsip_uri* PJUtils::translate_sip_uri_to_tel_uri(const pjsip_sip_uri* sip_uri,
...
+
+  if (npdi)
+  {
+    // Add the 'npdi' parameter
+    pjsip_param* tel_npdi_param = PJ_POOL_ALLOC_T(pool, pjsip_param);
+    pj_strdup(pool, &tel_npdi_param->name, &STR_NPDI);
+    tel_npdi_param->value.slen = 0;
+    pj_list_insert_after(&tel_uri->other_param, tel_npdi_param);
+  }
+
```
AlexHockey: Just the user params. The URI params only apply to SIP URIs.

steelyvoid: Thats done (and I've beefed up the test to make sure that we check that rn and npdi parameters are carried through the translation).  Can you recheck that its what you were expecting?

AlexHockey: LGTM

PR #1559, opened 2016-10-07
[Reviewer: Matt] Add management cached data APIs

AlexHockey:
Hi Matt, please can you review this change to add an HTTP management API to sprout to allow users to query a subscriber's bindings and subscriptions, and to administratively deregister them.

I've tested this in UT, and also live:

* Try to get the bindings/subscriptions for a subscriber who isn't registered.
* Get the bindings/subscriptions for a subscriber who *is* registered.
* Administratively deregister a subscriber, and check that a dereg request is sent to homestead. 

Graeme is doing a fuller live test.

### DIFF
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -719,3 +719,101 @@ HTTPCode AuthTimeoutTask::handle_response(std::string body)
 
   return success ? HTTP_OK : HTTP_SERVER_ERROR;
 }
+
+//
+// APIS for retrieving cached data.
+//
+
+const char* JSON_BINDINGS = "bindings";
+const char* JSON_SUBSCRIPTIONS = "subscriptions";
+
+void GetCachedDataTask::run()
+{
+  // This interface is read only so reject any non-GETs.
+  if (_req.method() != htp_method_GET)
+  {
+    send_http_reply(HTTP_BADMETHOD);
+    delete this;
+    return;
+  }
+
+  // Extract the IMPU that has been requested.
+  const std::string prefix = "/impu/";
+  std::string full_path = _req.full_path();
+  size_t end_of_impu = full_path.find('/', prefix.length());
+  std::string impu = full_path.substr(prefix.length(), end_of_impu - prefix.length());
+  TRC_DEBUG("Extracted impu %s", impu.c_str());
+
+  SubscriberDataManager::AoRPair* aor_pair = nullptr;
+  if (!sdm_access_common(&aor_pair,
+                         impu,
+                         _cfg->_sdm,
+                         _cfg->_remote_sdms,
+                         nullptr,
+                         trail()))
+  {
+    send_http_reply(HTTP_SERVER_ERROR);
+    delete this;
+    return;
+  }
+
+  std::string content = serialize_data(aor_pair->get_current());
+  _req.add_content(content);
+  send_http_reply(HTTP_OK);
+
+  delete aor_pair; aor_pair = NULL;
+  delete this;
+  return;
+}
+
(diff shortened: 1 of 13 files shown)


src/handlers.cpp
```diff
@@ -719,3 +719,193 @@ HTTPCode AuthTimeoutTask::handle_response(std::string body)
...
+    send_http_reply(HTTP_BADMETHOD);
+    delete this;
+    return;
+  }
+
+  // Extract the IMPU that has been requested. The URL is of the form
+  //
+  //   /impu/<public ID>/<element>
+  //
+  // When <element> is either "bindings" or "subscriptions"
```
mirw: Worth a high-level comment about how this binding/subscription stuff actually works?  (This code is in a superclass to the binding- or subscription-specific subclasses.)

src/handlers.cpp
```diff
@@ -719,3 +719,193 @@ HTTPCode AuthTimeoutTask::handle_response(std::string body)
...
+                         nullptr,
+                         trail()))
+  {
+    send_http_reply(HTTP_SERVER_ERROR);
+    delete this;
+    return;
+  }
+
+  // If there are no bindings we don't have any data for the requested
+  // subscriber so return a 404.
```
mirw: This is because we can't have subscriptions if we don't have bindings.  Worth a comment?

src/handlers.cpp
```diff
@@ -719,3 +719,193 @@ HTTPCode AuthTimeoutTask::handle_response(std::string body)
...
+                                       HSSConnection::DEREG_ADMIN,
+                                       trail(),
+                                       &hss_sc);
+
+  // Work out what status code to return.
+  if (all_bindings_expired)
+  {
+    // All bindings expired successfully, so the status code is determined by
+    // the response from homestead.
+    if ((hss_sc <= 200) && (hss_sc < 300))
```
mirw: Less than or equal to 200, and less than 300?  Same comment applies in several places below - please check!

src/main.cpp
```diff
@@ -2146,32 +2163,56 @@ int main(int argc, char* argv[])
...
+    {
+      http_stack_mgmt->register_handler("^/ping$",
+                                        &ping_handler);
+      http_stack_mgmt->register_handler("^/impu/[^/]+/bindings$",
+                                        &get_bindings_handler);
+      http_stack_mgmt->register_handler("^/impu/[^/]+/subscriptions$",
+                                        &get_subscriptions_handler);
+      http_stack_mgmt->register_handler("^/impu/[^/]+$",
+                                        &delete_impu_handler);
+      http_stack_mgmt->start(&reg_httpthread_with_pjsip);
```
mirw: Should the S-CSCF sproutlet be registering these?

modules/cpp-common
```diff
@@ -1 +1 @@
-Subproject commit 57fa09edc2b195be381433d27434b37c11c8e85a
+Subproject commit ebb8159bd487db401e436ed1d11ae264ba423da4
```
AlexHockey: I've been trolled by the github assignment API, and didn't assign you the homestead, cpp-common, or cpp-common-test PRs.

src/main.cpp
AlexHockey: This matches where we register the other handlers that act on registration/subscription state (e.g. the binding timeout handler). For consistency, are you happy to keep these where they are for the moment?

FWIW I think there's an argument for moving *all* of these somewhere else but it's not clear where: I could see arguments for the registrar, the S-CSCF sproutlet, or even the subscriber data manager.

AlexHockey:
@rkd-msw PR for cached data API

PR #1560, opened 2016-10-11
[Reviewer: Graeme] Don't bind a TCP socket unless running as an S-CSCF

AlexHockey:
As discussed, only initialize the HTTP stack and bind a TCP socket if running as an S-CSCF. I've tested live that sprout still listens on port 9886, and bono does not.


AlexHockey:
Actually the right fix is probably to move the bind call closer to where we start the stack. I'll do this tomorrow.

Issue #1451, opened 2016-05-31
Sprout does not honour integrity protection when deployed without an I-CSCF


AlexHockey:
#### Symptoms

Clearwater-sip-stress scenarios fail when running against a deployment that does not use an S-CSCF. They fail at the re-registration stage. 
#### Impact
- Calls made by the stress package fail.
- Unecessary re-authentication flows. 
- Tests made by 3rd party verification tools may fail.
#### Release and environment

release-98
#### Steps to reproduce

Set up a deployment with latest chef and release-98 packages (or alternatively, latest chef and packages, but using `--scscf-only`). Set up SIP stress. Monitor the SIPp output and wait 10 mins. The re-registration flows will all fail.

AlexHockey:
The failures are caused because this REGISTER is challenged with a 401:

```
REGISTER sip:2010000000@ajh2.cw-ngv.com SIP/2.0
Via: SIP/2.0/TCP 10.9.182.172:50458;rport;branch=z9hG4bKPjPlxMhebbrzfwOCNWezI1Ym6jefM2-3E2
Path: <sip:byN1DsIln/@10.9.182.172:5058;transport=TCP;lr;ob>
Via: SIP/2.0/TCP 10.81.150.198:5060;rport=34716;received=54.226.86.133;branch=z9hG4bK-14553-1-11-2010000000-4.000000
Max-Forwards: 70
From: <sip:2010000000@ajh2.cw-ngv.com>;tag=14553SIPpTag001
To: <sip:2010000000@ajh2.cw-ngv.com>
Call-ID: 2010000000///1-14553@10.81.150.198
CSeq: 5 REGISTER
User-Agent: Accession 4.0.0.0
Supported: outbound, path
Contact: <sip:2010000000@54.226.86.133:34716;transport=TCP;ob>;+sip.ice;reg-id=1;+sip.instance="<urn:uuid:00000000-0000-0000-0000-000000000001>"
Contact: <sip:2010000000@10.81.150.198:5060;transport=TCP;ob>;expires=0;+sip.ice;reg-id=1;+sip.instance="<urn:uuid:00000000-0000-0000-0000-000000000001>"
Expires: 3600
Allow: PRACK, INVITE, ACK, BYE, CANCEL, UPDATE, SUBSCRIBE, NOTIFY, REFER, MESSAGE, OPTIONS
[...]

PR #1565, opened 2016-10-17
[Reviewer: AJH] Correctly match next routing header on integrity protected messages

richardwhiuk:
This fixes issue #1451.

Sprout doesn't honour the integrity protected header when acting as an S-CSCF only as the next routing header is `sip:scscf.sprout.rjw2.cw-ngv.com:5054;transport=TCP;lr;orig` which doesn't match `sip:scscf.sprout.rjw2.cw-ngv.com:5054;transport=TCP`.

The brunt of this fix is to add some functions to PJUtils which allow two URIs to be compared in a way that we care about (that's a little nebulous, but I think fine).

This fix ended up being quite involved.

We now store the parsed S-CSCF URI on the stack. This is useful for several reasons, including the stack not needing to parse it every time it's needed - which takes memory out of the pool. This renames the existing string S-CSCF URI from `stack_data.scscf_uri` is now `stack_data.scscf_uri_str`.

The existing code has bono loading the registrar and subscription modules (since https://github.com/Metaswitch/sprout/pull/1508 ), which doesn't work, because the S-CSCF URI doesn't form a valid service route address, and thus we crash out with the following logs:

```
17-10-2016 15:59:29.980 UTC Error registrar.cpp:1197: Unable to set up Service-Route header for the registrar from sip:scscf.:0;transport=TCP
17-10-2016 15:59:29.980 UTC Error main.cpp:2074: Failed to enable S-CSCF registrar
```

Frankly, I can't see how this worked prior to this change, but I'm confident we shouldn't be enabling these modules.

I've also fixed up the UTs to use the new stack_data parameters. In doing so, I noticed we were passing in a host_mapping parameter, which was entirely unused, so I've removed that. This has been the case since https://github.com/Metaswitch/sprout/commit/79bd5f7bf8987c72a577ecc04d96e99210a1b0c5.

I've added a UT to cover integrity protection in general, and added a specific case to cover this flow.

I've run the UTs, and run under stress, with a minimal deployment, with only 5 calls, to check that we don't see the 401 on the integrity protected REGISTER.


AlexHockey:
It looks like the majority of the new code in this PR is adding custom URI comparison code. [RFC 3261](https://www.ietf.org/rfc/rfc3261.txt) defines how to compare SIP URIs and I think that if we followed its rules here the URIs would match. So I think it would be better to just call `pjsip_uri_cmp` instead of our own hand rolled method. Do you agree?

richardwhiuk:
I previously looked at this, and was under the impression that it rejected URIs which we needed to be considered valid, but having reassessed this, looking at both the code and RFC 3261, I agree we can use pjsip_uri_compare. I've updated the code to use that instead.

AlexHockey:
Reviewed 21 of 24 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[include/pjutils.h, line 311 at r2](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-KUN-3cW2eo3UcSrzPtL:-KUN-3cXv-fpAxA7doE7:bedsr60) ([raw file](https://github.com/metaswitch/sprout/blob/ac569650cb0c57ee676820f9985ff804b81871f8/include/pjutils.h#L311)):*
> ```objc
> // request URI if there's no route headers), and it's context.
> pjsip_uri* get_next_routing_uri(const pjsip_msg* msg,
>                                 pjsip_uri_context_e* context);
> ```

Might be worth mentioning that the return value is only valid for as long as the underlying message is. 

---

*[src/authentication.cpp, line 774 at r2](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-KUMzWpM6s0ReUefTSus:-KUMzWpM6s0ReUefTSut:b-9nf6o5) ([raw file](https://github.com/metaswitch/sprout/blob/ac569650cb0c57ee676820f9985ff804b81871f8/src/authentication.cpp#L774)):*
> ```cpp
>         pjsip_uri* next_routing_uri = PJUtils::get_next_routing_uri(rdata->msg_info.msg, &context);
> 
>         TRC_DEBUG("Routing headers check for S-CSCF chain: %s == %.*s ?",
> ```

I didn't really understand the wording of this log. It might be clearer to say something like "Needs authentication as next routing URI (%s) does not match this S-CSCF (%s)" and put it in an else branch to the if test below. 

---

*[src/main.cpp, line 2040 at r2](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-KUN-PjcJRLjCXdX7hY8:-KUN-PjcJRLjCXdX7hY9:bruzy44) ([raw file](https://github.com/metaswitch/sprout/blob/d6edcbb86fefa9b4979a44f298158f3c8b97fe5d/src/main.cpp#L2040)):*
> ```cpp
>   }
> 
>   if (opt.auth_enabled)
> ```

Is `auth_enabled` used any more? 

---

*[src/registrar.cpp, line 1177 at r2](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-KUN0IXPpOroGH2FvJcE:-KUN0IXPpOroGH2FvJcF:b-o1wmyl) ([raw file](https://github.com/metaswitch/sprout/blob/ac569650cb0c57ee676820f9985ff804b81871f8/src/registrar.cpp#L1177)):*
> ```cpp
>   if (service_route_uri != NULL)
>   {
>     service_route_uri->lr_param = 1;
> ```

You're mutating the configured S-CSCF URI here which isn't right.

---

*[src/registrar.cpp, line 1184 at r2](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-KUN0Q8p6JXqunOiYQNZ:-KUN0Q8p6JXqunOiYQN_:b-89a4l7) ([raw file](https://github.com/metaswitch/sprout/blob/ac569650cb0c57ee676820f9985ff804b81871f8/src/registrar.cpp#L1184)):*
> ```cpp
>     pj_strdup(stack_data.pool, &orig_param->name, &STR_ORIG);
>     pj_strdup2(stack_data.pool, &orig_param->value, "");
>     pj_list_insert_after(&service_route_uri->other_param, orig_param);
> ```

Ditto

---

*[src/stack.cpp, line 727 at r2](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-KUN1VDBbu8nZtDlyUwc:-KUN1VDBbu8nZtDlyUwd:bmvtcxn) ([raw file](https://github.com/metaswitch/sprout/blob/ac569650cb0c57ee676820f9985ff804b81871f8/src/stack.cpp#L727)):*
> ```cpp
> 
>   stack_data.scscf_uri = (pjsip_sip_uri*)PJUtils::uri_from_string(scscf_uri,
>                                                                   stack_data.pool);
> ```

What happens if this parse fails. Should initialization fail in some way?

---

*[src/stack.cpp, line 792 at r2](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-KUN1rGbymIwMgj9AQm8:-KUN1rGcVVRy1SV2s2W8:b-mjdex7) ([raw file](https://github.com/metaswitch/sprout/blob/ac569650cb0c57ee676820f9985ff804b81871f8/src/stack.cpp#L792)):*
> ```cpp
>   else
>   {
>     stack_data.scscf_uri = NULL;
> ```

I don't understand why this needs to be nulled out. 

---

*[src/ut/authentication_test.cpp, line 605 at r2](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-KUN2HGctTpk_qPuQEQ6:-KUN2HGctTpk_qPuQEQ7:b-xjrana) ([raw file](https://github.com/metaswitch/sprout/blob/ac569650cb0c57ee676820f9985ff804b81871f8/src/ut/authentication_test.cpp#L605)):*
> ```cpp
> 
>   EXPECT_EQ(0,((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_AUTHENTICATION_STATS_TABLES.sip_digest_auth_tbl)->_attempts);
>   EXPECT_EQ(0,((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_AUTHENTICATION_STATS_TABLES.ims_aka_auth_tbl)->_attempts);
> ```

There should be a test that the authentication module *does* handle the request the URIs *don't* match.

---

*[src/ut/registrar_test.cpp, line 681 at r2](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-KUN2w1aGVFS6N8qbCcM:-KUN2w1aGVFS6N8qbCcN:bbjime9) ([raw file](https://github.com/metaswitch/sprout/blob/ac569650cb0c57ee676820f9985ff804b81871f8/src/ut/registrar_test.cpp#L681)):*
> ```cpp
>   EXPECT_EQ(msg._path, get_headers(out, "Path"));
>   EXPECT_EQ("P-Associated-URI: <sip:6505550231@homedomain>", get_headers(out, "P-Associated-URI"));
>   EXPECT_EQ("Service-Route: <sip:all.the.sprout.nodes:5058;transport=TCP;lr;orig>", get_headers(out, "Service-Route"));
> ```

Why do you need this test? You didn't change the registrar much did you?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/sprout/1565)*

AlexHockey:
Generally looks good. A few comments inline, but only one functional issue.

---

Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/sprout/1565#-:-KUN421u88cetzfuxLHQ:brhp9m2)*

PR #1572, opened 2016-10-20
Reviewer RKD: Use syslog for analytics logs

graemerobertson:
Use syslog to log analytics logs.

This also contains the Sprout changes that match https://github.com/Metaswitch/cpp-common/pull/574.

### DIFF
--- a/sprout-base.root/etc/logrotate.d/sproutanalytics
+++ b/sprout-base.root/etc/logrotate.d/sproutanalytics
@@ -1,11 +1,16 @@
 /var/log/sprout/analytics.log {
   su sprout sprout
   daily
+  # Keep 36500 log files (i.e. 100 years worth). The log cleanup scripts will
+  # clean up this directory if the files become too big.
   rotate 36500
+  # Add a date and timestamp to the rotated log file.
   dateext
   dateformat .%Y-%m-%d-%s
   extension .log
   missingok
+  # After rotating, rsyslog still has the rotated file open; reload it so that
+  # it starts writing to the new file.
   postrotate
     reload rsyslog >/dev/null 2>&1 || true
   endscript
--- a/sprout-base.root/etc/logrotate.d/sproutanalytics
+++ b/sprout-base.root/etc/logrotate.d/sproutanalytics
@@ -1,0 +1,12 @@
+/var/log/sprout/analytics.log {
+  su sprout sprout
+  daily
+  rotate 36500
+  dateext
+  dateformat .%Y-%m-%d-%s
+  extension .log
+  missingok
+  postrotate
+    reload rsyslog >/dev/null 2>&1 || true
+  endscript
+}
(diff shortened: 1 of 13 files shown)


sprout-base.root/etc/logrotate.d/sproutanalytics
```diff
@@ -0,0 +1,12 @@
+/var/log/sprout/analytics.log {
+  su sprout sprout
+  daily
+  rotate 36500
+  dateext
+  dateformat .%Y-%m-%d-%s
+  extension .log
+  missingok
+  postrotate
+    reload rsyslog >/dev/null 2>&1 || true
```
rkd-msw: Worth an explanatory comment?

rkd-msw: In fact there are a couple of such comments that would be useful - e.g. what the 36500 is.

sprout-base.root/usr/share/clearwater/infrastructure/scripts/create-analytics-syslog-config
```diff
@@ -0,0 +1,72 @@
...
+# as those licenses appear in the file LICENSE-OPENSSL.
+
+# This file creates a syslog config file for Sprout's analytics logs.
+
+syslog_conf_file=/etc/rsyslog.d/30-sproutanalytics.conf
+temp_file=$(mktemp sproutanalytics.syslog.XXXXXXXX)
+
+. /etc/clearwater/config
+if [ -n "$remote_audit_logging_server" ]; then
+  remote_syslog_audit_server_str=":msg, startswith, \"<analytics> Registration:\" @@${remote_audit_logging_server}"
```
rkd-msw: Worth a comment to explain what this is doing?

src/analyticslogger.cpp
```diff
@@ -68,12 +69,12 @@ void AnalyticsLogger::registration(const std::string& aor,
...
   char buf[BUFFER_SIZE];
   snprintf(buf, sizeof(buf),
-           "Registration: USER_URI=%s BINDING_ID=%s CONTACT_URI=%s EXPIRES=%d\n",
+           "<analytics> Registration: USER_URI=%s BINDING_ID=%s CONTACT_URI=%s EXPIRES=%d\n",
            aor.c_str(),
            binding_id.c_str(),
            contact.c_str(),
            expires);
-  _logger->write(buf);
+  syslog(LOG_INFO, "%s", buf);
```
rkd-msw: Worth factoring out this call into a separate function that calls `syslog(LOG_INFO, "<analytics> %s", buf);`, so we can easily change the level or the prefix in one place in future?

rkd-msw:
Looks good.

src/ut/subscription_test.cpp
```diff
@@ -817,7 +817,7 @@ class SubscriptionTestMockStore : public SipTest
     _chronos_connection = new FakeChronosConnection();
     _local_data_store = new MockStore();
     _sdm = new SubscriberDataManager((Store*)_local_data_store, _chronos_connection, true);
-    _analytics = new AnalyticsLogger(&PrintingTestLogger::DEFAULT);
+    _analytics = new AnalyticsLogger();
```
rkd-msw: It feels like they should be more self-contained. Probably an existing problem, though, so shouldn't hold up merging this.

PR #1573, opened 2016-10-20
[Reviewer: Ellie] Add chronos-sprout-callback-uri option

sebrexmetaswitch:
rather than just using the configured sprout hostname for chronos callbacks.
This allows us to configure chronos to always call back to sprouts in the local
site

Testing done:
 - live tested

### DIFF
--- a/include/cfgoptions.h
+++ b/include/cfgoptions.h
@@ -151,7 +151,7 @@ struct options
   bool                                 sas_signaling_if;
   bool                                 disable_tcp_switch;
   std::string                          chronos_hostname;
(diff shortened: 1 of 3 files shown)


include/cfgoptions.h
```diff
@@ -151,6 +151,7 @@ struct options
   bool                                 sas_signaling_if;
   bool                                 disable_tcp_switch;
   std::string                          chronos_hostname;
+  std::string                          chronos_sprout_callback_uri;
```
eleanor-merry: Can this be sprout_chronos not chronos_sprout?

PR #1574, opened 2016-10-20
[Reviewer: Richard] Add script to determine registered subscribers and bindings from registration logs

MatMeredith:
Add script to parse Sprout registration logs to determine the number of registered subscribers / bindings at a specified time (and to generate lists of all such subscribers / bindings).

The purpose of this script is to validate licensing HWMs generated on the Stats Aggregator.   The script is intended to be run on an arbitrary remote syslog server (provided by the customer).   We are separately providing configuration for both Sprout and the syslog server that will stream these logs to the server and store them in the format expected by this script.

Note the script currently requires the timestamp to verify to be entered in seconds since the epoch.   This isn't very user-friendly and we might want to do better in future, but:
  - it maximizes portability by not requiring any non-standard libraries such as dateutils
  - the scripts on the stats aggregator that query HWMs output the timestamp of the HWM in seconds since the epoch (as well as in human readable form) so  this isn't as painful as it might otherwise be.

I guess there is also a question as to whether Sprout is the right place for this script to live -- I don't know if there is anywhere more generic for such "tools" -- but it is specific to Sprout logs so this doesn't seem daft...


sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+    buckets = max(1, int(args.num_subs / 100000))
+
+    # Check that the temporary directory we are going to use for our interim log
+    # files does not already exist.  If it does, prompt the user to delete it
+    # and re-run, otherwise create it.
+    #
+    # WIBNI: Would be better to prompt the user and ask for confirmation that it
+    # is OK for us to overwrite the existing directory.
+    if os.path.exists(TEMP_DIR):
+        print "Temporary directory " + TEMP_DIR + " already exists -- please " \
```
richardwhiuk: Shouldn't this use logging.error?

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+    # Make subdirectories for each of the buckets we're going to use.
+    for bucket in range(buckets):
+        os.makedirs(bucket_dir(bucket))
+
+    log_filter = os.path.join(args.logfiles_directory, 'registrations*')
+    raw_file_list = glob.glob(log_filter)
+
+    if (len(raw_file_list) == 0):
+        print 'Error: No log files found matching %s' % log_filter
+        sys.exit()
```
richardwhiuk: I think either this is a potential fine condition, in which case we should handle it, or this is an error in which case we should return none zero.

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+
+    # Make subdirectories for each of the buckets we're going to use.
+    for bucket in range(buckets):
+        os.makedirs(bucket_dir(bucket))
+
+    log_filter = os.path.join(args.logfiles_directory, 'registrations*')
+    raw_file_list = glob.glob(log_filter)
+
+    if (len(raw_file_list) == 0):
+        print 'Error: No log files found matching %s' % log_filter
```
richardwhiuk: Again logging.error here.

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+                                  'Last update',
+                                  'Expires']
+            registered_binding_writer = csv.DictWriter(
+                                            registered_bindings_fh,
+                                            fieldnames=binding_fieldnames)
+            registered_binding_writer.writeheader()
+
+            # Now run through the buckets of logs we've created one by one and
+            # build up counts and lists of registered subscribers and bindings.
+            print "Processing %d buckets of logs:" % buckets
```
richardwhiuk: logging.debug surely

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+            for bucket in range(buckets):
+                print "  -- processing bucket %d..." % bucket
```
richardwhiuk: logging.debug surely

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
richardwhiuk: If not, it should probably go to stderr instead of stdout.

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+
+    match_result = re.match(LOG_REGEX, line)
+    if match_result is None:
+        logging.error("Cannot parse line: %s" % line)
+        raise Exception()
+
+    return {"timestamp_str": match_result.group("timestamp_str"),
+            "impu": match_result.group("impu"),
+            "binding": match_result.group("binding"),
+            "expires_ms": int(match_result.group("expires")) * 1000}
```
richardwhiuk: Or even a NamedTuple

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+# We can't be certain that logs appear in log files on the syslog server in
+# strict chronological order.  To determine whether a log file might contain
+# logs that are relevant to the time we are interested in, we look at the
+# timestamps of the first and last logs in the file, and parse the file if they
+# are witin TIME_MARGIN_MS (10 minutes) of being applicable.
+TIME_MARGIN_MS = 10 * 60 * 1000
+
+EPOCH_DT = datetime.datetime.utcfromtimestamp(0)
+
+
```
MatMeredith: Required for PEP8 compliance? (in fact I've bent the rules by NOT having two blank lines in front of def bucket_dir...

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+                # to log as the error will already have been logged by the
+                # decode function.
+                pass
+
+    for bucket in file_handles:
+        file_handles[bucket].close()
+
+
+def process_bucket(bucket_dir, subscriber_writer, binding_writer,
+                   time_of_interest_ms):
```
MatMeredith: Yes -- so would I.   I'm always surprised that PEP8 doesn't advise or mandate this...  https://www.python.org/dev/peps/pep-0008/#id17

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+                try:
+                    log_time_ms = string_to_ms(line_info["timestamp_str"])
+                except:
+                    # Can't decode the timestamp in this log.  Skip it.
+                    logging.error("Cannot decode timestamp in log %s from "
+                                  "file %s" % (line, fname))
+                    continue
+
+                log_impu = line_info["impu"]
+                log_binding = line_info["binding"]
```
MatMeredith: Just readability (and a little performance?).   Is that not a good reason?

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+                log_expires_ms = int(line_info["expires_ms"])
```
MatMeredith: Performance.   When we are bucketing logs we need to parse all of the lines and extract the IMPUs to determine which bucket to put them in.   At this stage though we don't need to decode all of the timestamp strings.   strptime is relatively expensive so I only do this when I need to.   Returning an object with a timestamp method which converts the timestamp when needed would be cleaner.

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+                # If this is a register that expires before the time we are
+                # interested in then treat it like a de-register.   (We don't
+                # want to just ignore it as it might cancel out an earlier
+                # register with a longer expiry time).
+                if ((log_expires_ms > 0) and ((log_time_ms + log_expires_ms) <
+                                              time_of_interest_ms)):
+                    log_expires_ms = 0
+
+                impu_record[log_binding] = {"updated_ms": log_time_ms,
+                                            "expires_ms": log_expires_ms}
```
MatMeredith: Agree Python object might be cleaner as above.   Any idea what performance / occupancy overhead this would add though?

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+            # Check whether the events in this file are too late to be of
+            # interest.  See comment on TIME_MARGIN_MS.
+            if ((first_line_ms - TIME_MARGIN_MS) > time_of_interest_ms):
+                logging.info("Skipping file %s (events too late)" % fname)
+                return False
+
+        except:
+            # Failed to parse the first line.  There may still be other logs in
+            # the file that are of interest...
+            logging.error("Failed to parse first line of file %s" % fname)
```
MatMeredith: I don't think I need to look for other first lines...  We just include the whole file, and will later on use any logs in the file that we can successfully decode?

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+    parser.add_argument('--subs',
+        metavar='N',
+        dest='num_subs',
+        action='store',
+        default=5000000,
+        type=int,
+        help='the approximate number of subscribers in the deployment. '
+             'Defaults to 5000000.  This value is only used to optimize '
+             'performance, but setting it too low may cause the script to '
+             'fail due to lack of resources.')
```
MatMeredith: Not sure that I understand...   "reject calls"?   This only impacts the performance of this verify_licensing_hwm script?!   I discussed with MIRW whether we should e.g. do something clever to dynamically tune the number of buckets but we both felt that this was overkill.   I don't except anyone to ever need to touch this unless they are running with a very large deployment (e.g. in excess of 50 million subs)

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+    # subscriber get assigned to the same bucket (meaning that the bucket can be
+    # processed in isolation) whilst reducing the number of subscribers in each
+    # bucket by a factor of the number of buckets.
+    # Calculate the number of buckets by dividing the expected number of
+    # subscribers by 10.  By default we assume 5 million subs (50 buckets).
+    buckets = max(1, int(args.num_subs / 100000))
+
+    # Check that the temporary directory we are going to use for our interim log
+    # files does not already exist.  If it does, prompt the user to delete it
+    # and re-run, otherwise create it.
```
MatMeredith: Looks cool.   Didn't know about that ;-)

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
MatMeredith: Certainly don't want to just log it to file.   I want console output to be sufficient for the user to know what happened at a high level.   I guess I could write to stderr...

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
MatMeredith: No.   This script can potentially takes minutes to run.   This way they get output that looks like:
Processing 10 buckets of logs...
-- Processing bucket 1...
-- Processing bucket 2...
etc.   
Ideally we'd have an "X% complete output" but there's a limit to how much effort is worth spending here.

MatMeredith:
@richardwhiuk I've applied most of your markups except where I've explicitly commented.   The only exception is changing e.g. decode_line to return a class rather than a dict where I've just run out of time.   Can you briefly cast your eye over my changes please to check that you are happy and confirm if you are OK for me to merge?

MatMeredith:
@richardwhiuk Ping...  Are you happy for me to merge this?

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
```diff
@@ -0,0 +1,432 @@
...
+    registered_subs = 0
+    registered_bindings = 0
+
+    for impu, impu_record in binding_data.iteritems():
+        # Track the number of bindings registered for this IMPU
+        registered_impu_bindings = 0
+        for binding, binding_record in impu_record.iteritems():
+            if binding_record["expires_ms"] > 0:
+                registered_impu_bindings = registered_impu_bindings + 1
+                registered_bindings = registered_bindings + 1
```
richardwhiuk: Yes, Python doesn't support ++, but does support +=.

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
richardwhiuk: If we are worried particularly about performance, we should be writing this in something which isn't Python. I think the call will be essentially the same performance to the level that we would care about.

sprout-base.root/usr/share/clearwater/bin/verify_licensing_hwm.py
richardwhiuk: If we fail due to lack of resources, presumably we've exhausted something on the host (memory?). If this is being run on a Sprout node, that might affect the Sprout process?

richardwhiuk:
Some feedback above, but I'm happy for you to merge this.

PR #1576, opened 2016-10-21
[Reviewer: RKD] Recognize AKAv2 when choosing stats table for auth

richardwhiuk:
Rob,

As discussed, here's a fix to our AKAv2 support.

I haven't add any new UTs - I had a look, but the current tests fall through to looking at the challenge instead, so it still works. I don't think there's anything sensible to do here. I could send in an authentication request where we've replied invalidly to a challenge (i.e we were challenged with DIGEST, but respond with AKAv2), but that feels wrong. As such, I suspect this fix has no impact in a real deployment.

### DIFF
--- a/src/authentication.cpp
+++ b/src/authentication.cpp
@@ -917,12 +917,12 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
     }
     else
     {
-      if (!pj_strcmp2(&credentials->algorithm, "MD5"))
+      if (!pj_strcmp(&credentials->algorithm, &STR_MD5))
       {
         auth_stats_table = auth_stats_tables->sip_digest_auth_tbl;
       }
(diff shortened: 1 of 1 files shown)


src/authentication.cpp
```diff
@@ -921,7 +921,8 @@ pj_bool_t authenticate_rx_request(pjsip_rx_data* rdata)
       {
         auth_stats_table = auth_stats_tables->sip_digest_auth_tbl;
       }
-      else if (!pj_strcmp2(&credentials->algorithm, "AKAv1-MD5"))
+      else if ((!pj_strcmp2(&credentials->algorithm, "AKAv1-MD5")) ||
```
rkd-msw: Can we use the constants from https://github.com/Metaswitch/sprout/blob/dev/include/constants.h#L79 ? I think that's why I missed this first time - I updated all the places that used STR_AKAV1_MD5.

PR #1579, opened 2016-10-24
[Reviewer: RKD] Stop sprout hanging when quiescing

### LINKED ISSUE #1530: Sprout hanging while quiescing when updating
#### Symptoms

When running the commands detailed in the Chef Install section of the Upgrading a Clearwater Deployment page on clearwater.readthedocs.io to update sprout, the command occasionally seems to hang on  "\* Quiescing Sprout SIP Router sprout". It had not progressed beyond that point for around 30mins. Manually stopping the sprout service causes the command to continue running as normal. 
#### Impact

Prevents the upgrade process from completing.
#### Release and environment

A Chef deployment on Amazon EC2 of the latest commit (https://github.com/Metaswitch/sprout/commit/2705ee68b1dd6ad6be9e6a510d48337b067d4ae6) of sprout.

Diagnostics:
[20160905154619Z.ec2-54-87-239-228.compute-1.amazonaws.com.gather_diags.tar.gz](https://github.com/Metaswitch/sprout/files/455336/20160905154619Z.ec2-54-87-239-228.compute-1.amazonaws.com.gather_diags.tar.gz)
#### Steps to reproduce

The problem seems to be very intermittent. It might be possible to reproduce it by repeatedly updating sprout.

sathiyan-sivathas:
Fixes #1530.

I've not been able to work out the actual reason for sprout hanging because this has been really hard to reproduce but this fix will stop us from quiescing indefinitely.

I've added a 5 minute time limit for quiescing, after which we will kill sprout.  I've also left in a debug log for the number of transactions we're quiescing because it seemed like a useful log to have.


sprout-base.root/etc/init.d/sprout
```diff
@@ -421,7 +421,7 @@ do_start_quiesce() {
 do_quiesce() {
         # The timeout after forever is irrelevant - start-stop-daemon requires one but it doesn't
```
rkd-msw: Comment no longer applies? (We should also explain the 'kill').

sathiyan-sivathas:
A few changes since you last saw this.  I've had to move a few class members to public access so I could view them from main.cpp, I assume that's OK?

src/main.cpp
```diff
@@ -2233,6 +2233,16 @@ int main(int argc, char* argv[])
 
   // Wait here until the quit semaphore is signaled.
   sem_wait(&term_sem);
+
+  // If we received a TERM while we were quiescing, log that as an error.  We
+  // should only receive a TERM when quiescing if something has gone wrong and
+  // we have taken longer than 5 minutes to quiesce.
+  if ((quiescing_mgr->_state == QuiescingManager::STATE_QUIESCING_FLOWS) ||
```
rkd-msw: I might define a new method - `quiescing_mgr->is_still_quiescing()` or something. That keeps the details of the quiescing state machine private (which is cleaner) and lets you just ask the QuiescingManager about the thing you really want to know.

PR #1580, opened 2016-10-28
[Reviewer: RKD] Add more quiescing logging

sathiyan-sivathas:
Add more quiescing logs so that we know exactly where we hang the next time we hit #1530 

I've added status logs throughout to make sure we see them even with log_level=2.  I think this covers the whole quiescing flow.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1212,7 +1212,7 @@ class QuiesceCompleteHandler : public QuiesceCompletionInterface
 public:
   void quiesce_complete()
   {
-    TRC_STATUS("Quiesce complete - send TERM");
+    TRC_STATUS("Quiesce complete");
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -1212,6 +1212,7 @@ class QuiesceCompleteHandler : public QuiesceCompletionInterface
 public:
   void quiesce_complete()
   {
+    TRC_STATUS("Quiesce complete - send TERM");
```
rkd-msw: We're not sending a SIGTERM here, we're unblocking to the semaphore that blocks the termination processing in main.cpp (http://www.csc.villanova.edu/~mdamian/threads/posixsem.html)

rkd-msw:
LGTM.

PR #1582, opened 2016-10-29
Reviewer RJW2: Analytics logs fixes

graemerobertson:
This is the `nocreate` fix we discussed on Friday, and also the permissions commands in the postinst that I claimed I'd made but apparently hadn't actually done...


graemerobertson:
I'm going to merge this for the following reasons.
* This function is currently broken in master.
* Rich has already agreed with these changes (just not explicitly on this PR).
* I forgot Rich wasn't here this week.

PR #1583, opened 2016-10-31
[Reviewer Alex] BGCF Routing by phone number

sebrexmetaswitch:
Route phone number URIs with no routing number parameter on the phone number.
Previously, they were always routed to the wildcard domain route. Now, try to route on the phone number. If no matching routes found, fall back to the wildcard domain as before.

**Testing done:**
 - unit tests

### DIFF
--- a/src/ut/bgcf_test.cpp
+++ b/src/ut/bgcf_test.cpp
@@ -345,7 +345,7 @@ void BGCFTest::doFailureFlow(BGCFMessage& msg, int st_code)
   inject_msg(msg.get_request());
   ASSERT_EQ(2, txdata_count());
 
-  // We get the 100 Trying from the SproutletProxy, so check this first  
+  // We get the 100 Trying from the SproutletProxy, so check this first
   pjsip_msg* out = current_txdata()->msg;
   RespMatcher(100).matches(out);
   free_txdata();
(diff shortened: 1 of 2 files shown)


src/ut/bgcf_test.cpp
```diff
@@ -356,8 +356,49 @@ void BGCFTest::doFailureFlow(BGCFMessage& msg, int st_code)
...
+TEST_F(BGCFTest, TestSipPhoneNumberUriNotMatched)
+{
+  add_host_mapping("ut.cw-ngv.com", "10.9.8.7");
+  SCOPED_TRACE("");
+  BGCFMessage msg;
+  msg._requri = "sip:16505551234@notamatchingdomain;user=phone";
+  list<HeaderMatcher> hdrs;
+  hdrs.push_back(HeaderMatcher("Route", ".*10.0.0.2:5060.*"));
+  doSuccessfulFlow(msg, testing::MatchesRegex("sip:16505551234@notamatchingdomain;user=phone"), hdrs);
+}
```
AlexHockey: It would also be good to test:

* tel URI doesn't match and picks up the default route (in case of a future bugs where the only SIP URIs can do domain routing).
* tel URI doesn't match and there is no default route - gets rejected.

AlexHockey:
Code looks good. I've suggested a couple of extra simple testcases but hopefully they are easy to implement.

sebrexmetaswitch:
As discussed, those test cases are already covered by ```BGCFTest.TestSimpleTelURIUnmatched``` and ```SCSCFTest.TestEnumLocalTelURI```

PR #1584, opened 2016-11-04
Reviewer MGM: Analytics logs minor fixes

graemerobertson:
I was printing the contents of the log and the timestamp the wrong way around. This also exposed the fact that my logs had an (unnecessary) new line in them, which was now being printed out (as #012).

I've also added a random comment.


MatMeredith:
LGTM

PR #1585, opened 2016-11-04
[Reviewer: Ellie] SDM Registration Analytics Logging

pdavies:
A few things -- fighting with circular dependencies, a couple of false starts, and an aborted attempt at writing unit tests for the logging -- meant I didn't have time to get this done as thoroughly as I would have liked, but it meets the task requirements.

### What I have:

- Registration logging moved from registrar.cpp to the Subscriber Data Manager. Removed and shortened bindings are logged before calling out to storage, and created and refreshed bindings are logged afterwards.

- Classification of bindings (CREATED / REFRESHED / EXPIRED, etc) refactored out into classify_bindings(), which does all the hard work once and is used by the logging functions.

- Registration logs being emitted correctly during unit tests.

### What I don't have:

- Unit testing of the logging itself (although that wasn't there before, either). Figuring out how to pipe dummy data into the unit tests to get the right kind of logging to come out turned out to be a huge job in terms of getting my head around new stuff, and I just didn't have time.

- The existing logic for sending NOTIFYs using the results of classify_bindings(). The stuff this spits out is exactly what send_notifys_for_current_subscriptions() and send_notifys_for_expired_subscriptions() need, so it should be straightforward to move them over to using the ClassifiedBindings, simplifying them substantially.

- Subscription logging moved over from subscription.cpp into subscriber_data_manager.cpp.


pdavies:
make full_test gives an XML file that's not particularly human readable but gives a number of heap memory leak errors. Hopefully the last commit fixes it?

src/subscriber_data_manager.cpp
```diff
@@ -178,14 +180,24 @@ Store::Status SubscriberDataManager::set_aor_data(
...
-  // so that only one piece of code has responsibility for this
+  // so that only one piece of code has responsibility for this. Furthermore,
+  // we want registration logs used for licensing counts to undercount in edge 
+  // cases where a Chronos or memcached call fails and we're in an uncertain
+  // state. Therefore, we log removed or shortened bindings before any such calls,
+  // and we log new or extended bindings afterwards.
+
+  ////////////////////////////////////////////
+  // 1. Expire any old bindings/subscriptions.
+  ////////////////////////////////////////////
```
eleanor-merry: I'm really not a fan of this commenting style...

src/subscriber_data_manager.cpp
```diff
@@ -214,12 +226,30 @@ Store::Status SubscriberDataManager::set_aor_data(
   TRC_DEBUG("Set AoR data for %s, CAS=%ld, expiry = %d",
             aor_id.c_str(), aor_pair->get_current()->_cas, max_expires);
 
-  // Set the chronos timers
-  if (_primary_sdm)
+  ClassifiedBindings classified_bindings;
+  
+  if(_primary_sdm)
```
eleanor-merry: Space between if and ( please

src/subscriber_data_manager.cpp
```diff
@@ -214,12 +226,30 @@ Store::Status SubscriberDataManager::set_aor_data(
...
   {
+    ///////////////////////////////////////
+    // 2. Log removed or shortened bindings
+    ///////////////////////////////////////
+    classify_bindings(aor_id, aor_pair, classified_bindings);
+
+    if (_analytics != NULL)
+    {
+      log_removed_shortened_bindings(classified_bindings);
```
eleanor-merry: Can this be log_removed_or_shortened_bindings instead?

src/subscriber_data_manager.cpp
```diff
@@ -250,13 +297,137 @@ Store::Status SubscriberDataManager::set_aor_data(
...
   return Store::Status::OK;
 }
 
+void SubscriberDataManager::classify_bindings(const std::string& aor_id,
+                                              SubscriberDataManager::AoRPair* aor_pair,
+                                              ClassifiedBindings& classified_bindings)
+{
+  // We should have been given an empty classified_bindings vector, but clear 
+  // it just in case 
+  classified_bindings.clear();
```
eleanor-merry: This would leak memory if this was ever not empty, wouldn't it? Worth having a helper method to delete the vector entries?

src/subscriber_data_manager.cpp
```diff
@@ -250,13 +297,137 @@ Store::Status SubscriberDataManager::set_aor_data(
...
+       ++it)
+  {
+    NotifyUtils::BindingNotifyInformation* classified_binding = *it;
+    
+    if (classified_binding->_contact_event == NotifyUtils::ContactEvent::EXPIRED)
+    {
+      _analytics->registration(classified_binding->_b->_address_of_record,
+                               classified_binding->_id,
+                               classified_binding->_b->_uri,
+                               0);
```
eleanor-merry: Why is classified_binding->_b->_expires not 0 at this point?

src/subscriber_data_manager.cpp
```diff
@@ -234,12 +264,29 @@ Store::Status SubscriberDataManager::set_aor_data(
...
 
   if (_primary_sdm)
   {
-    // We may have been given some messages to send by the caller.
+    /////////////////////////////////
+    // 5. Log new / extended bindings
+    /////////////////////////////////
+    if (_analytics != NULL)
+    {
+      log_new_extended_bindings(classified_bindings);
```
eleanor-merry: Likewise, can this be log_new_or_extended_bindings?

src/subscriber_data_manager.cpp
```diff
@@ -485,12 +623,8 @@ SubscriberDataManager::AoR* SubscriberDataManager::Connector::deserialize_aor(
 {
   AoR* aor = NULL;
 
-  for(std::vector<SerializerDeserializer*>::iterator it = _deserializers.begin();
-      it != _deserializers.end();
-      ++it)
+  for(SerializerDeserializer* deserializer : _deserializers)
```
eleanor-merry: Space between for and (

src/subscriber_data_manager.cpp
```diff
@@ -1532,20 +1661,18 @@ void SubscriberDataManager::NotifySender::send_notifys_for_expired_subscriptions
         reg_state = NotifyUtils::RegistrationState::TERMINATED;
       }
 
-      std::vector<NotifyUtils::BindingNotifyInformation*> binding_notify;
+      ClassifiedBindings binding_notify;
```
eleanor-merry: This isn't a classified binding (yes, it's the same type, but you're not using any classification here).
Same comment for the other uses as well.

src/subscriber_data_manager.cpp
```diff
@@ -1509,15 +1641,12 @@ void SubscriberDataManager::NotifySender::send_notifys_for_expired_subscriptions
...
-           aor_current_b != aor_pair->get_current()->bindings().end();
-           ++aor_current_b)
+      for (std::pair<std::string, SubscriberDataManager::AoR::Binding*> aor_current_b : 
+             aor_pair->get_current()->bindings())
       {
-        if (!aor_current_b->second->_emergency_registration)
+        if (!aor_current_b.second->_emergency_registration)
         {
           bindings_remaining = true;
-          break;
```
eleanor-merry: I'm not a fan of removing this break just for coverage reasons - surely LCOV would be a better choice?

src/subscriber_data_manager.cpp
pdavies: I think these other uses are classified bindings, since they are a binding paired with a contact event that classifies them as created, refreshed, expired, etc. I agree it wasn't in a nice state, because I only changed the uses in subscriber_data_manager, and was too timid to go around renaming things throughout the codebase. I've reverted these cases for consistency.

pdavies:
Where do you think is the best place to track them?

Issue #1586, opened 2016-11-10
Sprout handlers don't have a load monitor


johadalin:
#### Symptoms

Initially seen that the sprout ping_handler was not returning a 503 when overloaded. Tracing this back, we found that the sprout http stacks were not initialised with a load_monitor at https://github.com/Metaswitch/sprout/blob/dev/src/main.cpp#L2011 . This means that any request to these handlers does not pass through any load monitoring/overload control.

#### Impact

This that scripts such as poll-http will time out rather than correctly receiving a 503 in the case of overload. Other handlers would also not be subject to correct overload control behaviour.

Issue #1570, opened 2016-10-19
poll_sprout_http is too aggressive


eleanor-merry:
#### Symptoms

When running medium stress through Sprout I found that Sprout was being killed by poll_sprout_http failing (so monit sends an abort to Sprout) reasonably frequently (every half hour or so). It can also take more than 20 seconds for the poll_sprout_http script to start passing on start up, so this can lead to Sprout being killed multiple times in quick succession.
#### Impact

Sprout isn't stable at even medium stress. To work around this, I unmonitored the poll_sprout_http script.
#### Release and environment

From at least the Onix release

johadalin:
@eleanor-merry I have a couple of questions on this.

* The PRs above should fix any occurrence of being killed due to overload, but you also mention it taking more than 2 polling cycles to start passing; should we be addressing the monit scripts to resolve this?

* What setup did you have when this issue was visible? Before closing this issue, it would be good to test the same setup and see if the problem persists at all.

johadalin:
Discussed the concerns above, and agreed that closing this issue makes sense. 
The fix has been tested in isolation, and we are fairly certain that the original problem seen in this issue will be resolved by this change, so any new errors will be raised as separate issues. 
The Monit concern will also be raised separately if investigation shows it to be a true issue.

PR #1587, opened 2016-11-10
[Reviewer: Ellie] Add load monitor to the http stacks.

johadalin:
In testing https://github.com/Metaswitch/clearwater-infrastructure/pull/393, and investigating why sprout wouldn't build with the hacked in change for forcing 503 returns, i found that there was no load monitor on the signaling http stack. 
This meant that even when overloaded, sprout would not return a 503 to the poll-http script.

To test that this works, i altered https://github.com/Metaswitch/cpp-common/blob/54371f20e89c60a04218bdeebe90d687f26345fc/src/httpstack.cpp#L313 to test `!(_load_monitor->admit_request(trail)`, forcing overload while testing the load monitor.

This has not been tested on a system under significant load

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2010,7 +2010,8 @@ int main(int argc, char* argv[])
   // with it.
   HttpStack* http_stack_sig = new HttpStack(opt.http_threads,
                                             exception_handler,
-                                            access_logger);
+                                            access_logger,
+                                            load_monitor);
   try
   {
     http_stack_sig->initialize();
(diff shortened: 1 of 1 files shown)


johadalin:
This will resolve https://github.com/Metaswitch/sprout/issues/1586, and is part of resolving https://github.com/Metaswitch/sprout/issues/1570, as the fix in https://github.com/Metaswitch/clearwater-infrastructure/pull/393 needs this in place

eleanor-merry:
:shipit:

PR #1592, opened 2016-11-16
use already-selected transport for CANCEL

dimbleby:
Hopefully this is an uncontroversial fragment extracted from the various transport-related hacks that John made.

I don't know what exactly went wrong without this fix, but the PJSIP docs do say that users should prefer to call `pjsip_tsx_set_transport`, rather than `pjsip_tx_data_set_transport` directly, and this looks sensible enough.

Unit tests don't fail regardless of the state of this code (though it is covered) - if you can think of a likely testcase that would prove why this is necessary, let me know.


mirw:
Looks good.  Regarding UT, I think this is pushing the limits of what our UT infrastructure does.

PR #1593, opened 2016-11-16
Ignore route header, route on request URI

dimbleby:
NOT FOR MERGING!

This is the first of the interop fixes.  Incoming INVITEs don't have Route headers set up in the approved IMS way; so when handling such messages we should only look at the request URI.

For discussion.  I hope that we'll be able to merge this behind a config option - what do you think?


mirw:
Maybe a better fix here would be to allow Sproutlets to register based on the request URI rather than (or as well as) the Route header?  Thoughts?

dimbleby:
So the suggestion is:

-  SproutletProxy::target_sproutlet picks out both the route header and the request URI
-  Then in the loop when it asks "is it this sproutlet?" it uses whichever of those is appropriate, according to what the sproutlet asked for

I'd be OK with that.  

Is that indeed what you meant?  Ping this back to me if so and I'll code it up.

mirw:
That was what I meant, but on further thought, I'm wondering whether we should take this into mainline at all (as we won't be able to test it).  You're going to figure out how this is going to be managed going forward.

dimbleby:
This one-liner is trivial enough that I think we're OK to track it privately on our app server project

PR #1594, opened 2016-11-16
reuse incoming transport

dimbleby:
NOT FOR MERGING!

The ugliest part of our transport hacks: re-use incoming transport for outgoing messages.

John dabbled with making this configurable per application server but didn't get this into any sort of useful state, and I'm not convinced that it's the right scope to configure this anyway.  So I've removed that code from this diff.

I think I'd prefer to control this by a global config option whose meaning is basically "we're acting as an app server for a non-IMS call agent" (maybe we can come up with something that doesn't sound _quite_ so specific).  But definitely up for discussion.

### DIFF
--- a/include/sproutlet.h
+++ b/include/sproutlet.h
@@ -93,6 +93,13 @@ public:
   ///
   virtual pjsip_msg* original_request() = 0;
 
+  /// Sets the transport on the given message to be the same as on the
+  /// original incoming request.
+  ///
+  /// @param  req          - The request message on which to set the
+  //                         transport.
+  virtual void copy_original_transport(pjsip_msg* req) = 0;
+
   /// Returns the top Route header from the original incoming request.  This
   /// can be inpsected by the Sproutlet, but should not be modified.  Note that
   /// this Route header is removed from the request passed to the Sproutlet on
@@ -352,6 +359,13 @@ protected:
   ///
   pjsip_msg* original_request()
     {return _helper->original_request();}
+
+  /// Sets the transport on this request to be the same as on the original.
+  ///
+  /// @param  req          - The request message on which to set the
+  ///                        transport.
+  void copy_original_transport(pjsip_msg* req)
+    {_helper->copy_original_transport(req);}
 
   /// Returns a URI that could be used to route back to the current Sproutlet.
   /// This URI may contain pre-loaded parameters that should not be modified
--- a/src/basicproxy.cpp
+++ b/src/basicproxy.cpp
@@ -423,6 +423,7 @@ BasicProxy::UASTsx* BasicProxy::create_uas_tsx()
 BasicProxy::UASTsx::UASTsx(BasicProxy* proxy) :
   _proxy(proxy),
   _req(NULL),
+  _original_transport(NULL),
   _tsx(NULL),
   _lock(NULL),
   _trail(0),
@@ -493,6 +494,13 @@ BasicProxy::UASTsx::~UASTsx()
     }
   }
 
+  if (_original_transport != NULL)
+  {
+    TRC_DEBUG("Free original transport");
+    pjsip_transport_dec_ref(_original_transport);
+    _original_transport = NULL;
+  }
+
   if (_req != NULL)
   {
     TRC_DEBUG("Free original request");
@@ -550,6 +558,12 @@ pj_status_t BasicProxy::UASTsx::init(pjsip_rx_data* rdata)
     _pending_destroy = true;
     return PJ_ENOMEM;
     // LCOV_EXCL_STOP
+  }
+
+  _original_transport = rdata->tp_info.transport;
+  if (_original_transport != NULL)
+  {
+    pjsip_transport_add_ref(_original_transport);
   }
 
   if (rdata->msg_info.msg->line.req.method.id != PJSIP_ACK_METHOD)
--- a/src/sproutletproxy.cpp
+++ b/src/sproutletproxy.cpp
@@ -1228,16 +1228,20 @@ void SproutletWrapper::copy_original_transport(pjsip_msg* req)
   // Get the original transport.
   if (_original_transport == NULL)
   {
+    // LCOV_EXCL_START - defensive code not hit in UT
     TRC_WARNING("Sproutlet tried to copy transport from unknown original");
     return;
+    // LCOV_EXCL_STOP
   }
 
   // Get this request's tdata from the map of clones.
   Packets::iterator it = _packets.find(req);
   if (it == _packets.end())
   {
+    // LCOV_EXCL_START - defensive code not hit in UT
     TRC_WARNING("Sproutlet tried to copy transport on an unknown request");
     return;
+    // LCOV_EXCL_STOP
   }
   pjsip_tx_data* tdata = it->second;
 
--- a/src/sproutletproxy.cpp
+++ b/src/sproutletproxy.cpp
@@ -573,6 +573,7 @@ pj_status_t SproutletProxy::UASTsx::init(pjsip_rx_data* rdata)
                                    sproutlet,
                                    alias,
                                    _req,
+                                   _original_transport,
                                    trail());
     }
   }
@@ -827,6 +828,7 @@ void SproutletProxy::UASTsx::schedule_requests()
                                                             sproutlet,
                                                             alias,
                                                             req.req,
+                                                            _original_transport,
                                                             trail());
 
         // Set up the mappings.
@@ -1080,6 +1082,7 @@ SproutletWrapper::SproutletWrapper(SproutletProxy* proxy,
                                    Sproutlet* sproutlet,
                                    const std::string& sproutlet_alias,
                                    pjsip_tx_data* req,
+                                   pjsip_transport* original_transport,
                                    SAS::TrailId trail_id) :
   _proxy(proxy),
   _proxy_tsx(proxy_tsx),
@@ -1089,6 +1092,7 @@ SproutletWrapper::SproutletWrapper(SproutletProxy* proxy,
   _id(""),
   _req(req),
   _req_type(),
+  _original_transport(original_transport),
   _packets(),
   _send_requests(),
   _send_responses(),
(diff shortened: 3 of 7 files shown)


mirw:
Is there a reason this can't be done by the Sproutlet itself (by calling into the newly-added PJUtils::set_tx_transport method explicitly)?

...and I didn't quite understand why this was required for responses as well as forwarded requests?  Can you elaborate on that?

dimbleby:
The Sproutlet doesn't see the original message.  SproutletWrapper [takes a clone](https://github.com/Metaswitch/sprout/blob/bfd66cd3f807b46a33c0bf2b3eb0983eb0fefda2/src/sproutletproxy.cpp#L1582-L1583) and passes that to the Sproutlet.  So the Sproutlet has no idea what the original transport was, unless this is preserved in the clone - right?

Also we need to preserve the transport for messages that don't hit the sproutlet eg polling OPTIONs.

Also there's the case where we statelessly forward a late response, which is already an odd man out.

You may be correct that this isn't necessary for responses, I'm not sure - it's possible that John erred on the side of caution and overshot.  I can try taking that instance of `set_tx_transport` out and see if anything breaks.

mirw:
Thanks!  Can't we provide an API on the SproutletTsxHelper to retrieve the transport from the original request (either stored off on the SproutletTsxHelper or in the UASTsx)?  Then we don't have to clone it onto the tx_data (which feels wrong).

dimbleby:
We can do that and I even have a version of the code along those lines, which I'll push.  But I think you're still not going to be quite happy.

In particular, the 'original' version of the request that the SproutletTsxHelper has isn't really the original at all but it itself a clone (from [here](https://github.com/Metaswitch/sprout/blob/094ef5c4c78954212a509416920c6ed3c2813119/src/basicproxy.cpp#L544), I think).

And also I still need to copy the transport for CANCELs.  

So even in this updated version of the fix, I still have to copy the transport in the rx version of clone_msg.

Views?

src/basicproxy.cpp
```diff
@@ -167,6 +167,16 @@ pj_bool_t BasicProxy::on_rx_response(pjsip_rx_data *rdata)
       res_addr.dst_host.addr.port = hvia->sent_by.port;
     }
 
+    // Re-use incoming transport for response.
```
mirw: This feels like a big issue as it's not controllable by the sproutlet (which would imply we'd need global configuration to enable this behavior).  Is there a reason that we can't just use the transport encoded in the Via header?

src/pjutils.cpp
```diff
@@ -698,6 +698,16 @@ pjsip_tx_data* PJUtils::clone_msg(pjsip_endpoint* endpt,
     set_trail(clone, get_trail(rdata));
     TRC_DEBUG("Cloned %s to %s", pjsip_rx_data_get_info(rdata), clone->obj_name);
 
+    pjsip_transport* transport = rdata->tp_info.transport;
```
mirw: As you say, we can't do this without affecting more flows.  Do we actually need to keep the original transport attached to the pjsip_tx_data, or can we propagate it separately?

src/basicproxy.cpp
dimbleby: That's what the existing code already does!

But alas, the peer in question does not include an rport - so it just doesn't give the desired behaviour in this particular case.

src/pjutils.cpp
dimbleby: Yes, possibly - more in other comment.

dimbleby:
re CANCELs: we also need the transport clone that I'm currently getting from [this](https://github.com/Metaswitch/sprout/blob/e88a6faa1d879e8f0760b0d97a13e70d2939fbed/src/sproutletproxy.cpp#L604) call to the rx version of `msg_clone()`.

Looks like the handling of CANCELs is quite different than other messages.  I guess the best position we're likely to get to is something like this:

- carry the transport around separately from the `_req`, so that we can implement `clone_original_transport()` cleanly for non-CANCELs
- do something parallel for CANCELs, so that sproutlets can also `clone_original_transport` for a CANCEL from `on_rx_cancel()`
  - I haven't yet understood how different exactly the CANCEL code will need to be
- not so clear what to do about the forwarding of late responses, per comments above, for peers that don't include rport in Via headers.
  - Maybe we should take that out of this pull request and decide separately

mirw:
Yes, that sounds sensible.  I'm happy to move the Via discussion into a separate pull request.

dimbleby:
OK, the latest set of changes should bring this to something close to acceptable, I hope...

Remarks:
* contrary to what I said above, there is no special-case handling for CANCELs
  * I'm sure I had previously done live testing showing that this was required, but I'm currently seeing the opposite.
  * I _think_ it makes sense that I shouldn't have to do anything; that the CANCEL should automatically follow the corresponding INVITE.  So In that case I'm reasonably happy that my previous testing was just wrong somehow and that this version of the fix is sufficient...
  * ... right?
* I've referenced (and dereferenced) the stored transport at pretty much every opportunity
  * I suspect I've ended up doing this more than actually required, but I don't really understand everything's lifetime so have played it safe
  * Maybe something to consider in review
* Existing UTs pass but I haven't written anything new
  * If you're happy with the way that this fix is going, I expect you'll want something: that has a sproutlet call `copy_original_transport()` and verifies that this does what's expected
  * ... and if you can point me at an existing test that would make a good starting point for this, that would be marvellous

mirw:
I think this looks OK to me, but I'd like to get a second opinion (and input on the UT question).

@AlexHockey, please can you take a look, as I suspect you know this code better than me?

Also, @dimbleby, I think we need to run clearwater-live-test too here.

src/sproutletproxy.cpp
```diff
@@ -1114,6 +1118,11 @@ SproutletWrapper::SproutletWrapper(SproutletProxy* proxy,
   _pending_timers(),
   _trail_id(trail_id)
 {
+  if (_original_transport != NULL)
+  {
+    pjsip_transport_add_ref(_original_transport);
+  }
```
AlexHockey: I don't think you need this reference (or the dec ref below) as a reference is held by the UasTsx and this exists for as long as it has any `SproutletWrappers`. However I don't think holding the reference will do any harm either.

dimbleby:
reassigning to get your reply to https://github.com/Metaswitch/sprout/pull/1594#discussion_r89502222

include/sproutlet.h
```diff
@@ -93,6 +93,13 @@ class SproutletTsxHelper
   ///
   virtual pjsip_msg* original_request() = 0;
 
+  /// Sets the transport on the given message to be the same as on the
+  /// original incoming request.
+  ///
+  /// @param  req          - The request message on which to set the
+  //                         transport.
+  virtual void copy_original_transport(pjsip_msg* req) = 0;
```
AlexHockey: Ah yes, I hadn't properly considered the "sandboxing" aspect of the sproutlet API. Let's leave your code as it is, but we'll probably deprecate the new API at some point in the future.

dimbleby:
Super, thanks.  Now with added unit test.

Ah, I see I have a conflict.  Well, this was getting messy anyway - I'll rebase on master, squash, and force push the branch.

dimbleby:
Anything else need doing or are we good to go?

src/ut/sproutletproxy_test.cpp
```diff
@@ -2327,3 +2343,69 @@ TEST_F(SproutletProxyTest, ServiceExtraction)
...
+  RespMatcher(100).matches(tdata->msg);
+  tp->expect_target(tdata);
+  EXPECT_EQ("To: <sip:bob@awaydomain>", get_headers(tdata->msg, "To")); // No tag
+  free_txdata();
+
+  // Request is forwarded to the node in the second Route header, reusing the
+  // incoming transport
+  ASSERT_EQ(1, txdata_count());
+  tdata = current_txdata();
+  expect_target("TCP", "10.10.28.1", 49152, tdata);
```
AlexHockey: All this checks is that the request is sent on a transport with the right type, dest IP address and dest port. It doesn't test that it comes back on exactly the same transport. You need to call to call `TransportFlow::Expect_target` with the strict flag. 

In fact, if you've only ever told PJSIP about one transport, it feels like there is a reasonable chance it picks that transport by accident rather than down to the code you've added. So it might be to send in two distinct INVITEs on two different transports and check each one comes out on the right one.

src/ut/sproutletproxy_test.cpp
```diff
@@ -2327,3 +2343,69 @@ TEST_F(SproutletProxyTest, ServiceExtraction)
...
+
+  // Check the RequestURI has not been altered.
+  EXPECT_EQ("sip:bob@awaydomain", str_uri(tdata->msg->line.req.uri));
+
+  // Check the first Route header has been removed.
+  EXPECT_EQ("Route: <sip:node1.awaydomain;transport=TCP;lr>",
+            get_headers(tdata->msg, "Route"));
+
+  // Check no Record-Route headers have been added.
+  EXPECT_EQ("", get_headers(tdata->msg, "Record-Route"));
```
AlexHockey: Do you need to check the req URI, route header and record-route header? They don;t feel particularly relevant to the test.

dimbleby:
Good stuff, thanks - does the latest revision look good?

AlexHockey:
:+1:

dimbleby:
Matt - you've said that we should run clearwater-live-test here.  I'm not set up to do that.  

Can you [point me at someone who can] help, please?

sebrexmetaswitch:
@dimbleby I've picked up looking at this now.

First off, I merged this branch to my local dev branch and ran `make full_test` and it looks like you've broken coverage in `sproutletproxy.cpp`

When I run `make coverage_raw | grep "sproutletproxy"`, I see:

> src/sproutletproxy.cpp                       854     850    99%   1253-1254,1261-1262

Those line numbers don't quite align with your PR since I've merged it to my dev branch, but I think it's just the warning logs you've added:

https://github.com/Metaswitch/sprout/pull/1594/files#diff-75af8e842ec03e21817a448de6bdf026R1231
https://github.com/Metaswitch/sprout/pull/1594/files#diff-75af8e842ec03e21817a448de6bdf026R1239

It may just make sense to add an exclusion around those two, since it may not make sense to UT that.

Back to you to fix up coverage. When running the UTs, please make sure you run `make full_test` as well :)

dimbleby:
done!

sebrexmetaswitch:
Thanks.

I've now run the clearwater-live-tests over this and they passed, so I think it's good for you to merge.

dimbleby:
fab - thanks!

PR #1596, opened 2016-11-17
[Reviewer: Alex] Skip logging SAS markers if we can get away with it

rkd-msw:
As discussed by email. I'll arrange some testing to see how much of a difference this makes.

### DIFF
--- a/src/common_sip_processing.cpp
+++ b/src/common_sip_processing.cpp
@@ -229,10 +229,16 @@ static void sas_log_rx_msg(pjsip_rx_data* rdata)
   // Store the trail in the message as it gets passed up the stack.
   set_trail(rdata, trail);
 
-  PJUtils::report_sas_to_from_markers(trail, rdata->msg_info.msg);
+  // Raise SAS markers on initial requests only - responses in the same
+  // transaction will have the same trail ID so don't need additional markers
+  if (rdata->msg_info.msg->type == PJSIP_REQUEST_MSG)
+  {
+    PJUtils::report_sas_to_from_markers(trail, rdata->msg_info.msg);
 
-  pjsip_cid_hdr* cid = (pjsip_cid_hdr*)rdata->msg_info.cid;
-  PJUtils::mark_sas_call_branch_ids(trail, cid, rdata->msg_info.msg);
+    pjsip_cid_hdr* cid = (pjsip_cid_hdr*)rdata->msg_info.cid;
+
+    PJUtils::mark_sas_call_branch_ids(trail, cid, rdata->msg_info.msg);
+  }
 
   // Log the message event.
   SAS::Event event(trail, SASEvent::RX_SIP_MSG, 0);
@@ -256,9 +262,14 @@ static void sas_log_tx_msg(pjsip_tx_data *tdata)
   }
   else if (trail != 0)
   {
-    PJUtils::report_sas_to_from_markers(trail, tdata->msg);
+    // Raise SAS markers on initial requests only - responses in the same
+    // transaction will have the same trail ID so don't need additional markers
+    if (tdata->msg->type == PJSIP_REQUEST_MSG)
+    {
+      PJUtils::report_sas_to_from_markers(trail, tdata->msg);
 
-    PJUtils::mark_sas_call_branch_ids(trail, NULL, tdata->msg);
+      PJUtils::mark_sas_call_branch_ids(trail, NULL, tdata->msg);
+    }
 
     // Log the message event.
     SAS::Event event(trail, SASEvent::TX_SIP_MSG, 0);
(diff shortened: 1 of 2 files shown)


src/common_sip_processing.cpp
```diff
@@ -229,10 +229,16 @@ static void sas_log_rx_msg(pjsip_rx_data* rdata)
   // Store the trail in the message as it gets passed up the stack.
   set_trail(rdata, trail);
 
-  PJUtils::report_sas_to_from_markers(trail, rdata->msg_info.msg);
+  // Raise SAS markers on initial requests only - responses in the same
+  // transaction will have the same trail ID so don't need additional markers
+  if (rdata->msg_info.msg->type == PJSIP_REQUEST_MSG)
```
AlexHockey: On further thought, I think we should raise markers if we haven't found an existing trail to use. This means we typically won't raise markers on negative ACKs, CANCELs or responses, but we will raise them if we receive a response late and are forced to forward it statelessly.

src/common_sip_processing.cpp
```diff
@@ -256,9 +262,14 @@ static void sas_log_tx_msg(pjsip_tx_data *tdata)
   }
   else if (trail != 0)
   {
-    PJUtils::report_sas_to_from_markers(trail, tdata->msg);
+    // Raise SAS markers on initial requests only - responses in the same
+    // transaction will have the same trail ID so don't need additional markers
+    if (tdata->msg->type == PJSIP_REQUEST_MSG)
```
AlexHockey: But here's its correct to only log on a request. The logic being we don't need to send on the response as we'll already have logged one way or another - either on the request, or when we received a response that got statelessly forwarded.

src/common_sip_processing.cpp
```diff
@@ -354,6 +362,10 @@ static pj_bool_t process_on_rx_msg(pjsip_rx_data* rdata)
 
     if (rdata->msg_info.msg->type == PJSIP_REQUEST_MSG)
     {
+      
+      PJUtils::report_sas_to_from_markers(trail, rdata->msg_info.msg);
+      PJUtils::mark_sas_call_branch_ids(trail, rdata->msg_info.cid, rdata->msg_info.msg);
```
AlexHockey: Do we need this at all? Haven't we already called `sas_log_rx_msg`?

src/pjutils.cpp
```diff
@@ -1662,34 +1662,43 @@ std::string PJUtils::get_header_value(pjsip_hdr* header)
...
-    scope = SAS::Marker::Scope::None;
-  }
-  else
-  {
-    scope = SAS::Marker::Scope::Trace;
+    // We should only log SAS markers on a request, to minimise data written to
+    // SAS - responses will be raised on the same trail so don't need
+    // additional markers for correlation.
+    TRC_DEBUG("mark_sas_call_branch_ids incorrectly called on a response");
+    return;
```
AlexHockey: Follow-on from the above comment, I think we do sometimes need to log on responses.

src/pjutils.cpp
```diff
@@ -1662,34 +1662,43 @@ std::string PJUtils::get_header_value(pjsip_hdr* header)
...
+                                (pjsip_method_cmp(&msg->line.req.method, pjsip_get_subscribe_method()) == 0) ||
+                                (pjsip_method_cmp(&msg->line.req.method, pjsip_get_notify_method()) == 0));
+  
   if (cid_hdr != NULL)
   {
     TRC_DEBUG("Logging SAS Call-ID marker, Call-ID %.*s", cid_hdr->id.slen, cid_hdr->id.ptr);
     SAS::Marker cid_marker(trail, MARKER_ID_SIP_CALL_ID, 1u);
     cid_marker.add_var_param(cid_hdr->id.slen, cid_hdr->id.ptr);
-    SAS::report_marker(cid_marker, scope);
+    SAS::report_marker(cid_marker, branch_id_correlation ? SAS::Marker::Scope::Trace : SAS::Marker::Scope::None);
```
AlexHockey: Pretty sure this ternary is the wrong way round.

src/pjutils.cpp
```diff
@@ -1662,34 +1662,43 @@ std::string PJUtils::get_header_value(pjsip_hdr* header)
   return std::string(buf2, len);
 }
 
-/// Add SAS markers for the specified call ID and branch IDs on the message (either may be omitted).
+/// Add SAS markers for the specified call ID and branch IDs on the message (call ID may be omitted, but not message).
 void PJUtils::mark_sas_call_branch_ids(const SAS::TrailId trail, pjsip_cid_hdr* cid_hdr, pjsip_msg* msg)
```
AlexHockey: (Not your change, but...) Do we even need to pass the Call ID header as a separate parameter?

rkd-msw: Probably not, but that would be a more widespread change than I want to make here, and it's not directly relevant to this change.

rkd-msw:
I've applied your suggestions, and @pdavies has tested with SASHole to confirm that:

- association markers on each REGISTER drop from 31 to 24
- association markers on an on-net call (counting both caller and callee) drop from 38 to 15
- there's no change to the ability to associate these traces (i.e. I've fixed the regression you pointed out)

I'll merge once code freeze ends.

src/common_sip_processing.cpp
```diff
@@ -224,15 +225,22 @@ static void sas_log_rx_msg(pjsip_rx_data* rdata)
...
       trail = SAS::new_trail(1u);
     }
+    first_message_in_trail = true;
   }
 
   // Store the trail in the message as it gets passed up the stack.
   set_trail(rdata, trail);
 
-  PJUtils::report_sas_to_from_markers(trail, rdata->msg_info.msg);
+  // Raise SAS markers on the first message in a trail only only - subsequent
```
pdavies: only only

PR #1597, opened 2016-11-17
Fix obvious typo in comparison of Targets

dimbleby:
just a random drive-by fix for you, because it seemed like the right thing to do


AlexHockey:
:+1:

PR #1605, opened 2016-11-21
[Reviewer: Ellie] Handle digests and keys with null bytes in them

### LINKED ISSUE #1599: Sprout will incorrectly fail AKA authentication if the RES/XRES contains an 0x00 byte
#### Symptoms

I added an AKAv1 live test, and noticed it intermittently failing (with my authenticated REGISTER being rejected with a 403 Forbidden). A SAS trace is at \\ad\csd\bigfiles\clearwater\issue-attachments\.

Looking at this SAS trace, the SIP-Authorization in the MAA is 0x346667c3e9200037 - i.e. one of the bytes is 0x00. I added some logging to my test script, which confirms that the failures happen when (and only when) the RES contains a zero byte.

#### Impact

This causes a REGISTER to be rejected with a 403 when it happens, and https://tools.ietf.org/html/rfc3261#section-21.4.4 says that 403 means a request SHOULD NOT be repeated, so a subscriber might lose service when this happens.

There are 8 bytes in the RES, and any one of them being 0x00 causes this, so this will happen on (1/256 + 1/255 ... + 1/249) of REGISTERs - i.e. this will make about 3.17% of AKAv1 REGISTERs fail. (That matches up with the frequency of test failures I'm seeing).

#### Release and environment 

Quilava release, on VMware (but unlikely to be specific to either). I'm using the vmware.obu-enfield.test deployment for this testing.

#### Steps to reproduce

Attempt AKAv1 registration with a RES that includes 0x00. (This could probably be reproduced in UT, but RES isn't easy to control live, so you'd just need to REGISTER repeatedly.)

richardwhiuk:
This fixes #1599.

We fix this by using the known string length, rather than just up until the first null byte when we have decoded the hex from Homestead. I've added a UT for handling null bytes in a response, which fails before this fix, and succeeds afterwards.

This also fixes the other issue Rob spotted, where if the digest includes a null byte, we'll fail to handle the request correctly. It also fixes an issue where the buffer given to us by OpenSSL isn't totally guaranteed to be null terminated (but probably is).

This needs https://github.com/Metaswitch/pjsip-upstream/pull/52 for the pj_strdup4 function.

### DIFF
--- a/src/ut/authentication_test.cpp
+++ b/src/ut/authentication_test.cpp
@@ -206,6 +206,8 @@ class AuthenticationTest : public BaseAuthenticationTest
     ASSERT_EQ(PJ_SUCCESS, ret);
   }
 
+  void TestAKAAuthSuccess(char* key);
+
   static void TearDownTestCase()
   {
     destroy_authentication();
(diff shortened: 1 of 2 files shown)


src/ut/authentication_test.cpp
```diff
@@ -1369,6 +1361,35 @@ TEST_F(AuthenticationTest, AKAAuthSuccess)
   _hss_connection->delete_result("/impi/6505550001%40homedomain/av/aka?impu=sip%3A6505550001%40homedomain");
 }
 
+TEST_F(AuthenticationTest, AKAAuthSuccess)
```
eleanor-merry: Can you comment what these UTs are testing?

eleanor-merry:
One thing actually - when you merge this fix it will close the associated issue. Should you be passing it back to @rkd-msw for retest instead?

rkd-msw:
@nicolemcp - you might want to backport this.

PR #1606, opened 2016-11-23
Reuse existing transport to forward late response

dimbleby:
NOT FOR MERGING!

As discussed in #1594 - when we want to reuse the incoming transport for outgoing messages, and when the peer does not support the rport parameter, we've resorted to this change.

Can we think of a way of making this acceptable?

Or perhaps we should treat this in the same way as whatever we do with #1593 (which we haven't yet absolutely decided - but expect that it likely won't be merged into sprout-proper)


mirw:
I can't really see a way of making this acceptable for the mainline.  If we absolutely had to have it in, I think we'd want to either implement a per-IP-address set of fixbits or MMF-style function.  I think we should wait until you've resolved #1593.

dimbleby:
Actually, I think I may be being unfair on our peer and there could be a sprout issue here.  

* Sprout receives INVITE from 192.91.191.29:59851
* Sprout sends 100 Trying, like this:

````
.uSIP/2.0 100 Trying
Via: SIP/2.0/TCP 127.0.1.1:5060;received=192.91.191.29;branch=z9hG4bK-17758-1-0
Call-ID: 1-17758@127.0.1.1
From: "sipp" <sip:16468460100@127.0.1.1>;tag=17758SIPpTag001
To: "sut" <sip:16468460101@127.0.1.1>
CSeq: 1 INVITE
Content-Length:  0
````

Observe: no rport in the Via header.

Isn't this where the rport is supposed to have been introduced into the flow?

**Comment by dimbleby** [comment]
Ah no, I see, the peer needs to include an rport-with-no-value to convince sprout that it would understand the parameter.

**Comment by dimbleby** [comment]
Ah, but that's because the transport is TCP - in which case I think it's probably _correct_ not to include an rport parameter.  RFC3581 seems to be specifically aimed at UDP.  

We're not the first people to wonder about this - eg see [this discussion](https://www.ietf.org/mail-archive/web/sip/current/msg09548.html).

(pjsip and hence sprout does include rport on TCP transport - but at least some people [think that this is a mistake](https://freeswitch.org/jira/si/jira.issueviews:issue-html/FS-6612/FS-6612.html).)

So if we can't rely on the rport, it seems that failure to forward a late response in this case is just correct behaviour.

So in that case, I think that the fix that I want to make is quite different.  The trigger for all this was that we were seeing this error log:

```
22-09-2016 02:12:35.527 UTC Error pjsip: tcpc0x7fb770b6 TCP connect() error: Operation timed out (PJ_ETIMEDOUT) [code=70009]
```

from [here](https://github.com/Metaswitch/pjsip-upstream/blob/de8993a27f734db4f65b150ecfbe61f9428f87a5/pjsip/src/pjsip/sip_transport_tcp.c#L1532).  

I propose downgrading this so that it's only a Warning and not an Error.

(which seems reasonable anyway: logging an error because of something a peer does or doesn't do isn't usually right).

What do you think - would you be inclined to take a pull request along those lines?

mirw:
I'm happy with downgrading the error to a warning.  This might be a bit complicated because it's generated by PJSIP.

dimbleby:
Cool, let's abandon this one then.

PR #1613, opened 2016-11-28
A retransmitted CANCEL caused an Error log

dimbleby:
Specifically, this one:

    Error basicproxy.cpp:1798: Error sending CANCEL, Object already exists (PJ_EEXISTS)

this pull request includes a unit test case that hits this, and the simplest possible fix to avoid the unwarranted error log.

### DIFF
--- a/src/basicproxy.cpp
+++ b/src/basicproxy.cpp
@@ -1776,7 +1776,21 @@ void BasicProxy::UACTsx::cancel_pending_tsx(int st_code)
           status = pjsip_tsx_send_msg(_cancel_tsx, cancel);
         }
 
-        if (status != PJ_SUCCESS)
+        // There are some known but hard-to-hit ways for this to fail - we
+        // only WARN for these:
+        //
+        // - EEXISTS can be hit if we've already sent out the CANCEL and then
+        //   see a transport failure on the incoming side of the call - that
+        //   brings us through this path for a second time.
+        //
+        // - EINVALIDOP can be hit if the transport on which we're trying to
+        //   send the CANCEL is unavailable.
+        if ((status == PJ_EEXISTS) || (status == PJ_EINVALIDOP))
+        {
+          TRC_WARNING("Error sending CANCEL, %s",
+                      PJUtils::pj_status_to_string(status).c_str());
+        }
+        else if (status != PJ_SUCCESS)
         {
           //LCOV_EXCL_START
           TRC_ERROR("Error sending CANCEL, %s",
--- a/src/ut/basicproxy_test.cpp
+++ b/src/ut/basicproxy_test.cpp
@@ -3876,3 +3876,94 @@ TEST_F(BasicProxyTest, StatelessProxyNoBlacklistOnTimeout)
   free_txdata();
   delete tp;
 }
+
+TEST_F(BasicProxyTest, TransportFailureWithCancelPending)
+{
+  // Tests transport failure while a CANCEL is outstanding.
+
+  // Create a TCP connection to the listening port.
+  TransportFlow* tp = new TransportFlow(TransportFlow::Protocol::TCP,
+                                        stack_data.scscf_port,
+                                        "1.2.3.4",
+                                        49152);
+
+  // Add three test targets for bob@homedomain, all with a URI plus a path with a
+  // single proxy.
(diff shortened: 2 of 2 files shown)


dimbleby:
Another non-urgent one for you, please.

dimbleby:
We're also seeing this same error log but with error `PJ_EINVALIDOP`:

    30-11-2016 00:54:58.019 UTC Error basicproxy.cpp:1801: Error sending CANCEL on call ID A751E4D5@172.16.19.50, Invalid operation (PJ_EINVALIDOP)

(I added the extra detail about the call ID to the log in a private build, to help me track this down).

I've been able to reproduce this live by Ctrl-C'ing in the middle of a SIPP script, at which point:

* The TCP connection fails, which we notice on the UAS side of the call
* That causes us to want to send a CANCEL out on the UAC side
* This fails (because it's using the same TCP transport)

I _think_ that ultimately `PJ_EINVALIDOP` is coming from [here](https://github.com/Metaswitch/pjsip-upstream/blob/de8993a27f734db4f65b150ecfbe61f9428f87a5/pjlib/src/pj/activesock.c#L683-L684).

I guess our particular service is more than usually exposed to this because the transport-reuse thing forces a narrow window to be wide open, but I imagine that in principle something like this could happen for anyone.

I'd very much like to stop seeing this error in our logs too.  What do you think - can I just add `PJ_EINVALIDOP` to the "don't log" cases?  Maybe make a warning rather than an error log for this one?

sebrexmetaswitch:
I'd prefer it if we didn't lose any info from our logs, so rather than suppressing the logs entirely, could you make this a warning log instead? And then I think your suggestion of adding PJ_EINVALIDOP to the warning log case is fine.

Can you check that `make full_test` still passes after your additions too (you may have already, but just wanted to check)

dimbleby:
Oh, interesting - doing that and checking the coverage has revealed that my unit test wasn't hitting this in the first place (though we do hit it live).  I really thought that I was...

So I now don't know how to reproduce this - are you OK for me to check it in without coverage? ie just change the code to go "if one of the two special cases, log warning, else if any other failure log error" and no additional testcases.

(I have spent some time failing to figure out how to go about hitting this, without getting anywhere.  Didn't I recently link some other issue to this?  Sprout issues seem no longer to be visible to me so I can't tell any more!)

sebrexmetaswitch:
If your new UT isn't doing anything, then please remove it (no point running a duplicate UT).

Given that the change here is just to some logs, I'd be OK with you checking it in without coverage in principle.

However, if you can't repro it with a UT, I think that probably means you need to do some more thinking about this first to understand what the repro scenario is (I don't think your reasoning that it's not an error case can hold up if you don't know what the cause is).

dimbleby:
As I say, I spent some time yesterday trying and failing!  Any ideas?  

Also: any idea where all the sprout issues have gone - per previous note (again) I seem to remember finding a very relevant one recently and linking it to this, but that all seems to have disappeared.

dimbleby:
OK, so a SIPP repro where I simply send in an extra CANCEL really does reproduce this:

> 30-03-2017 09:37:39.161 UTC Warning pjsip:   tsx0x3581b78 Failed to send Request msg CANCEL/cseq=1 (tdta0x35824f0)! err=70013 (Invalid operation (PJ_EINVALIDOP))

I'll have a think about how the unit test is different - suggestions still welcome.  But at least I'm not barking up completely the wrong tree...

dimbleby:
OK, so I can reproduce the EEXISTS case live.  What you do is send in a CANCEL, catch the 487 and onwards CANCEL, and then cause a transport failure.  (Typically, in my case, by writing a faulty SIPP script).

Then sprout tries to send a CANCEL out on the outbound side of the call and pjsip goes "can't do it, transaction already exists".

So, do we have a sensible way of causing transport failures in UT?  

* If so then I'm certainly willing to put this into a unit test
* else, propose that we now understand this enough just to make the logging change

dimbleby:
OK, now with unit test that actually does hit it - should be good to go I think?

sebrexmetaswitch:
Yep, looks good.

Assuming `make full_test` passes, happy for this to be merged.

dimbleby:
great, thanks, and that's the last of my outstanding requests - thanks for your time!

PR #1617, opened 2016-11-29
[Reviewer: Ellie] Plugins create own blank templates

johadalin:
Adds the updated plugins submodule, and removes the old template files, including where they were used in the upload_blah.json scripts.


sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
```diff
@@ -57,10 +57,11 @@ then
   exit 2
 fi
 
-# Populate the file with defaults if not already present
+# Check the file is already present. If not, there may be problems with the plugin
```
eleanor-merry: What problems? (If I have to debug this in 6 months time this isn't a particularly useful prompt for what could have gone wrong)

Can you explain that the configuration file should either be written by the user, or it will be put in place by the config plugin in the case where it starts up, finds there's no value for the plugin's key, then writes the file using its default value. The file could be missing if the disk was full at this point, or if the user has deleted it, etc...

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
```diff
@@ -57,10 +57,11 @@ then
...
 if [[ ! -f $CONFPATH/$FILENAME ]]
 then
-  cp $CONFPATH/sample/$FILENAME $CONFPATH/$FILENAME
+  echo "No configuration file present. There may be a problem with the Clearwater Configuration Manager"
```
eleanor-merry: I don't think this is a useful error message - what's the user meant to do at this point? I'd just say something like 'No configuration file to upload - looking in FILENAME'

PR #1627, opened 2016-12-07
[Reviewer: AJH] Check separator is not NULL

sathiyan-sivathas:
...so that we don't crash when we get a route header in without a period.

This change fixed the UT failure I was seeing in the subscription sproutlet.

### DIFF
--- a/src/sproutletproxy.cpp
+++ b/src/sproutletproxy.cpp
@@ -329,7 +329,7 @@ bool SproutletProxy::does_uri_match_sproutlet(const pjsip_uri* uri,
     return false;
     // LCOV_EXCL_STOP
   }
-  
+
   // Now we know we have a SIP URI, cast to one.
   pjsip_sip_uri* sip_uri = (pjsip_sip_uri*)uri;
   std::list<std::string> possible_service_names = extract_possible_services(sip_uri);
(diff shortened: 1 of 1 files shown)


src/sproutletproxy.cpp
```diff
@@ -401,8 +401,11 @@ bool SproutletProxy::is_uri_local(const pjsip_uri* uri)
...
     char* sep = pj_strchr(&hostname, '.');
-    hostname.slen -= sep - hostname.ptr + 1;
-    hostname.ptr = sep + 1;
+    if (sep != NULL)
+    {
+      hostname.slen -= sep - hostname.ptr + 1;
+      hostname.ptr = sep + 1;
+    }
 
     return is_host_local(&hostname);
```
AlexHockey: If `sep` is NULL, we still end up calling `is_host_local` for a second time. This isn't a huge problem as the function is probably not very expensive, but it has potential to confuse the reader. I would be tempted to move this line to within the if block and end the function with a `return false`.

AlexHockey:
One comment inline. Have you checked that the other sprout UTs still work?

sathiyan-sivathas:
Markups applied. I've run the tests and they still pass.

PR #1630, opened 2016-12-09
Vary priority of bindings, for code coverage

dimbleby:
Merging #1597 damaged code coverage (which is very interesting... I guess the old bogus code was being optimised away and then nothing could tell that we weren't hitting it).

This tweak varies the priority in some unit tests so that we do now hit all code branches.  

NB Per the existing testcases, we don't actually check that the comparison gives the right answer!  I propose that fixing that is out of scope for this pull request.

Alex - sending to you since you merged #1597 for me.


AlexHockey:
Ha! I completely didn't think about that. Your fix looks good. I agree that improving the tests is out of scope (though I will add it to our tech debt article).

Issue #1555, opened 2016-10-04
Deregistration in GR deployment causes sprout to send 2 deregistration HTTP requests to homestead


sebrexmetaswitch:
#### Symptoms

For a deregistration in a 2 site deployment, sprout sends 2 deregistration HTTP requests to homestead.
The second of these correctly receives a 400 Bad Request, triggering Sprout to log an error:

`04-10-2016 14:25:50.967 UTC Error httpclient.cpp:615: cURL failure with cURL error code 0 (see man 3 libcurl-errors) and HTTP error code 400`
`04-10-2016 14:25:50.968 UTC Error hssconnection.cpp:617: Could not get subscriber data from HSS`
#### Impact

This is benign, but causes a Sprout error log and a scary-sounding "Fatal HTTP error encountered" log in SAS.
#### Release and environment

Latest dev release
#### Steps to reproduce

Spin up a 2-site deployment, and register then deregister a subscriber.

sebrexmetaswitch:
This is caused by the fact that the `write_to_store` function in registrar.cpp is called once per site, and that function is what triggers the HTTP request

sebrexmetaswitch:
loosely related to #552

MatMeredith:
Fixed by PR #1635

PR #1635, opened 2016-12-13
[Reviewer: Graeme] Sprout issue 1555 -- Sprout sends multiple deregistration requests to Homestead 

MatMeredith:
Issue #1555 

Fix: Only dereg at HSS once -- not once per local / remote SDM.

- Enhanced UTs to police the number of calls to HSSConnection::update_registration_state (successfully replicating the bug)
- Fix Sprout
- Live test with clearwater_live_test

FTR: The approach of using the MockHSSConnection object within the FakeHSSConnection as a way of observing calls has already been discussed with @AlexHockey...


src/ut/mock_hss_connection.h
```diff
@@ -70,6 +70,17 @@ class MockHSSConnection : public HSSConnection
                         std::map<std::string, Ifcs >& service_profiles,
                         std::vector<std::string>& associated_uris,
                         SAS::TrailId trail));
+  MOCK_METHOD9(update_registration_state,
+                HTTPCode(const std::string& public_user_identity,
```
graemerobertson: Misalignment

src/ut/registrar_test.cpp
```diff
@@ -461,6 +483,10 @@ class RegistrarTest : public SipTest
     msg._contact = "*";
     msg._contact_instance = "";
     msg._contact_params = "";
+    EXPECT_CALL(*_hss_connection_observer,
+                update_registration_state("sip:6505550231@homedomain", _, HSSConnection::REG, _, _, _, _, _, _)).WillOnce(Return(HTTP_OK));
```
graemerobertson: I can't work out why this happens? Why not just the de-reg call below? Do you know?

graemerobertson: Ah yes, of course, I knew that - thanks.

(The issue you're referring to is https://github.com/Metaswitch/sprout/issues/552)

PR #1638, opened 2016-12-14
[REVIEWER: Ellie] Add a stat to track forked INVITEs due to multiple bindings.

NigelThorpe:
This is supposed to track the number of additional INVITEs sent because a target has multiple bindings, e.g. if a user has 3 registered devices then the stat is incremented twice per call, and it is not incremented if a user has only 1 registered device.

### DIFF
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -99,11 +99,12 @@ SCSCFSproutlet::SCSCFSproutlet(const std::string& scscf_name,
                                                                  "1.2.826.0.1.1578918.9.3.32");
   _invites_cancelled_after_1xx_tbl = SNMP::CounterTable::create("invites_cancelled_after_1xx",
                                                                 "1.2.826.0.1.1578918.9.3.33");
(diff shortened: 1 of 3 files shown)


src/scscfsproutlet.cpp
```diff
@@ -1662,6 +1664,11 @@ void SCSCFSproutletTsx::route_to_ue_bindings(pjsip_msg* req)
       // in case we get a 430 Flow Failed response.
       int fork_id = send_request(to_send);
       _target_bindings.insert(std::make_pair(fork_id, targets[ii].binding_id));
+
+      if ((_req_type == PJSIP_INVITE_METHOD) && (ii != 0))
```
eleanor-merry: Can you comment here what this stat actually is?

NigelThorpe: Done:
```
      if ((_req_type == PJSIP_INVITE_METHOD) && (ii != 0))
      {
        // Increment stat tracking the number of additional INVITEs generated
        // due to there being multiple registered targets.
        _scscf->_forked_invite_tbl->increment();
      }
```

Issue #1600, opened 2016-11-19
AKAv2 authentication is not thread-safe


rkd-msw:
https://github.com/Metaswitch/sprout/blob/190cf993a5629fffeb607d69ac3f49e9e4cca21d/src/authentication.cpp#L496 calls the OpenSSL HMAC function, but leaves the last two arguments as NULL. According to https://linux.die.net/man/3/hmac, this places the result in a static array, which is presumably not thread-safe.

#### Symptoms

None (spotted while code-reading as part of #1599).

#### Impact

AKAv2 authentications would fail, if two calls to HMAC overlapped and the static array got scribbled.

#### Release and environment 

Quilava release, no environment (spotted by code-reading).

#### Steps to reproduce

Attempt AKAv2 authentication under valgrind/helgrind (?).

PR #1639, opened 2016-12-15
[Reviewer: Matt] Make AKAv2 Auth thread safe

MatMeredith:
Fix to issue #1600

### DIFF
--- a/src/authentication.cpp
+++ b/src/authentication.cpp
@@ -440,7 +440,7 @@ void create_challenge(pjsip_digest_credential* credentials,
       // AKA version defaults to 1, for back-compatibility with pre-AKAv2
       // Homestead versions.
       int akaversion = 1;
-      
+
       JSON_SAFE_GET_STRING_MEMBER(aka, "challenge", nonce);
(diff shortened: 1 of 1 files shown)


src/authentication.cpp
```diff
@@ -494,12 +494,13 @@ void create_challenge(pjsip_digest_credential* credentials,
...
         unsigned int digest_len;
+        unsigned char hmac[EVP_MAX_MD_SIZE];
         unsigned char* digest = HMAC(EVP_md5(),
                                      (unsigned char*)joined.data(),
                                      joined.size(),
                                      (unsigned char*)(formality.data()),
                                      formality.size(),
-                                     NULL,
+                                     hmac,
                                      &digest_len);
```
mirw: Is it right for digest_len to be passed in uninitialized, or does it need to be set to EVP_MAX_MD_SIZE?  (Also, do we get any compiler warnings from not initializing it?)

MatMeredith: No compiler warnings and no suggestion here (https://linux.die.net/man/3/hmac) that it needs to be initialised.   OK?

mirw: OK.

PR #1652, opened 2016-12-19
[Reviewer: Graeme] Allow Mangelwurzel to handle REGISTER requests

rkd-msw:
This:

* gets Mangelwurzel to respond with 200 OK to REGISTER requests, allowing us to take standard IFCs and swap in Mangelwurzel's URI
* gets Mangelwurzel to recognise itself in the Request-URI as well as the top Route header, as that's how REGISTER requests to app servers are formed (this required changing a lot of indentation, sorry)
* removes two checks in main.cpp that don't make sense (one is that you must have a Homestead unless you set `authentication=N`, which doesn't make sense for app-server-only nodes, and the other is that you must have a Homestead URL if you have a Homer URL, which is fine for Mangelwurzel but doesn't make sense for MMTel AS)
    * there is a check that you must have a Homestead URL if you're an I-CSCF or S-CSCF, so I think these checks are redundant

### DIFF
--- a/docs/Mangelwurzel.md
+++ b/docs/Mangelwurzel.md
@@ -8,7 +8,7 @@ Mangelwurzel is installed on top of sprout-base. It can be installed through the
 
 ## Invocation and configuration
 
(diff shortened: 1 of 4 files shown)


graemerobertson:
Looks good. Are you allowed to merge or do I have to merge for you?

rkd-msw:
I'll merge once code is unfrozen.

rkd-msw:
I realised I hadn't made corresponding UT changes, but I now have.

PR #1654, opened 2016-12-20
Sproutlet refactor integration

AlexHockey:
This PR contains the code to plumb the new authentication, registrar and subscription sproutlets together with the existing S-CSCF proxy sproutlet. Please can you review:

* @sathiyan-sivathas please can you be the primary reviewer
* @eleanor-merry please can you cast you eyes over this and check it matched what we discussed. 

I have checked that all the UTs run, as do all the tests in clearwater-live-test (except the B2BUA test which I think might be just broken - though I'll check this when I'm back after Christmas).

You'll see that I have not removed the ability for the sproutlets to specify what ports and URis they want to register. Instead the SCSCF plugin just does not configure these. This is for two reasons:

1. Removing the URi and port makes the sproutlets really awkward to UT in isolation. 
2. One of the benefits of sproutlets is the ability to re-use them. Having sproutlets that can't specify ports and URIs isn't really in keeping with this vision.

### DIFF
--- a/src/authenticationsproutlet.cpp
+++ b/src/authenticationsproutlet.cpp
@@ -1199,10 +1199,12 @@ void AuthenticationSproutletTsx::on_rx_initial_request(pjsip_msg* req)
 
 void AuthenticationSproutletTsx::forward_request(pjsip_msg* req)
 {
+  const pjsip_route_hdr* route = route_hdr();
+  pjsip_sip_uri* base_uri = (pjsip_sip_uri*)(route ? route->name_addr.uri : nullptr);
   pjsip_sip_uri* uri =
     (pjsip_sip_uri*)get_uri_for_service(_sproutlet->_next_hop_service,
                                         get_pool(req),
-                                        (pjsip_sip_uri*)route_hdr()->name_addr.uri);
+                                        base_uri);
   PJUtils::add_top_route_header(req, uri, get_pool(req));
   send_request(req);
 }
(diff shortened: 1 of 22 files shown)


AlexHockey:
If either of you are interested in what it looks like if I do remove the ports and URIs, take a look at https://github.com/Metaswitch/sprout/pull/1655.

src/forwardingsproutlet.cpp
```diff
@@ -0,0 +1,55 @@
...
+ * under which the OpenSSL Project distributes the OpenSSL toolkit software,
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+
+#include "forwardingsproutlet.h"
+#include "pjutils.h"
+
+ForwardingSproutletTsx::ForwardingSproutletTsx(SproutletTsxHelper* helper,
+                                               const std::string& upstream_service_name) :
+  SproutletTsx(helper), _upstream_service_name(upstream_service_name)
```
sathiyan-sivathas: Can we have this split over separate lines?

src/authenticationsproutlet.cpp
```diff
@@ -111,24 +115,30 @@ bool AuthenticationSproutlet::init()
 
   params.options = PJSIP_AUTH_SRV_IS_PROXY;
   status = pjsip_auth_srv_init2(stack_data.pool, &_auth_srv_proxy, &params);
-
   return (status == PJ_SUCCESS);
```
sathiyan-sivathas: Related to a comment you made on my PR.

SCSCFPlugin::load() returns false if any of the sproutlets fail to load. The registrar and scscf sproutlet's init functions have error logs if they fail, maybe we should have one here too for debugging purposes?

AlexHockey:
@eleanor-merry any idea when you'll get round to reviewing this?

eleanor-merry:
LGTM

PR #1672, opened 2017-01-09
[Reviewer: EM] Re-added listen_port to config options

GAB-MS:
listen_port is an old configuration option which specifies a port to be used for all sproutlets. This PR re-adds the option, as it is currently the easiest way to specify a port for a node only running a single, non-default sproutlet.


eleanor-merry:
LGTM

PR #1673, opened 2017-01-09
[Reviewer: EM] Move listen_port from sproutlet_options to main

GAB-MS:
Following from pull request #1672, this removes the listen_port functionality from sproutlet_options so that it does not affect default sproutlets. This is replaced by a block in main which, if listen_port is set, adds the port to the list of ports to be configured. This allows for listen_port to be used to specify a port to be configured for non-default sproutlets without affecting any default sproutlets.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1465,6 +1465,12 @@ int main(int argc, char* argv[])
   std::vector<std::string> sproutlet_uris;
   SPROUTLET_MACRO(SPROUTLET_VERIFY_OPTIONS)
 
+  if (opt.listen_port != 0)
+  {
+    TRC_INFO("Opening port %d for non-default sproutlets", opt.listen_port);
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -1465,6 +1465,12 @@ int main(int argc, char* argv[])
   std::vector<std::string> sproutlet_uris;
   SPROUTLET_MACRO(SPROUTLET_VERIFY_OPTIONS)
 
+  if (opt.listen_port != 0)
```
eleanor-merry: Rather than insert it here, you might as well insert it when you parse the option (e.g. https://github.com/Metaswitch/sprout/blob/dev/src/main.cpp#L1161 - and get rid of options->listen_port as well)

GAB-MS:
I've implemented the change suggested by Ellie. It's been re-tested and confirmed to work. Merging in now

Issue #1687, opened 2017-01-18
Cases where pjsip_tx_send_msg return status should not be used


johadalin:
Following on from #1607. 
Calls to `pjsip_tsx_send_msg` will end up calling `on_tsx_state` as well as returning a pj_status. as #1607 highlighted, if we take action based on a returned PJ_FAILURE, we may end up with conflicts in behaviour with the result of the call to `on_tsx_state`. 

In `BasicProxy::UACTsx::retry_request`, we still check and take action on the returned status. While we need to investigate the risk of potential window conditions this might open up, it is notably lower risk than #1607, and so is being split out to be covered by this issue. 
At time of writing, the only other location in BasicProxy where we check on the return code of `pjsip_tsx_send_msg` is in `BasicProxy::UACTsx::cancel_pending_tsx`, where we simply write a log if the return code is not PJ_SUCCESS.

PR #1688, opened 2017-01-18
[Reviewer: Matt] Ignore PJ_FAILURE from pj_tsx_send_msg in send_request

johadalin:
As discussed, this is the small fix for send_request.

Just want to run this quickly by you in case it isn't what you were expecting. If you're happy i'll merge down and talk to Rob about system testing it properly. In the process of building it for a quick chef deployment test to make sure it isn't somehow completely broken.

I have opened #1687  to cover further checking in e.g. `retry_request`, to make sure that there aren't further window conditions.
As the issue states, the only location other than `retry_request` where we check the return value of `pjsip_tsx_send_msg` results in us writing a log if it fails, and so is not a serious concern; `retry_request` will want investigation, but as discussed, we are much less likely to hit any error there, as it is already in a retry branch.

### DIFF
--- a/src/basicproxy.cpp
+++ b/src/basicproxy.cpp
@@ -1697,15 +1697,13 @@ void BasicProxy::UACTsx::send_request()
       {
         start_timer_c();
       }
-      else
-      {
-        // We do not want to take any action on a failure returned from
-        // pjsip_tsx_send_msg, as it will have also triggered a call into
-        // on_tsx_state. In the event of failure, this will, or already has
-        // cause us to call into retry_request; we do not want to call into
-        // on_client_not_responding below.
-        status = PJ_SUCCESS;
-      }
+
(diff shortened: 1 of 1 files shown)


src/basicproxy.cpp
```diff
@@ -1697,6 +1697,15 @@ void BasicProxy::UACTsx::send_request()
...
         start_timer_c();
       }
+      else
+      {
+        // We do not want to take any action on a failure returned from
+        // pjsip_tsx_send_msg, as it will have also triggered a call into
+        // on_tsx_state. In the event of failure, this will, or already has
+        // cause us to call into retry_request; we do not want to call into
+        // on_client_not_responding below.
+        status = PJ_SUCCESS;
```
mirw: I don't think this needs to be in an else block... and it's a bit unintuitive because the parent "if" is "if success && INVITE"

PR #1697, opened 2017-01-24
Fix SCSCF plugin to correctly configure reg-max-expires

AlexHockey:
It looks like when refactoring the SCSCF modules into sproutlets, we broken the reg-max-expires config option that controls how log a device can register for. Please can you review my (simple) fix. 

I ran into this problem when running the P-CSCF-emulating stress tests (the tests try to register for 1hr, but were only registering for 5mins so some calls didn't work). I've checked that with this fix (and an appropriate value for reg_max_expires) the tests now work.


sathiyan-sivathas:
Oops, I think that was me. Fix looks good.

PR #1698, opened 2017-01-24
[Reviewer: Ellie] Have the init.d abort remove pidfile if it fails

### LINKED ISSUE #1326: Sprout might never restart (race condition)
There appears to be the possibility of a rare race condition in Sprout. 

Due to the logic of [the Sprout monit script](https://github.com/Metaswitch/sprout/blob/51024458ddc5391f5c29c7be2eb882f33c5ab218/sprout-base.root/usr/share/clearwater/infrastructure/scripts/sprout.monit), if sprout is quiesced/stopped, and another process starts with the same PID as it had just after this, before monit has had time to check, we do not revive.

The first check in this script just checks that the PID file exists, and that the PID is a running process, not that that process is sprout. 
As such, in the condition that another process has taken the PID, we do not attempt to start sprout.

Further to this, when the poll scripts fail, the abort commands run fail silently if they cannot find the right PID from within the subset of processes named Sprout. As Sprout is not running, it will never find the PID. Because of this, the PID file is also never cleaned up.

This will be a possible issue in all of our monit files

johadalin:
If the abort command fails to kill a process, we attempt to remove the pidfile. 
This is to combine with https://github.com/Metaswitch/clearwater-monit/pull/47, to allow us to resolve #1326 . 
If there is no process running with the pid provided in the pidfile, and matching the expected process name, the abort will fail. We already have code in place to kill rogue processes that match on name, but not on pid. Removing the pidfile allows monit to realise the process is dead if we hit the referenced race condition, and kick off starting the process up again.

If you're happy with this I will go around and make the same change to the init.d scripts for our other processes.

### DIFF
--- a/debian/bono.init.d
+++ b/debian/bono.init.d
@@ -254,6 +254,11 @@ do_abort()
         #   other if a failure occurred
         start-stop-daemon --stop --quiet --retry=ABRT/60/KILL/5 --user $NAME --pidfile $PIDFILE --name $EXECNAME
         RETVAL="$?"
+        # If the abort failed, it may be because the PID in PIDFILE doesn't match the right process
+        # In this window condition, we may not recover, so remove the PIDFILE to get it running
+        if [ $RETVAL != 0 ]; then
+          rm -f $PIDFILE
+        fi
         return "$RETVAL"
 }
 
--- a/debian/restund.init.d
+++ b/debian/restund.init.d
@@ -145,9 +145,12 @@ do_abort()
         #   other if a failure occurred
         start-stop-daemon --stop --quiet --retry=ABRT/60/KILL/5 --name $NAME
         RETVAL="$?"
+        # If the abort failed, it may be because the PID in PIDFILE doesn't match the right process
+        # In this window condition, we may not recover, so remove the PIDFILE to get it running
+        if [ $RETVAL != 0 ]; then
+          rm -f $PIDFILE
+        fi
         [ "$RETVAL" = 2 ] && return 2
-        # Many daemons don't delete their pidfiles when they exit.
-        #rm -f $PIDFILE
         return "$RETVAL"
 }
 
--- a/debian/restund.init.d
+++ b/debian/restund.init.d
@@ -145,11 +145,6 @@ do_abort()
         #   other if a failure occurred
         start-stop-daemon --stop --quiet --retry=ABRT/60/KILL/5 --name $NAME
         RETVAL="$?"
-        # If the abort failed, it may be because the PID in PIDFILE doesn't match the right process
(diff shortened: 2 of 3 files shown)


johadalin:
I think a full list of init scripts that should be modified is as below:

astaire
bono
chronos
clearwater-cluster-manager
clearwater-config-manager
clearwater-etcd
clearwater-queue-manager
clearwater-snmp-alarm-agent
ellis
homer
homestead
homestead-prov
ralf
restund
sprout

eleanor-merry:
Sure. I think there are some more init scripts though. Have you checked in clearwater-infrastructure?

johadalin:
As discussed, i'm going to move the step removing the pidfile up into `do_abort` section, so that both `abort` and `abort-restart` get us out of the error condition where we have a pidfile pointing to the wrong process.
I'm also going to check the init scripts which don't have an `abort` entry to see what makes sense, checking what Monit uses to abort the processes in question in these cases.

johadalin:
https://github.com/Metaswitch/crest/pull/331
https://github.com/Metaswitch/homestead/pull/426
https://github.com/Metaswitch/astaire/pull/80
https://github.com/Metaswitch/chronos/pull/339
https://github.com/Metaswitch/ralf/pull/260
https://github.com/Metaswitch/ellis/pull/203
https://github.com/Metaswitch/clearwater-etcd/pull/392

johadalin:
So i think other areas we could improve are the following:

* This all adds pidfile removal to the abort commands. We may potentially want the same for stop commands, to make sure we don't end up in the weird monit window condition.
* Some processes only have stop, rather than abort, in their init scripts. I think this is just:
    * Clearwater-etcd
    * Clearwater-diags-monitor
    * Cassandra
    * other things we don't modify (nginx, mysql)
* Some of our processes do not have health-checking beyond the monit `check process` call. These will not currently be able to get out of the window condition described in #1326 . If they hit this, will not recover, as there is no other means to detect the process is not running.

I think this is all ok, as these changes should at least make it less likely that we hit the problem, and is overall a better position to be in.

johadalin:
Linking this here for future references:
https://github.com/Metaswitch/crest/pull/334

As part of this work, I added a sanity check to kill any process matching our `DAEMON`, but not matching our pid, as was already present in e.g. sprout. This Is to catch any processes we may leak. 
Some of our processes however, run the same thing in a different directory, as is the case in crest. In this case, the sanity check will lead to a cycle of process death, with each killing the other, and monit attempting to start them again.

If in future work we want to have multiple instances of any of our processes running on the same node, this issue may re-appear. In that case, similar logic as in the linked crest PR will need to be applied to differentiate between the different processes.

PR #1707, opened 2017-01-30
Fix Sprout analytics logs

MatMeredith:
Three parts to this...
- Fix REGISTER audit log issues introduced by https://github.com/Metaswitch/sprout/pull/1585/files 
  -  REGISTER audit log expires value should be the time interval (e.g. 300s).  The PR above inadvertently changed this to be the epoch time at which the register will expire
  - de-REGISTER should be logged as a REGISTER with expires=0.  The PR above inadvertently changed this so that it was the previous expiry time (i.e. the logs for a REGISTER and subsequent de-REGISTER were identical).

- Add UTs so that we don't break this again so easily in future.

- Adding UTs for Subscriptions to match those for Registers reveals that we are writing Subscription audit logs on both the primary and backup SDM.   This is inconsistent with what we do for Registers so fix it.

### DIFF
--- a/src/subscription.cpp
+++ b/src/subscription.cpp
@@ -139,7 +139,7 @@ SubscriberDataManager::AoRPair* write_subscriptions_to_store(
                                                               ///<backup stores to read from if no entry in store and no backup data
                    SAS::TrailId trail,                        ///<SAS trail
                    std::string public_id,                     ///
-                   bool send_ok,                              ///<Should we create an OK
+                   bool is_primary,                           ///Is this the primary SDM we are writing to?
                    ACR* acr,                                  ///
                    std::deque<std::string> ccfs,              ///
                    std::deque<std::string> ecfs)              ///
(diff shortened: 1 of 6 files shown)


src/subscription.cpp
```diff
@@ -370,7 +370,7 @@ SubscriberDataManager::AoRPair* write_subscriptions_to_store(
   }
   while (set_rc == Store::DATA_CONTENTION);
 
-  if (analytics != NULL)
+  if ((analytics != NULL) && is_primary)
```
eleanor-merry: Can is_primary be in brackets too?

src/ut/mock_analytics_logger.h
```diff
@@ -0,0 +1,77 @@
+/**
+ * @file mock_impi_store.h
```
eleanor-merry: Wrong name and date?

rkd-msw:
As Ellie's comments were basically cosmetic, I'm going to merge this - that means it should show up in tomorrow morning's build and unblock testing. Sorry if that causes problems!

PR #1715, opened 2017-02-01
[Reviewer: Ellie] BGCF phone number routing

sebrexmetaswitch:
Fallback to wildcard domain route if there is one, if a phone number with routing number parameter doesn't match any configured phone number routes.

This ensures that the URIs tel:xxx and tel:yyy;rn=xxx will be routed the same.

There is a readthedocs change too in https://github.com/Metaswitch/clearwater-readthedocs/pull/241

### Testing
 - UTs added and verified they failed before the change and pass after.


eleanor-merry:
LGTM

Issue #1667, opened 2017-01-04
'Warning scscfsproutlet.cpp:1856: No charging role in Route header, assume originating' when using a sprout_rr_level other than the default


rkd-msw:
#### Symptoms

I saw the following log on every SIP request while running stress:

```
04-01-2017 17:32:29.754 UTC Warning scscfsproutlet.cpp:1856: No charging role in Route header, assume originating
04-01-2017 17:32:29.926 UTC Warning scscfsproutlet.cpp:1856: No charging role in Route header, assume originating
04-01-2017 17:32:29.927 UTC Warning scscfsproutlet.cpp:1856: No charging role in Route header, assume originating
04-01-2017 17:32:30.441 UTC Warning scscfsproutlet.cpp:1856: No charging role in Route header, assume originating
```

Looking at the Route headers for an in-dialog request, they look like this:

```
Route:  <sip:scscf.sprout.stress.clearwater.test;transport=TCP;lr;service=scscf;billing-role=charge-orig>, <sip:mangelwurzel.infra.clearwater.test:5060;transport=TCp;lr;dialog>, <sip:scscf.sprout.stress.clearwater.test;transport=TCP;lr;service=scscf>, <sip:scscf.sprout.stress.clearwater.test;transport=TCP;lr;service=scscf>, <sip:mangelwurzel.infra.clearwater.test:5060;transport=TCp;lr;dialog>, <sip:scscf.sprout.stress.clearwater.test;transport=TCP;lr;service=scscf;billing-role=charge-term>,<sip:10.225.166.12:5082;transport=TCP;lr>
```

Specifically, the Route headers added at the end of originating processing and at the start of terminating processing don't have a billing-role parameter, causing this error. 

#### Impact

This causes log spam at the rate of 115 bytes per in-dialog message, which will increase the size/unwieldiness of diagnostics packages, unnecessarily worry users who see it, and make it harder to spot genuine errors.

This will also presumably cause billing information to be wrong, as when the S-CSCF is invoked before a terminating AS, it will assume its billing role is originating.

#### Release and environment 

10.0.0-161216-~255.00.0-161216.020317

I have `sprout_rr_level=pcscf,icscf,as` set for this testing, in order to test the supported features that are most likely to impact performance.

#### Steps to reproduce

Set `sprout_rr_level=pcscf,icscf,as` in `shared_config`, make a call, and check the log files.

sebrexmetaswitch:
[...]

PR #1716, opened 2017-02-02
Use billing-role none to suppress unwanted ACRs

MatMeredith:
This is a fix to issue #1667.  
- Previous log wasn't harmless (good spot @rkd).   We were actually generating billing-records for interim hops where we shouldn't have been, and furthermore we were incorrectly marking them as originating when some where in fact terminating.
- Fix is, when record-routing, to explicitly make routes as charge-none if we don't want to generate billing records for that hop, and then when processing an in-dialog request not generate an ACR if the top route header indicates charge-none.   We could have just done this when there was no billing-role specified, but this way is more explicit and allows us to continue to generate a warning if we get a request with a route header that doesn't correctly identify a billing-role.

UTs test that we correctly use charge-none in record-route headers and I've tested live to verify that we now send the correct ACRs to Ralf in a couple of different scenarios (simple call, and call diversion 408 from clearwater-live-test).

I've also raised an issue in the dirt-doc / with CNE tracking the fact that we don't have any regression tests that verify that we generate the correct Rf requests in different calling scenarios.

### DIFF
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -547,13 +547,24 @@ void SCSCFSproutletTsx::on_rx_in_dialog_request(pjsip_msg* req)
   _record_routed = true;
   _se_helper.process_request(req, get_pool(req), trail());
 
-  // Create an ACR for this request and pass the request to it.
-  _in_dialog_acr = _scscf->get_acr(trail(),
-                         ACR::CALLING_PARTY,
-                         get_billing_role());
+  ACR::NodeRole billing_role;
+  bool bill_request = get_billing_role(billing_role);
 
-  // @TODO - request timestamp???
-  get_acr()->rx_request(req);
+  if (bill_request)
+  {
+    // Create an ACR for this request and pass the request to it.
+    _in_dialog_acr = _scscf->get_acr(
+      trail(),
+      ACR::CALLING_PARTY,
+      billing_role);
+
+    // @TODO - request timestamp???
+    get_acr()->rx_request(req);
+  }
+  else
+  {
+    _in_dialog_acr = NULL;
+  }
 
   send_request(req);
 }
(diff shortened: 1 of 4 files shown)


src/scscfsproutlet.cpp
```diff
@@ -1769,9 +1784,13 @@ bool SCSCFSproutletTsx::lookup_ifcs(std::string public_id, Ifcs& ifcs)
 }
 
 
-/// Add the S-CSCF sproutlet into a dialog.  The parameter passed will
-/// be attached to the Record-Route and can be used to recover the billing
-/// role that is in use on subsequent in-dialog messages.
+/// Add the S-CSCF sproutlet into a dialog.
+///
+/// Params:
+/// - billing_rr:   set to true if this is a hop where future in-dialog requests
```
graemerobertson: Worth describing the msg param for completeness?

src/scscfsproutlet.cpp
```diff
@@ -1799,42 +1818,57 @@ void SCSCFSproutletTsx::add_to_dialog(pjsip_msg* msg,
...
+  // previous call to this function within this transaction).  Therefore the
+  // Record-Route header we added then must be present (and must be the top
+  // such header).
+  assert(rr != NULL);
+
+  pjsip_sip_uri* uri = (pjsip_sip_uri*)rr->name_addr.uri;
+  pjsip_param* param = pjsip_param_find(&uri->other_param,
+                                        &STR_BILLING_ROLE);
+
+  // Work out what our billing-role shoudld be set to for this hop.
```
graemerobertson: *should

src/scscfsproutlet.cpp
```diff
@@ -1799,42 +1818,57 @@ void SCSCFSproutletTsx::add_to_dialog(pjsip_msg* msg,
...
+  else if (billing_role == ACR::NODE_ROLE_ORIGINATING)
   {
-    // We've records routed before (either earlier in this function or in a
-    // previous call to this function within this transaction).  Therefore the
-    // Record-Route header we added then must be present (and must be the top
-    // such header).
-    assert(rr != NULL);
+    charging_role = &STR_CHARGE_ORIG;
+  }
+  else
```
graemerobertson: I understand that this must mean `billing_role` is `ACR::NODE_ROLE_TERMINATING` but I can't really see the disadvantage of checking and it feels slightly more maintainable going forward.

src/scscfsproutlet.cpp
```diff
@@ -1799,42 +1818,57 @@ void SCSCFSproutletTsx::add_to_dialog(pjsip_msg* msg,
...
+  }
+  else if ((billing_rr) && !pj_strcmp(&param->value, &STR_CHARGE_NONE))
+  {
+    // We had previously set the billing-role for this hop to NONE but have now
+    // decided that we should be billing it. E.g. we first treated this as
+    // just a standard AS hop, but have now decided that it is the final
+    // terminating hop.  Update the billing role.
+    pj_strdup(pool, &param->value, charging_role);
   }
 }
```
graemerobertson: I found the use of the variables charging_role and billing_role (which both represented identical concepts, just in a slightly different form?) pretty confusing! Would something like `acr_billing_role` and `pjsip_billing_role` be better? Possibly not, that doesn't feel very nice either.

src/scscfsproutlet.cpp
```diff
@@ -547,13 +547,28 @@ void SCSCFSproutletTsx::on_rx_in_dialog_request(pjsip_msg* req)
...
+  ACR::NodeRole billing_role;
+  bool bill_request = get_billing_role(billing_role);
 
-  // @TODO - request timestamp???
-  get_acr()->rx_request(req);
+  if (bill_request)
+  {
+    // Create an ACR for this request and pass the request to it.
+    _in_dialog_acr = _scscf->get_acr(
+      trail(),
```
MatMeredith: Good point.   It was like this originally because one of the parameters was too long to fit aligned to the right of "get_acr(".   Even then it probably wasn't right though...

src/scscfsproutlet.cpp
MatMeredith: Are you suggesting turning the else into an else if billing_role == TERMINATING?   And then presumably having an else branch saying that "We really shouldn't hit this, but if we do, assume we are [e.g.] TERMINATING" and making a warning log?

src/scscfsproutlet.cpp
MatMeredith: Yep -- like the suggestion.   I've also renamed billing_rr -> bill_this_hop since I always found that name confusing too.   I don't know if "hop" is the right word for what we're talking about here or whether there is an official IMS or SIP word for this thing?

src/scscfsproutlet.cpp
graemerobertson: Yes, probably. I'm not even sure I care about adding the else branch....

src/scscfsproutlet.cpp
graemerobertson: Hop sounds fine to me.

src/scscfsproutlet.cpp
MatMeredith: In which case if we would have hit it we just continue on and de-reference a null pointer?    How about I just add an assert to the existing else branch?  I'll do that unless you object.

src/scscfsproutlet.cpp
```diff
@@ -1850,6 +1850,7 @@ void SCSCFSproutletTsx::add_to_dialog(pjsip_msg* msg,
   }
   else
   {
+    assert(acr_billing_role == ACR::NODE_ROLE_TERMINATING);
```
rkd-msw: Generally we try and avoid asserts - if we hit this case, I think Sprout will now crash, whereas previously it would've kind of worked. TRC_ERROR would be preferable, I think.

MatMeredith: @g should have just left me leave my code as it was ;-)   Fair enough -- I'll change to TRC_ERROR then.

MatMeredith: (Although that wasn't meant to be a criticism -- the assert was my idea...)

graemerobertson: Pft, I'm not convinced asserting is wrong. If we somehow hit it we don't want to risk not noticing - it'll be indicative of a pretty fundamental code bug. I guess we pick up error logs reasonably reliably.

PR #1717, opened 2017-02-02
[Reviewer: AJH] Sproutlet routing robustness

### LINKED ISSUE #1696: Sprout S-CSCF cannot handle Route header URIs with "scscf" in the user-part
#### Symptoms

If sprout receives an INVITE (or any request actually) where the top Route header contains "scscf" in the user part, this causes a routing loop where the authentication sproutlet is invoked repeatedly until the max-forwards count is exceeded. The request is then rejected with "483 Too Many Hops".

#### Impact

Sprout cannot process calls. 

#### Release and environment 

release-115

#### Steps to reproduce

The easiest way to reproduce is use the P-CSCF emulating stress test scripts in https://github.com/Metaswitch/clearwater-perf-utils as of commit 89c5b3bdcb68a8d44a440117639f131ee1f39226. These produce URLs of the form that sprout cannot handle.

sathiyan-sivathas:
Fixes #1696 

As discussed in the whiteboard, this change:
- Strips the user part off the header that caused us to route to sprout before building a route header to route between sproutlets.
- Specifies a service parameter in the odi route header added by the scscf-proxy.

This fixes the bug where we hit the max-forwards limit if a sproutlet is specified the route header and also makes sure that the scscf-proxy can still see the odi token for requests that come back from app servers.

UTs pass, but I need to check the live tests before I merge.

### DIFF
--- a/src/forwardingsproutlet.cpp
+++ b/src/forwardingsproutlet.cpp
@@ -47,10 +47,6 @@ void ForwardingSproutletTsx::forward_request(pjsip_msg* req)
 {
   const pjsip_route_hdr* route = route_hdr();
   pjsip_sip_uri* base_uri = (pjsip_sip_uri*)(route ? route->name_addr.uri : nullptr);
-  if (base_uri != NULL)
-  {
-    base_uri->user = pj_str(const_cast<char*>(""));
-  }
   pjsip_sip_uri* uri =
     (pjsip_sip_uri*)get_uri_for_service(_upstream_service_name,
                                         get_pool(req),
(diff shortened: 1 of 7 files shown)


src/forwardingsproutlet.cpp
```diff
@@ -47,6 +47,10 @@ void ForwardingSproutletTsx::forward_request(pjsip_msg* req)
 {
   const pjsip_route_hdr* route = route_hdr();
   pjsip_sip_uri* base_uri = (pjsip_sip_uri*)(route ? route->name_addr.uri : nullptr);
+  if (base_uri != NULL)
+  {
+    base_uri->user = pj_str(const_cast<char*>(""));
+  }
```
AlexHockey: I think `get_uri_for_service` should do this. If you rely on the caller you run the risk of another sproutlet having the same bug. Please also update the API comments to explain that the function will strip the user-part.

AlexHockey:
One minor comment on your code. 

Stepping back though, in a way this is a bug fix. When fixing a bug it's usually a good idea to write a unit test that reproduces it. That way you're sure you've fixed the problem, and you've get regression tests to stop it reoccurring. Is it worth adding one here (that a genuine question not a rhetorical one ;-)

Issue #981, opened 2015-03-16
Sprout crash at high traffic load when configured with a CFS TAS.


plfrancis:
When attempting to run 100+ CPS on a single Sprout instance configured with a CFS TAS, Sprout crashes when trying to build an ACR. This behavior was triggered by enabling billing headers in the TAS. At first glance it would appear that  store_charging_addresses() may need reentrancy protection for modifications to the _ccfs and _ecfs lists. The following is the backtrace for the offending thread (a diagnostic bundle was also collected).

```
(gdb) bt
#0  0x00007fdd0b480425 in __GI_raise (sig=<optimized out>) at ../nptl/sysdeps/unix/sysv/linux/raise.c:64
#1  0x00007fdd0b483b8b in __GI_abort () at abort.c:91
#2  0x00000000005a8e7c in exception_handler (sig=6) at main.cpp:841
#3  <signal handler called>
#4  0x00007fdd0b480425 in __GI_raise (sig=<optimized out>) at ../nptl/sysdeps/unix/sysv/linux/raise.c:64
#5  0x00007fdd0b483b8b in __GI_abort () at abort.c:91
#6  0x00007fdd0b4be39e in __libc_message (do_abort=2, fmt=0x7fdd0b5c8748 "*** glibc detected *** %s: %s: 0x%s ***\n")
    at ../sysdeps/unix/sysv/linux/libc_fatal.c:201
#7  0x00007fdd0b4c8b96 in malloc_printerr (action=3, str=0x7fdd0b5c8938 "double free or corruption (fasttop)", ptr=<optimized out>) at malloc.c:5039
#8  0x00000000004e6177 in _M_dispose (__a=..., this=<optimized out>) at /usr/include/c++/4.6/bits/basic_string.h:244
#9  ~basic_string (this=0x7fdc9467bf00, __in_chrg=<optimized out>) at /usr/include/c++/4.6/bits/basic_string.h:534
#10 ~_List_node (this=0x7fdc9467bef0, __in_chrg=<optimized out>) at /usr/include/c++/4.6/bits/stl_list.h:105
#11 destroy (__p=0x7fdc9467bef0, this=<optimized out>) at /usr/include/c++/4.6/ext/new_allocator.h:118
#12 std::_List_base<std::string, std::allocator<std::string> >::_M_clear (this=0x7fdc946797f0) at /usr/include/c++/4.6/bits/list.tcc:76
[...]

PR #1733, opened 2017-02-13
[Reviewer: Ellie] Add locking to RalfACR

sebrexmetaswitch:
The RalfACR can be accessed by multiple SCSCFSproutletTsxs, so we need to add locking.

This is a fix for issue #981 

#### Testing
 - tested on a machine that was hitting this crash under load before, and now doesn't

### DIFF
--- a/include/acr.h
+++ b/include/acr.h
@@ -172,8 +172,17 @@ public:
   /// @param session_id       The session ID to use.
   virtual void override_session_id(const std::string& session_id);
 
-  // Get/release the lock
+  /// Get a lock on the ACR.
+  ///
+  /// An ACR can be accessed from multiple threads (if there are multiple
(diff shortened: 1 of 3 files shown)


src/acr.cpp
```diff
@@ -96,6 +96,14 @@ void ACR::override_session_id(const std::string& session_id)
 {
 }
 
+void ACR::lock()
```
eleanor-merry: Can you comment why it's OK that these functions are empty? Also, you might as well just have them in the header.

include/acr.h
```diff
@@ -172,6 +172,10 @@ class ACR
   /// @param session_id       The session ID to use.
   virtual void override_session_id(const std::string& session_id);
 
+  // Get/release the lock
```
eleanor-merry: Worth explaining why this needs locking?

PR #1737, opened 2017-02-15
[Reviewer: EM] Change logging level to prevent spammy log

### LINKED ISSUE #1261: Sprout generates log spam for REGISTERs with reused credentials.
When a UE performs a re-registration with reused credentials Sprout generates the following log spam at log level 2:

```
17-11-2015 15:44:33.428 UTC Warning authentication.cpp:242: Received an authentication request for 3196910000@vm1.richlab.datcon.co.uk with nonce 79424c77495f1dc6, but no matching AV found
```

Given this is typical behavior for many UEs, this log should probably be moved to a lower logging level (or modified to identify a more specific form of error). This log becomes particularly annoying during traffic runs which simulate re-registrations with reused credentials.

KristaHyer:
Fixes spammy log

Fixes #1261

### DIFF
--- a/src/authenticationsproutlet.cpp
+++ b/src/authenticationsproutlet.cpp
@@ -298,7 +298,7 @@ pj_status_t AuthenticationSproutletTsx::user_lookup(pj_pool_t *pool,
 
   if (auth_challenge == NULL)
   {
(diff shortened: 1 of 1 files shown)


eleanor-merry:
LGTM

Issue #1609, opened 2016-11-27
upload_enum_json (and presumably other JSON uploaders) make it easy to upload an invalid file


rkd-msw:
#### Symptoms

I accidentally created an enum.json that was invalid JSON, and ran `upload_enum_json`, and it accepted it without a problem. (I could diagnose it after the fact from a "The ENUM file is invalid" ENT log.)

It feels like it would be fairly straightforward to automatically check that a JSON file is valid before uploading it, e.g. by running:

`python -c "import json; json.load(open('/etc/clearwater/enum.json'))" 2&>1`

and checking `$?`.

#### Impact

If all you do is `upload_enum_json`, I think the only impact will be confusion about why the new rules aren't taking effect (as the old rules don't get overwritten).

If you make a shared_config change at the same time, however (which the docs don't warn against):

- the `upload_shared_config` script will cause Sprout to restart
- it won't be able to read the s-cscf.json/enum.json/bgcf.json file because it's invalid
- it will start up without any rules, but will otherwise start up normally, so the queue manager will restart all the nodes in the cluster
- if the ENUM, S-CSCF or BGCF rules are necessary for calls to succeed, there might be an outage

#### Release and environment 

`1.0-cc-10.0.0-20161116-~255.00.0-161116.175811` on VMware.

#### Steps to reproduce

Create an enum.jsonfile that looks like `{"number_blocks": []` and run `sudo /usr/share/clearwater/clearwater-config-manager/scripts/
upload_enum_json`.

rkd-msw:
Andrew - you said you'd be able to pick this up.

PR #1745, opened 2017-02-20
[Reviewer: Adam]Enum prefix matching

Ying-metaswitch:
UT: added a json file where the more specific prefix comes later, test that EnumService is matching to the more specific prefix.
Implementation: current implementation uses a vector containing Pfix, I added a Map containing (prefix, Pfix) and do the matching search using the map, as Map is ordered by key length.

UT passed. 
#1609

### DIFF
--- a/src/enumservice.cpp
+++ b/src/enumservice.cpp
@@ -172,6 +172,7 @@ void JSONEnumService::update_enum()
   try
   {
     std::vector<NumberPrefix> new_number_prefixes;
+    std::map<std::string, NumberPrefix> new_prefix_regex_map;
 
     JSON_ASSERT_CONTAINS(doc, "number_blocks");
     JSON_ASSERT_ARRAY(doc["number_blocks"]);
@@ -196,6 +197,7 @@ void JSONEnumService::update_enum()
         if (parse_regex_replace(regex, pfix.match, pfix.replace))
         {
           new_number_prefixes.push_back(pfix);
+          new_prefix_regex_map.insert(std::make_pair(prefix, pfix));
           TRC_STATUS("  Adding number prefix %s, regex=%s",
                      pfix.prefix.c_str(), regex.c_str());
         }
@@ -217,6 +219,7 @@ void JSONEnumService::update_enum()
     // Take a write lock on the mutex in RAII style
     boost::lock_guard<boost::shared_mutex> write_lock(_number_prefixes_rw_lock);
     _number_prefixes = new_number_prefixes;
+    _prefix_regex_map = new_prefix_regex_map;
   }
   catch (JsonFormatError err)
   {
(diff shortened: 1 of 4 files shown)


src/ut/enumservice_test.cpp
```diff
@@ -198,6 +198,12 @@ TEST_F(JSONEnumServiceTest, BadRegex)
   ET("+15108580275", "").test(enum_);
 }
 
+TEST_F(JSONEnumServiceTest, PrefixMatching)
```
johadalin: This test name is not very descriptive. Can you add a comment to the test to describe exactly what it is testing? i.e. that whatever order we add enum rules in, we test that we match on the most specific match first, not matching by order added.

src/ut/test_enum_prefix_matching.json
```diff
@@ -0,0 +1,14 @@
...
+    {
+      "regex": "!(^.*$)!tel:\\1;npdi!",
+      "prefix": "+22",
+      "name": "+22 Numbers"
+    },
+    {
+      "regex": "!(^.*$)!tel:\\1;extra-parameter-here;npdi!",
+      "prefix": "+2223",
+      "name": "+2223 Numbers"
+    }
```
johadalin: Are you happy that the test is complete with only the two entries in the enum.json? 
Is there any other unusual behaviour that could come out of having a file with e.g. the following prefixes to match?:
```
+22
+2234
+223
```

I think this test does check the behaviour works as expected, but i want to check that you've done some thinking about any other possibilities

src/ut/enumservice_test.cpp
```diff
@@ -198,6 +198,12 @@ TEST_F(JSONEnumServiceTest, BadRegex)
...
+{
+  JSONEnumService enum_(string(UT_DIR).append("/test_enum_prefix_matching.json"));
+  ET("+22238899", "tel:+22238899;extra-parameter-here;npdi").test(enum_);
+}
```
johadalin: Are you happy that we don't need to test the less specific prefix still matches correctly? i.e. `+22456`

src/ut/enumservice_test.cpp
johadalin: I see there aren't many comments in the rest of the tests. You'd be welcome to add comments for them as well, if you've looked into what they are specifically testing. You can see an example of the sort of commenting i mean at https://github.com/Metaswitch/chronos/blob/dev/src/ut/test_timer.cpp#L310

src/enumservice.cpp
```diff
@@ -293,24 +296,22 @@ std::string JSONEnumService::lookup_uri_from_user(const std::string &user, SAS::
...
-  // For simplicity this uses a linear scan since we don't expect too many
-  // entries.  Should shift to a radix tree at some point.
-  for (std::vector<NumberPrefix>::const_iterator it = _number_prefixes.begin();
-       it != _number_prefixes.end();
+  // Iterate through map in reverse order (already sorted by key length during construction) to find 
+  // the most specific matching prefix
+  for (std::map<std::string, NumberPrefix>::const_reverse_iterator it = _prefix_regex_map.rbegin();
+       it != _prefix_regex_map.rend();
        it++)
   {
```
johadalin: I think it might be a good idea to add a read_lock here using the `_number_prefixes_rw_lock`, similar to the bgcfservice, at https://github.com/Metaswitch/sprout/blob/dev/src/bgcfservice.cpp#L247
Grab me if you're unsure at all

src/enumservice.cpp
```diff
@@ -296,9 +296,10 @@ std::string JSONEnumService::lookup_uri_from_user(const std::string &user, SAS::
 // the object.
 const JSONEnumService::NumberPrefix* JSONEnumService::prefix_match(const std::string& number) const
 {
-  // Iterate through map in reverse order (already sorted by key length during construction) to find 
-  // the most specific matching prefix
-  for (std::map<std::string, NumberPrefix>::const_reverse_iterator it = _prefix_regex_map.rbegin();
+  // Iterate through map in reverse order (already sorted by key length during 
+  // construction) to find  the most specific matching prefix
```
johadalin: double space between 'find' and 'the'

src/ut/enumservice_test.cpp
```diff
@@ -190,29 +190,44 @@ TEST_F(JSONEnumServiceTest, BadRegex)
...
+  // and three digits
   JSONEnumService enum_(string(UT_DIR).append("/test_enum_prefix_matching.json"));
-  ET("+22238899", "tel:+22238899;extra-parameter-here;npdi").test(enum_);
+  // This string is modified by the rule applying to three-digits-prefix, rather
+  // than the two-digits-prefix that comes earlier in the file
+  ET("+22238899", "tel:+22238899;three-digits-prefix-match;npdi").test(enum_);
+  ET("+22338899", "tel:+22338899;two-digits-prefix-match;npdi").test(enum_);
+  // Matches to the most specific four-digits-prefix, rather than any rule
+  // before or after it
+  ET("+22228899", "tel:+22228899;four-digits-prefix-match;npdi").test(enum_);
```
johadalin: Great :)

PR #1741, opened 2017-02-17
[BGCF-396] Protect against unbound etcd memory growth

huws:
Co-requisite change for this clearwater-etcd PR:  https://github.com/Metaswitch/clearwater-etcd/pull/409

PR #1747, opened 2017-02-21
[Reviewer: Adam]Check upload json file size & Allow large file using third argument

Ying-metaswitch:
Fix according to Rob's suggestion in https://github.com/Metaswitch/clearwater-etcd/pull/409#issuecomment-281081813

In /usr/share/clearwater/clearwater-config-manager/scripts/upload_json, check if the file size is larger than specified (10KB) while NOT using third argument --allow-large. If so, print an error message and exit early.

Tested in real system by uploading a 7KB and 14KB enum.json (using upload_enum_json) and getting the key to see if the upload is successful. 7KB worked, 14KB threw the error message, and 14KB with --allow-large set in the upload_enum_json worked.

#1741

### DIFF
--- a/sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
+++ b/sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
@@ -35,6 +35,7 @@
 CONFPATH=/etc/clearwater
 FILENAME=$1
 KEY=$2
+ALLOW_LARGE_UPLOADS =$3
 
 function usage() {
   echo "Usage: upload_json <filename> <key>"
@@ -64,6 +65,18 @@ fi
 if [[ ! -f $CONFPATH/$FILENAME ]]
 then
   echo "No configuration file at $CONFPATH/$FILENAME, unable to upload"
+  exit 2
+fi
+
+# Reject file that is too large for etcd. But allow larger file if the script is
+# called with --allow-large, so that the wrapper script is backward compatible.
+# All scripts calling this function will pass the optional argument directly to
+# this function to check if it's --allow-large
+uploadsize=10000
+filesize=$(wc -c <"$CONFPATH/$FILENAME")
+if [[ $filesize -ge $uploadsize && $MODE != "--allow-large" ]]
+then
+  echo "File is larger than the recommended size $uploadsize bytes, use --allow-large if you want to upload"
   exit 2
 fi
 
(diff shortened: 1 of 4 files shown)


sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
```diff
@@ -67,6 +68,17 @@ then
   exit 2
 fi
 
+# Check the file is not too large for etcd. Allow larger file if certain
+# argument is used,
+# so that the wrapper script is backward compatible
```
johadalin: I think saying 'if a certain argument is used' is a little vague. 
It's clear in the code below, but better to have a comment saying something like 'Allow uploads of larger files if the script was called with the '--allow-large' option'.

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
```diff
@@ -35,6 +35,7 @@
 CONFPATH=/etc/clearwater
 FILENAME=$1
 KEY=$2
+MODE=$3
```
johadalin: It isn't clear what 'Mode' means at this point in the script.
I think it would be good to have a name making clear what this is, i.e. ALLOW_LARGE_UPLOADS or something?

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
```diff
@@ -67,6 +68,17 @@ then
...
 fi
 
+# Check the file is not too large for etcd. Allow larger file if certain
+# argument is used,
+# so that the wrapper script is backward compatible
+uploadsize=10000
+filesize=$(wc -c <"$CONFPATH/$FILENAME")
+if [[ $filesize -ge $uploadsize && $MODE != "--allow-large" ]]
+then
+  echo "File is larger than the recommended size $uploadsize bytes, use --allow-large if you want to upload"
```
johadalin: Good, but could be clearer on where to use this option. 
This error message will be printed if they run e.g. `/usr/share/clearwater/clearwater-config-manager/scripts/upload_enum_config`. If they re-run this with `--allow-large` on the end, nothing will change. The actual point is that `upload_json` needs to be called with the extra argument.

We can talk about a couple of options here later.

Ying-metaswitch: All (three) bash scripts that use upload_json should be updated to pass an optional argument directly through to upload_json. Upload_json has been modified to check this optional argument for exact string '--allow-large'. This involves around six lines of change in total and exposes this allow-large options for all relevant cases to end users.
The other option is to modify the error message only asking end users to change the bash scripts to put in this option, which isn't ideal for usability.

Ying-metaswitch:
Fixed and squashed to one commit

PR #1748, opened 2017-02-22
[Reviewer: Adam] Better SIP error code on 500/503 from homestead

sebrexmetaswitch:
Turn a 500/503 error from homestead when looking up IFCs into a SIP 504 error.

This only covers this specific case, not the more general problem of turning HTTP error codes into the correct SIP error code. We think we should be doing this kind of thing more generally, but we don't want to be making such a wide-reaching change at this point.

So the plan is to merge this to release-114-fixes and then dev, and then raise an issue against dev to ensure we look at a fuller fix for this in future.

#### Testing done
- UTs added to check we're sending a 504 in the 2 places that it was added
- manually live tested to ensure we actually send a 504 when we get a 503 response from homestead for the HTTP PUT that we send as part of the INVITE flow to set up a call
- clearwater-live-tests run over a deployment including this fix and verified to pass

### DIFF
--- a/include/scscfsproutlet.h
+++ b/include/scscfsproutlet.h
@@ -142,8 +142,9 @@ private:
                       const std::string& binding_id,
                       SAS::TrailId trail);
 
-  /// Read data for a public user identity from the HSS.
-  bool read_hss_data(const std::string& public_id,
+  /// Read data for a public user identity from the HSS. Returns the HTTP result
+  /// code obtained from homestead.
+  long read_hss_data(const std::string& public_id,
                      const std::string& private_id,
                      const std::string& req_type,
                      bool cache_allowed,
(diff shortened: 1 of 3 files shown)


johadalin:
All looks reasonable. I'm assuming you're going raise the issue to track remaining work after this gets merged.
Function changes look fine. The second unit test is a bit gnarly, but i assume is copied and modified from other tests. Looks like some of the rest of them are pretty big as well, so that's fine. 
I think it all looks fine, but don't know the flows very well, so if it's not a pretty direct copy and you aren't that sure on it either, might be worth getting Ellie or Rob to check the flow logic. Otherwise, fine to merge :+1:

sebrexmetaswitch:
Yep, the second one was copied from an existing App Server flow. All I'm trying to hit with that is the other case where we return a 504.

PR #1749, opened 2017-02-22
[Reviewer: Adam] Enum visual separators

### LINKED ISSUE #1622: bgcf.json and enum.json are inconsistent in whether they allow visual separators in numbers
#### Symptoms

Initially, I had my system configured like this:

```
clearwater@ame21-sprout-1:~$ cat /etc/clearwater/enum.json
{
  "number_blocks": [
    {
      "regex": "!(^.*$)!tel:\\1;npdi!",
      "prefix": "+22233",
      "name": "+22233 Numbers"
    }
  ]
}
clearwater@ame21-sprout-1:~$ cat /etc/clearwater/bgcf.json
{
    "routes" : [
       {    "name" : "off-net calling",
            "number" : "+2.2-2-(33)",
            "route" : ["sip:icscf.sprout.ame22.clearwater.test;transport=TCP;myparam"]
       }
    ]
}
```

The visual separators in bgcf.json (i.e. representing the number as `+2.2-2-(33)`) worked - see Working.sas.

Then I changed my enum.json to:

```
{
  "number_blocks": [
    {
      "regex": "!(^.*$)!tel:\\1;npdi!",
      "prefix": "+2.2-2-(33)",
      "name": "+22233 Numbers"
    }
  ]
}
```

i.e. with the "prefix" field in the same format as the "number" field in bgcf.json. In this case, I got the SAS log saying "ENUM processing for +22233 failed to find a terminal rule" - see ENUM_failed.sas. (By chance, the call happened to succeed because the ENUM failure meant it got sent to the BGCF, but it would have failed or been misrouted if I was relying on ENUM to change the number in some way.)

#### Impact

This is a minor inconsistency, but it will confuse the user and make the details of what each fie has to look like harder to remember.

Additionally, because the enum.json file is accepted rather than rejected with a parsing error (as it would be for invalid JSON), it's possible to make a mistake that causes a service outage for some subscribers (because their ENUM rule stops matching) and not notice.

#### Release and environment 

`10.0.0-20161116-~255.00.0-161116.175811`

#### Steps to reproduce

Configure an enum.json rule with (, ), - or . in the prefix. Attempt to make a call which should match that prefix.

Ying-metaswitch:
Fixed #1622 to allow visual separators in enum json file, by using the same strip-separators functions as used in bgcfservice. Tested.

### DIFF
--- a/src/enumservice.cpp
+++ b/src/enumservice.cpp
@@ -46,6 +46,7 @@
 #include <sys/types.h>
 #include <netdb.h>
 
+#include "pjutils.h"
 #include "enumservice.h"
 #include "dnsresolver.h"
 #include "utils.h"
@@ -189,13 +190,17 @@ void JSONEnumService::update_enum()
         std::string regex;
         JSON_GET_STRING_MEMBER(*nb_it, "regex", regex);
 
-        // Entry is well-formed, so add it.
+        // Entry is well-formed, so strip off visual separators and add it.
         TRC_DEBUG("Found valid number prefix block %s", prefix.c_str());
         NumberPrefix pfix;
+        prefix = PJUtils::remove_visual_separators(prefix);
         pfix.prefix = prefix;
 
         if (parse_regex_replace(regex, pfix.match, pfix.replace))
         {
+          // Create an array in order of entries in json file, and a map
+          // (automatically sorted in order of key length) so we can later
+          // match numbers to the most specific prefixes
           new_number_prefixes.push_back(pfix);
           new_prefix_regex_map.insert(std::make_pair(prefix, pfix));
           TRC_STATUS("  Adding number prefix %s, regex=%s",
@@ -308,7 +313,8 @@ const JSONEnumService::NumberPrefix* JSONEnumService::prefix_match(const std::st
     TRC_DEBUG("Comparing first %d numbers of %s against prefix %s",
               len, number.c_str(), (*it).first.c_str());
 
-    if (number.compare(0, len, (*it).first, 0, len) == 0)
+    if (PJUtils::remove_visual_separators(number).
+                                      compare(0, len, (*it).first, 0, len) == 0)
     {
       // Found a match, so return it. 
       TRC_DEBUG("Match found");
--- a/src/ut/enumservice_test.cpp
+++ b/src/ut/enumservice_test.cpp
@@ -215,6 +215,16 @@ TEST_F(JSONEnumServiceTest, PrefixMatching)
   ET("+22228899", "tel:+22228899;four-digits-prefix-match;npdi").test(enum_);
 }
 
+// Test if prefix matching correctly ignores all visual separators
+TEST_F(JSONEnumServiceTest, StripSeparators)
+{
+  // Test enum file contains prefixes with visual separators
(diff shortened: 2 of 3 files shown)


src/enumservice.cpp
```diff
@@ -192,12 +193,16 @@ void JSONEnumService::update_enum()
         // Entry is well-formed, so add it.
         TRC_DEBUG("Found valid number prefix block %s", prefix.c_str());
         NumberPrefix pfix;
-        pfix.prefix = prefix;
+        pfix.prefix = PJUtils::remove_visual_separators(prefix);
 
         if (parse_regex_replace(regex, pfix.match, pfix.replace))
         {
+          // Create an array in order of entries in json file, and a map in
```
johadalin: Isn't `new_number_prefixes` a vector?
I think we can re-work this comment. The important point is that we put the information from the json file into a map, keyed by the prefix, as this will automatically give us sorting by prefix length. Something to clarify this could be useful, like 'Mapping NumberPrefix information to their prefix value automatically sorts by key length, so we can later easily match numbers to the most specific prefixes first'

src/enumservice.cpp
```diff
@@ -192,12 +193,16 @@ void JSONEnumService::update_enum()
         // Entry is well-formed, so add it.
         TRC_DEBUG("Found valid number prefix block %s", prefix.c_str());
         NumberPrefix pfix;
-        pfix.prefix = prefix;
+        pfix.prefix = PJUtils::remove_visual_separators(prefix);
```
johadalin: Rather than calling remove visual separators twice in this function, could we instead do something like `prefix = PJUtils::remove_visual_separators(prefix);` after we parse the string from the json file, and then just use the string as normal here?

src/ut/enumservice_test.cpp
```diff
@@ -200,7 +200,8 @@ TEST_F(JSONEnumServiceTest, BadRegex)
 }
 
 // Test if phone number matches to the most specific prefix in the enum json
-// file, rather than the first prefix that matches
+// file, rather than the first prefix that matches; and if visual separators are
+// correctly ignored in matching
```
johadalin: I think we should separate these checks out into two separate tests. Keeping them self-contained will make debugging easier if we break either piece of functionality in the future.
Can you create a separate test for 'IgnoreVisualSeparators'?

Ying-metaswitch:
Simple fixes according to review. Ready to merge into release-114-fixes?

src/ut/enumservice_test.cpp
```diff
@@ -210,11 +210,21 @@ TEST_F(JSONEnumServiceTest, PrefixMatching)
   // than the two-digits-prefix that comes earlier in the file
   ET("+22238899", "tel:+22238899;three-digits-prefix-match;npdi").test(enum_);
   ET("+22338899", "tel:+22338899;two-digits-prefix-match;npdi").test(enum_);
-  // Matches to the most specific four-digits-prefix, rather than any rule
+  // Matches to the most specific four-digits-prefix, rather than any rule
```
johadalin: whitespace at end of line. I think there is a vim or gitconfig option to remove all trailing whitespace. Grab me and we can look into it

PR #1754, opened 2017-02-27
[Reviewer: AJH] Sproutlet API change

sathiyan-sivathas:
This PR implements the Sproutlet API change that I sent you an email about.

The changes:
- Split out 100 trying handling into a separate function.
- Send 200 OKs to CANCELs between sproutlet wrappers.
- Send ACKs to negative responses between sproutlet wrappers.
- Add an observation API that sproutlets can use.


src/ut/mock_sproutlet.h
```diff
@@ -72,9 +72,9 @@ class MockSproutletTsx : public SproutletTsx
 
   MOCK_METHOD1(on_rx_initial_request, void(pjsip_msg*));
   MOCK_METHOD1(on_rx_in_dialog_request, void(pjsip_msg*));
-  MOCK_METHOD2(on_tx_request, void(pjsip_msg*, int));
+  MOCK_METHOD2(obs_tx_request, void(pjsip_msg*, int));
   MOCK_METHOD2(on_rx_response, void(pjsip_msg*, int));
-  MOCK_METHOD1(on_tx_response, void(pjsip_msg*));
+  MOCK_METHOD1(obs_tx_response, void(pjsip_msg*));
```
AlexHockey: Might as well add mocks for the other new methods you've added at the same time.

sathiyan-sivathas:
I've applied all the markups - the only outstanding issue is the one about ACRs which I've emailed you about separately. Please could you re-review?

On the max-forwards question - I changed the ordering to store the received request without decrementing the max-forwards because the upstream sproutlet uses the same pointer to see the request that it sent downstream. If we decrement max-forwards on receiving a request then the ACK generated to any negative responses will have the wrong mf header. This becomes a problem if the upstream sproutlet sends a request with mf=1 and we decrement it to 0 on receiving it downstream. Then any ACK to a negative response that is generated upstream will have mf=0 and will be dropped.

sathiyan-sivathas:
A few more changes. I have:
- Added a flag `tsx_mgmt` to indicate to the `obs_*` callbacks when they are being passed a transaction management message.
- Moved the 100 Trying generation to the sproutlet wrappers.
- Covered some edge cases e.g. send ACK to negative response from UASTsx to the root sproutlet, and 200 OK to CANCEL from the UASTsx to the last sproutlet in the chain.

I've done some initial testing for LI and all the messages we need seem to be available. Please review.

src/sproutletproxy.cpp
```diff
@@ -1074,6 +1108,34 @@ void SproutletProxy::UASTsx::tx_cancel(SproutletWrapper* upstream,
...
+  {
+    // Pass the ACK request to the downstream Sproutlet.
+    SproutletWrapper* downstream = i->second;
+    TRC_DEBUG("Route negative ACK to %s", downstream->service_name().c_str());
+    downstream->rx_negative_ack(ack);
+
+    // This is an ACK to a final response so disconnect the sproutlets.
+    _dmap_sproutlet.erase(std::make_pair(upstream, fork_id));
+    UMap::iterator j = _umap.find((void*)downstream);
+    _umap.erase(j);
```
AlexHockey: Worth commonalizing this code?

src/sproutletproxy.cpp
```diff
@@ -1607,6 +1681,14 @@ void SproutletWrapper::rx_request(pjsip_tx_data* req)
...
   }
 
+  // Decrement Max-Forwards if present.
+  pjsip_max_fwd_hdr* mf_hdr = (pjsip_max_fwd_hdr*)
+                      pjsip_msg_find_hdr(clone, PJSIP_H_MAX_FORWARDS, NULL);
+  if (mf_hdr != NULL)
+  {
+    --mf_hdr->ivalue;
+  }
+
```
AlexHockey: Hmm. When do we add the message to the list of messages that the wrapper owns? If we make any changes to it between storing it and passing it into the `on_rx_*` methods then the message passed to `on_rx_*` is different from messages returned by `original_request`, which feels like an issue.

src/sproutletproxy.cpp
```diff
@@ -1643,58 +1725,122 @@ void SproutletWrapper::rx_response(pjsip_tx_data* rsp, int fork_id)
     log_inter_sproutlet(rsp, false);
   }
 
-  register_tdata(rsp);
-  if ((PJSIP_IS_STATUS_IN_CLASS(rsp->msg->line.status.code, 100)) &&
-      (_forks[fork_id].state.tsx_state == PJSIP_TSX_STATE_CALLING))
+
+  int status_code = rsp->msg->line.status.code;
+  if (status_code == PJSIP_SC_TRYING)
```
AlexHockey: Doesn't this need to cover CANCEL 200s as well?

AlexHockey: Ah, I see you handle this below. It would be clearer to put all the observation stuff in one place, even if it means repeating some if tests.

src/sproutletproxy.cpp
```diff
@@ -1643,58 +1725,122 @@ void SproutletWrapper::rx_response(pjsip_tx_data* rsp, int fork_id)
...
+    tx_negative_ack(rsp, fork_id);
+  }
+
+  // If this is a 200 OK to a CANCEL, we don't want to do anything with it, so
+  // just notify the sproutlet that it has been received and then drop our
+  // reference to it.
+  if (PJSIP_MSG_CSEQ_HDR(rsp->msg)->method.id == PJSIP_CANCEL_METHOD)
+  {
+    _sproutlet_tsx->obs_rx_response(rsp->msg, fork_id, true);
+    pjsip_tx_data_dec_ref(rsp);
```
AlexHockey: Don't we also need to drop our reference to a 100 trying?

AlexHockey: Maybe you need to do this further down for 100 tryings. But in that case I would move both the 100 and 200 CANCEL processing further down.

src/sproutletproxy.cpp
```diff
@@ -1879,6 +2025,8 @@ void SproutletWrapper::aggregate_response(pjsip_tx_data* rsp)
     return;
   }
 
+  // Sproutlets don't forward 100 Trying responses so we should never hit this
+  // code. But let's keep it here just in case.
```
AlexHockey: You've missed the point. even though they shouldn't, sproutlet's *can* forward 100 trying responses (there is nothing stopping them) so we *need* to keep this code.

src/sproutletproxy.cpp
```diff
@@ -625,9 +639,6 @@ void SproutletProxy::UASTsx::process_cancel_request(pjsip_rx_data* rdata)
     // Pass the CANCEL to the Sproutlet at the root of the tree.
     pjsip_tx_data* tdata = PJUtils::clone_msg(stack_data.endpt, rdata);
     _root->rx_cancel(tdata);
-
-    // Schedule any requests generated by the Sproutlet.
-    schedule_requests();
```
sathiyan-sivathas: The sproutlet shouldn't have generated any requests from the CANCEL, that's why I removed this code.

If it makes you uneasy, I can add it back.

AlexHockey: As discussed, the TSX shouldn't be forwarding the CANCEL but it can do other work, and *that* would need to be scheduled.

PR #1756, opened 2017-02-28
Fix up record-routing for AS default handling

AlexHockey:
Previously we had an issue where if a session continued AS was bypassed, the onward request didn't have the correct record route headers. This is because we record-route when determining the served user, but then call `original_request` when retrying to the next AS. 

I've fixed this by saving off the request when the S-CSCF record-routes itself, and using this request as a baseline when bypassing an AS. 

I've reproduced the original error in UT, and used this to checked the fix works.

### DIFF
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -2197,7 +2197,7 @@ std::string SCSCFSproutletTsx::fork_failure_reason_as_string(int fork_id, int si
 
 pjsip_msg* SCSCFSproutletTsx::get_base_request()
 {
-  if (_base_req)
+  if (_base_req != nullptr)
   {
     return clone_msg(_base_req);
   }
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -404,7 +404,8 @@ SCSCFSproutletTsx::SCSCFSproutletTsx(SproutletTsxHelper* helper,
   _video_call(false),
   _impi(),
   _auto_reg(false),
-  _se_helper(stack_data.default_session_expires)
+  _se_helper(stack_data.default_session_expires),
+  _base_req(nullptr)
 {
   TRC_DEBUG("S-CSCF Transaction (%p) created", this);
 }
@@ -443,6 +444,11 @@ SCSCFSproutletTsx::~SCSCFSproutletTsx()
   }
 
   _target_bindings.clear();
+
+  if (_base_req != nullptr)
+  {
+    free_msg(_base_req);
+  }
 }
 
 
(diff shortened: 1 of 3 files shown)


src/scscfsproutlet.cpp
```diff
@@ -1903,6 +1909,14 @@ void SCSCFSproutletTsx::add_to_dialog(pjsip_msg* msg,
     // terminating hop.  Update the billing role.
     pj_strdup(pool, &param->value, pjsip_billing_role);
   }
+
+  // Store off the modified message - we may need it later if we need to invoke
+  // default handling for an AS.
+  if (_base_req != nullptr)
```
eleanor-merry: Why sometimes `if (_base_req != nullptr)` and sometimes `if (_base_req)`

src/ut/scscf_test.cpp
```diff
@@ -4628,6 +4628,347 @@ TEST_F(SCSCFTest, DefaultHandlingMalformed)
...
+  // The S-CSCF should record-route itself for both originating and terminating
+  // billing.
+  EXPECT_THAT(get_headers(out, "Record-Route"),
+              MatchesRegex("Record-Route:.*billing-role=charge-term.*"
+                           "Record-Route:.*billing-role=charge-orig.*"));
+
+  free_txdata();
+}
+
+TEST_F(SCSCFTest, DefaultHandlingContinueTimeoutRRTest)
```
eleanor-merry: Can you give a high level comment for what each test is testing for?

src/ut/scscf_test.cpp
```diff
@@ -4628,6 +4628,347 @@ TEST_F(SCSCFTest, DefaultHandlingMalformed)
...
+  //    header should indicate no billing.
+  // -  There was only one signaling hop on the originating side, so there
+  //    should only be one RR header from originating processing, which should
+  //    indicate originating billing).
+  EXPECT_THAT(get_headers(out, "Record-Route"),
+              MatchesRegex("Record-Route:.*billing-role=charge-none.*"
+                           "Record-Route:.*billing-role=charge-orig.*"));
+  free_txdata();
+
+  stack_data.record_route_on_initiation_of_terminating = false;
```
eleanor-merry: Is this something that should happen in the tidy up of the test, rather than in an individual test case?

src/scscfsproutlet.cpp
AlexHockey: Because I was inconsistent for some reason. I'll fix up.

src/ut/scscf_test.cpp
AlexHockey: Yeah, good idea.

src/ut/scscf_test.cpp
AlexHockey: I quite like the symmetry that if you change something you're responsible for putting it back, as it makes the code easier to reason about (similarly - it's nicer if a function that takes a lock also releases it). It also gives us more flexibility in future - if we wanted to add a new fixture that modifies the `stack_data` settings in `SetUpTestCase`, resetting the values after each run would stop us from doing that.

eleanor-merry: Is it worth storing the current values at the start of the test then, and setting them back to the original values?

AlexHockey: Yeah, that's probably more correct. I'll do that.

PR #1762, opened 2017-03-01
[Reviewer: RKD] Allow list headers to have a trailing comma

Neil-Bruce:
Rob,

Could you review this change to the tests for the Supported: header with a trailing comma?  Passing direct to you since you've looked at this already - hope that's OK.

Note that I have not run the live tests against this fix as it seems fairly minor.  Let me know if you think I need to run them before releasing.

Neil

### DIFF
--- a/src/ut/common_sip_processing_test.cpp
+++ b/src/ut/common_sip_processing_test.cpp
@@ -461,6 +461,50 @@ TEST_F(CommonProcessingTest, BadResponseDropped)
   ASSERT_EQ(0, txdata_count());
 }
 
+TEST_F(CommonProcessingTest, BadSupportedHeader1)
+{
+  // Tests that a trailing comma on request is accepted.
+  pjsip_tx_data* tdata;
+  pjsip_endpt_register_module(stack_data.endpt, &mod_ok);
+
+  // Inject a request with a trailing comma.
+  Message msg1;
+  msg1._extra = "Supported: timer,";
+  inject_msg(msg1.get_request(), _tp);
+
+  // Expect a response from mod_ok.
+  ASSERT_EQ(1, txdata_count());
+
+  tdata = current_txdata();
+  RespMatcher r1(200);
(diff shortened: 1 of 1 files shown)


rkd-msw:
Looks good, but see https://github.com/Metaswitch/pjsip-upstream/pull/61#issuecomment-283387532 for some more suggested tests.

rkd-msw:
If you add more testcases, it might be best to parameterise this, e.g.:

rkd-msw:
```
TEST_F(CommonProcessingTest, CommasInSupportedHeader)
{
  // Tests that a trailing comma on request is accepted.
  pjsip_tx_data* tdata;
  pjsip_endpt_register_module(stack_data.endpt, &mod_ok);

  std::vector<std::string> headers = {"Supported: timer,", "Supported: \n  ,,,,"};

  for (std::string h: headers) {
    SCOPED_TRACE(h);
    // Inject a request with a trailing comma.
    Message msg1;
    msg1._extra = h;
    inject_msg(msg1.get_request(), _tp);

    // Expect a response from mod_ok.
    ASSERT_EQ(1, txdata_count());

    tdata = current_txdata();
    RespMatcher r1(200);
    r1.matches(tdata->msg);

    free_txdata();
  }
  pjsip_endpt_unregister_module(stack_data.endpt, &mod_ok);
}
```

Neil-Bruce:
I tried running multiple tests in a single test case before.  The first one runs OK but all subsequent cases get a "503 Service Unavailable" error instead of the expected 200 OK.  Couldn't get it to work before and the same thing happens with your code above.  I'll try and work it out but do you have any idea why?

Neil-Bruce:
Ah, I think it is the load monitor running out of tokens.

Neil-Bruce:
Extra test cases added as above but with a new load monitor to allow all tests to complete.

rkd-msw:
👍

PR #1768, opened 2017-03-07
[Reviewer: AJH] Store full path headers

sathiyan-sivathas:
Alex, this PR implements the fix to the path header issue that you sent me a design for. I think everything is as in the design.

I've not made the Bono changes yet and I haven't added UTs yet. Those will follow.


src/ut/contact_filtering_test.cpp
```diff
@@ -648,6 +650,60 @@ TEST_F(ContactFilteringBindingToTargetTest, SimpleConversion)
...
+    EXPECT_EQ(path, *j);
+    ++j;
+  }
+
+  EXPECT_EQ((pjsip_transport*)NULL, target.transport);
+  EXPECT_EQ(0, target.liveness_timeout);
+  EXPECT_EQ(300, target.contact_expiry);
+  EXPECT_EQ((unsigned)1234, target.contact_q1000_value);
+}
+TEST_F(ContactFilteringBindingToTargetTest, SimpleConversionDownlevel)
```
AlexHockey: As the code evolves and we do more of this kind of upgrade, there may be many different views on what downlevel means. I would call this test something to do with Path URIs rather than just downlevel.

PR #1769, opened 2017-03-07
[Reviewer: Richard] Update analytics generation to not override user changes

johadalin:
This should mean that users can modify the config file created, and then not lose those changes when cw-infra restarts. 
Seem reasonable?

### DIFF
--- a/sprout-base.root/usr/share/clearwater/infrastructure/scripts/create-analytics-syslog-config
+++ b/sprout-base.root/usr/share/clearwater/infrastructure/scripts/create-analytics-syslog-config
@@ -39,17 +39,6 @@
 syslog_conf_file=/etc/rsyslog.d/30-sproutanalytics.conf
 temp_file=$(mktemp sproutanalytics.syslog.XXXXXXXX)
 
-if [ -f $syslog_conf_file ]; then
-  # If the config file exists, and has been manually modified from
-  # the default set up, we don't want to override it. Users should
-  # have removed this line if they were modifying the file.
-  grep "# Clearwater auto-generated configuration" $syslog_conf_file > /dev/null 2>&1
-  rc=$?
-  if [ $rc != 0 ]; then
-    exit 0
-  fi
-fi
-
 . /etc/clearwater/config
 
 if [ -n "$remote_audit_logging_server" ]; then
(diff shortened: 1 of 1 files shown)


sprout-base.root/usr/share/clearwater/infrastructure/scripts/create-analytics-syslog-config
```diff
@@ -39,6 +39,17 @@
 syslog_conf_file=/etc/rsyslog.d/30-sproutanalytics.conf
 temp_file=$(mktemp sproutanalytics.syslog.XXXXXXXX)
 
+if [ -f $syslog_conf_file ]; then
+  # If the config file exists, and has been manually modified from
+  # the default set up, we don't want to override it. Users should
+  # have removed this line if they were modifying the file.
+  grep "# Clearwater auto-generated configuration" $syslog_conf_file > /dev/null 2>&1
```
richardwhiuk: Surely you can just use `grep -q` rather than output redirection. Also, if grep throws an error, I suspect we may want it, rather than throwing it away.

sprout-base.root/usr/share/clearwater/infrastructure/scripts/create-analytics-syslog-config
```diff
@@ -39,6 +39,17 @@
...
+  rc=$?
+  if [ $rc != 0 ]; then
```
richardwhiuk: You can just do `if grep -q PATTERN file.txt; then` then rather than catching the result code in a variable.

sprout-base.root/usr/share/clearwater/infrastructure/scripts/create-analytics-syslog-config
```diff
@@ -39,6 +39,17 @@
 syslog_conf_file=/etc/rsyslog.d/30-sproutanalytics.conf
 temp_file=$(mktemp sproutanalytics.syslog.XXXXXXXX)
```
richardwhiuk: You should do move your code prior to creation of the temp file, or you'll just leave it lying around (it gets deleted on line 86/91.

sprout-base.root/usr/share/clearwater/infrastructure/scripts/create-analytics-syslog-config
```diff
@@ -39,6 +39,17 @@
...
 
+if [ -f $syslog_conf_file ]; then
+  # If the config file exists, and has been manually modified from
```
richardwhiuk: I'd normally put the comment above the block of code.

johadalin:
Ok. I've set this up to handle it in a different way, using shared_config. I'll be updating the docs with the right corresponding changes on the logging server too.
Thoughts?

sprout-base.root/usr/share/clearwater/infrastructure/scripts/create-analytics-syslog-config
```diff
@@ -49,7 +49,25 @@ if [ -n "$remote_audit_logging_server" ]; then
...
 fi
 
-cat > $temp_file << EOF
+# Add tls setup to the top of the config file if the option is specified
+shopt -s nocasematch
+if [[ $remote_audit_logging_use_tls == "true" ]]; then
+  cat > $temp_file << EOF
+\$ActionSendStreamDriver gtls
+
+#certificate files
```
richardwhiuk: Can you make this correctly capitalized, and with a space - so `# Certificate Files`

PR #1772, opened 2017-03-08
[Reviewer: Krista] Add wildcard support

eleanor-merry:
Sprout changes for wildcard support. Tested in UTs and live.


src/ut/custom_headers_test.cpp
```diff
@@ -843,6 +843,56 @@ TEST_F(CustomHeadersTest, PAssertedIdentity)
   pj_pool_release(clone_pool);
 }
 
+TEST_F(CustomHeadersTest, PProfileKey)
```
graemerobertson: Yep, rare comment slip-up here!

src/ut/custom_headers_test.cpp
```diff
@@ -843,6 +843,56 @@ TEST_F(CustomHeadersTest, PAssertedIdentity)
...
+
+  string str("INVITE sip:6505554321@homedomain SIP/2.0\n"
+             "Via: SIP/2.0/TCP 10.0.0.1:5060;rport;branch=z9hG4bKPjPtVFjqo;alias\n"
+             "Max-Forwards: 63\n"
+             "From: <sip:6505551234@homedomain>;tag=1234\n"
+             "To: <sip:6505554321@homedomain>\n"
+             "Contact: <sip:6505551234@10.0.0.1:5060;transport=TCP;ob>\n"
+             "Call-ID: 1-13919@10.151.20.48\n"
+             "CSeq: 1 INVITE\n"
+             "P-Profile-Key: <sip:uri!.*!!!@example.com>;test-param;test-param2=value\n"
```
graemerobertson: Worth throwing in a few more random valid characters into this URI to check we don't crash?!

src/ut/registrar_test.cpp
```diff
@@ -1974,6 +1974,33 @@ TEST_F(RegistrarTest, NonPrimaryAssociatedUri)
   delete aor_data; aor_data = NULL;
 }
 
+/// Register with an IMPU that only matches a wildcarded IMPU.
```
graemerobertson: Er, isn't this supposed to fail?

graemerobertson: Well, I thought the Register would fail, although possibly not at Sprout :S and I think this test is really testing the P-Associated-URIs?

rkd-msw:
Should we have tests for:

* a Service-Profile with two wildcards, one of which matches the user and one which doesn't?
* the case I think we discussed, where P-Profile-Key contains characters that need to be percent-escaped?

src/hssconnection.cpp
```diff
@@ -368,18 +393,40 @@ bool decode_homestead_xml(const std::string public_user_identity,
...
+        // (and will be shortly). It's therefore not fully UT'd; this will be
+        // done when it moves.
+        // LCOV_EXCL_START
+        rapidxml::xml_node<>* extension = public_id->first_node("Extension");
+        if (extension)
+        {
+          rapidxml::xml_node<>* type = extension->first_node("IdentityType");
+          if (type)
+          {
+            if (std::string(type->value()) == "2")
```
eleanor-merry: Yup :)

src/hssconnection.cpp
```diff
@@ -368,18 +393,40 @@ bool decode_homestead_xml(const std::string public_user_identity,
...
+            {
+              rapidxml::xml_node<>* new_identity = extension->first_node("WildcardedPSI");
+              uri = std::string(new_identity->value());
```
eleanor-merry: Good point - I'll fix this up in the commonising code PR

src/ut/registrar_test.cpp
eleanor-merry: Yes - I've updated the test so that it's got a realistic response from the HSS and renamed it.

eleanor-merry:
I've added the suggested tests

PR #1779, opened 2017-03-09
[Reviewer: Rob] Validate json files in the upload_json script, and log when they fail

### LINKED ISSUE #1609: upload_enum_json (and presumably other JSON uploaders) make it easy to upload an invalid file

johadalin:
Rob, 
We thought we had fixed #1609 , but had actually not done so amid a bit of confusion.

I've put the fix you suggested in now. Can you take a quick look at this to assess:

* Is putting this validation in the base `upload_json` file the right thing. i.e. can you think of anything we upload that is invalid, but still uses the upload_json script?
* Are you happy that this can go in at this point? We thought it was in before, and i think it's no-risk, but want to check. Andrew was going to test it anyway.


johadalin:
Ok. Fixed this up now. I've properly live tested, and confirmed valid json passes, invalid fails with the log, and is not uploaded.

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
```diff
@@ -80,6 +80,14 @@ then
   exit 2
 fi
 
+# Ensure that the file is valid JSON
+python -c "import json; json.load(open(\"$CONFPATH/$FILENAME\"))" 2&>1
```
rkd-msw: `python -m json.tool < $CONFPATH/$FILENAME > /dev/null 2>&1` might be a bit cleaner (as you avoid inline Python and escaping bugs).

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
```diff
@@ -80,6 +80,14 @@ then
...
+rc=$?
+if [ $rc != 0 ]; then
+  echo "File $CONFPATH/$FILENAME contains invalid json, and so will not be uploaded"
```
rkd-msw: This script now has five different error messages:

```
The Clearwater Configuration store (etcd) is not running
Please start it before uploading configuration
```

```
No configuration file at $CONFPATH/$FILENAME, unable to upload
```

```
File is larger than the recommended size 10000 bytes, use --allow-large if you want to upload
```

```
File $CONFPATH/$FILENAME contains invalid json, and so will not be uploaded
```

```
Upload JSON configuration file failed to $keypath
```

They all just feel inconsistent - is it "JSON" or "json"? Is it a "file" or a "configuration file"? Do we say "please" when asking the user to do something? Do we say "unable to upload" or "will not be uploaded" (or not explicitly say that)?

Should we have and follow a style guide for this help text, along the lines of https://www.postgresql.org/docs/9.5/static/error-style-guide.html or https://www.gov.uk/service-manual/design/help-text?

rkd-msw:
Let's just put it into V11 at this point (once you've addressed my markups), not V10.

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
rkd-msw: Neither option (the current thing or my suggestion) will upload an empty file. Do we need to handle that?

steelyvoid:
Adding dummy comment to get notifications

johadalin:
closing this PR, as i've rebased against master

PR #1793, opened 2017-03-23
[Reviewer: Krista] Parse SIFC XML file, and parse XML from Homestead

eleanor-merry:
Krista, can you review this PR to add some SIFC support

This S-CSCF can now pull out the shared IFCs from the User-Data XML. It then matches up the IDs with the IFCs set in /etc/clearwater/sifc.xml, which it parses at start of day, and on reload.

Tested in UTs only.


include/sifcservice.h
```diff
@@ -0,0 +1,77 @@
...
+#include <functional>
+
+#include "updater.h"
+#include "sas.h"
+#include "ifc.h"
+
+class SIFCService
+{
+public:
+  SIFCService(std::string configuration = "./sifc.xml");
```
eleanor-merry: We've discussed, and we'll wait for spec sign off before changing this (or any of the related comments)

src/ut/sifcservice_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+{
+  if (ifc._ifc->first_node("Priority"))
+  {
+    return std::atoi(ifc._ifc->first_node("Priority")->value());
+  }
+
+  return 0;
+}
+
+// Test a valid shared IFC file is parsed correctly
```
eleanor-merry: Only by looking at the internal map - but that's not ideal as it's not an interface we expect any code to use

src/ut/sifcservice_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+  SIFCService sifc(string(UT_DIR).append("/test_sifc.xml"));
+
+  // Load the IFC file, and check that it's been parsed correctly
+  std::set<int> id; id.insert(2);
+  std::multimap<int32_t, Ifc> ifc_map;
+  sifc.get_ifcs_from_id(ifc_map, id, 0);
+  EXPECT_EQ(ifc_map.size(), 1);
+  EXPECT_EQ(get_server_name(ifc_map.find(0)->second), "publish.example.com");
+
+  // Reload the file, then repeat the check. Nothing should have changed,
```
eleanor-merry: D'oh - added

src/ut/sifcservice_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+  // and there should be no memory issues
+  std::multimap<int32_t, Ifc> ifc_map_reload;
+  sifc.get_ifcs_from_id(ifc_map_reload, id, 0);
+  EXPECT_EQ(ifc_map_reload.size(), 1);
+  EXPECT_EQ(get_server_name(ifc_map_reload.find(0)->second), "publish.example.com");
+}
+
+// Test that the priorities are parsed correctly
```
eleanor-merry: I've moved this test to be with the other tests that have a part-valid XML file - I do want to test that a file with a mix of valid/invalid priorities is treated correctly though rather than isolate the two cases.

src/ut/sifcservice_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+  EXPECT_TRUE(log.contains("Invalid shared IFC set configuration file - missing Sets block"));
+  EXPECT_TRUE(sifc._shared_ifc_sets.empty());
+}
+
+// Test that we cope with the case that the shared IFC file is valid but empty
+TEST_F(SIFCServiceTest, NoEntries)
+{
+  CapturingTestLogger log;
+  SIFCService sifc(string(UT_DIR).append("/test_sifc_no_entries.xml"));
+  EXPECT_FALSE(log.contains("Failed"));
```
eleanor-merry: The logs indicating that something's gone wrong start with Failed, but then differ depending on the log - this test is checking that we haven't detected a failure, so this is checking that Failed is absent.

src/sifcservice.cpp
```diff
@@ -0,0 +1,227 @@
...
+  delete _root; _root = root;
+
+  rapidxml::xml_node<>* sets = _root->first_node("Sets");
+  rapidxml::xml_node<>* set = NULL;
+
+  for (set = sets->first_node("Set");
+       set != NULL;
+       set = set->next_sibling("Set"))
+  {
+    rapidxml::xml_node<>* set_id_node = set->first_node("SetId");
```
KristaHyer: When these string are updated to match the spec I think they should also be made into constants in the header file.

src/sifcservice.cpp
```diff
@@ -0,0 +1,227 @@
...
+      {
+        std::string priority_str = std::string(priority_node->value());
+        Utils::trim(priority_str);
+        priority = std::atoi(priority_str.c_str());
+
+        if (priority_str != std::to_string(priority))
+        {
+          TRC_ERROR("Invalid shared IFC set block - Priority (%s) isn't an int. Skipping this entry",
+                    priority_str.c_str());
+          CL_SPROUT_SIFC_FILE_INVALID_PRIORITY.log(set_id_str.c_str());
```
KristaHyer: Sorry I didn't spot this sooner - you pass in the set id string.
But in sprout_pd_definitions.h you expect to have received the priority string.

src/ut/sifcservice_test.cpp
```diff
@@ -0,0 +1,303 @@
...
+  std::set<int> id; id.insert(2);
+  std::multimap<int32_t, Ifc> ifc_map;
+  sifc.get_ifcs_from_id(ifc_map, id, 0);
+  EXPECT_EQ(ifc_map.size(), 1);
+  EXPECT_EQ(get_server_name(ifc_map.find(0)->second), "publish.example.com");
+
+  // Change the file the sifc service is using (to mimic the file being changed),
+  // then reload the file, and repeat the check. Nothing should have changed,
+  // and there should be no memory issues
+  sifc._configuration = "/test_sifc_parse_error.xml";
```
KristaHyer: Shouldn't this be:
string(UT_DIR).append("/test_sifc_parse_error.xml");
?

PR #1796, opened 2017-03-27
[Reviewer: Krista] Add the correct namespace to the XML

eleanor-merry:
Add the ere namespace to the NOTIFY XML.

Tested in UTs.


eleanor-merry:
This is the fix for https://github.com/Metaswitch/clearwater-issues/issues/1997

PR #1797, opened 2017-03-27
UTs for SiFCs

KristaHyer:
I've written three tests for shared ifcs
 - to check that ifcs are returned in a shared ifc is present
 - to check that shared ifcs are compatible with normal ifcs
 - to check that invalidly formatted shared ifcs aren't accepted.

Full_test passes, and branch "add_sifc_support" now has full coverage.


src/ut/hssconnection_test.cpp
```diff
@@ -59,6 +62,8 @@ class HssConnectionTest : public BaseTest
   AlarmManager _am;
   CommunicationMonitor _cm;
   HSSConnection _hss;
+  HSSConnection _sifc_hss;
```
eleanor-merry: I don't really like that you've got two HSSConnections in the test. I think you should either create a new HSSConnectionWithSIFCTest class, or make the _hss take the _sifc_service and remove _sifc_hss entirely (which would be simple to do given you added the check that the id set shouldn't be empty before calling get_ifcs_from_id).

src/ut/hssconnection_test.cpp
```diff
@@ -333,8 +350,76 @@ class HssConnectionTest : public BaseTest
           "<ECF priority=\"1\">ecf1</ECF>"
         "</ChargingAddresses>"
       "</ClearwaterRegData>";
- }
-
+    fakecurl_responses_with_body[std::make_pair("http://10.42.42.42:80/impu/pubid60/reg-data", "{\"reqtype\": \"reg\", \"server_name\": \"server_name\"}")] =
```
KristaHyer: done

KristaHyer:
@eleanor-merry 
This originally consisted of three tests, and after your first review I've added four more.
Since this is quite a large change, can you rereview these changes, looking at these four new tests before I merge this code in?
Thanks!

src/ut/hssconnection_test.cpp
```diff
@@ -625,3 +628,523 @@ TEST_F(HssConnectionTest, CacheNotAllowed)
...
+// Check that some iFCs are returned when a shared iFC set is encountered.
+TEST_F(HssWithSifcTest, SimpleSiFC)
+{
+  std::vector<std::string> uris;
+  std::map<std::string, Ifcs> ifcs_map;
+  std::string regstate;
+
+  // Mock out the function that converts the list of shared iFC set ids into a
+  // list of iFCs - this function is tested elsewhere.
+  Ifc* fake_ifc = new Ifc(NULL);
```
eleanor-merry: Mm. In each of your tests you create identical fake IFCs - I think this would be simpler if you created these as member variables of the test itself (and then you don't have to have the repeated clean up code either). It would also make the case where you need a real IFC simpler. I've pushed an example of what I mean to the sifc_ut_ifc_member branch,

src/ut/hssconnection_test.cpp
```diff
@@ -333,8 +337,7 @@ class HssConnectionTest : public BaseTest
           "<ECF priority=\"1\">ecf1</ECF>"
         "</ChargingAddresses>"
       "</ClearwaterRegData>";
- }
-
```
KristaHyer: done

src/ut/hssconnection_test.cpp
```diff
@@ -625,3 +628,523 @@ TEST_F(HssConnectionTest, CacheNotAllowed)
...
+    .WillOnce(SetArgReferee<0>(std::multimap<int32_t, Ifc>(ifcs_from_id)));
+
+  // Send in a message, and check the expected number of iFCs are present, as
+  // well as checking that the expected number with each priority are present.
+  _sifc_hss.update_registration_state("sifcifcmix", "", HSSConnection::REG, regstate, ifcs_map, uris, 0);
+  int32_t map_size = ifcs_map.begin()->second.size();
+  EXPECT_TRUE(map_size == 9);
+  std::vector<int32_t> expected_priorities = {1, 1, 1, 2, 2, 2, 3, 3, 4};
+  std::vector<int32_t> priorities;
+  int32_t ii;
```
KristaHyer: done?

eleanor-merry: Why ?

src/ut/hssconnection_test.cpp
```diff
@@ -625,3 +628,523 @@ TEST_F(HssConnectionTest, CacheNotAllowed)
...
+  ifcs_from_id.insert(std::pair<int32_t, Ifc>(1, Ifc(ifc_priority_one)));
+  ifcs_from_id.insert(std::pair<int32_t, Ifc>(2, Ifc(ifc_priority_two)));
+  ifcs_from_id.insert(std::pair<int32_t, Ifc>(2, Ifc(ifc_priority_two)));
+  // Expect input of two shared iFC sets, with ids 3 and 4.
+  const std::set<int32_t> id_set_one = {3, 4};
+  EXPECT_CALL(_sifc_service, get_ifcs_from_id(_, id_set_one, _))
+    .WillOnce(SetArgReferee<0>(std::multimap<int32_t, Ifc>(ifcs_from_id)));
+
+  // Send in a message, and check the expected number of iFCs are present, as
+  // well as checking that the expected number with each priority are present.
```
KristaHyer: done

src/ut/hssconnection_test.cpp
KristaHyer: why no _ in front?

src/ut/hssconnection_test.cpp
```diff
@@ -625,3 +628,523 @@ TEST_F(HssConnectionTest, CacheNotAllowed)
...
+  doc.append_node(ifc_priority_one);
+
+  rapidxml::xml_node<>* ifc_priority_two = doc.allocate_node(
+                           rapidxml::node_type::node_element, "IFCPriorityTwo");
+  rapidxml::xml_node<>* priority_two = doc.allocate_node(
+                            rapidxml::node_type::node_element, "Priority", "2");
+  ifc_priority_two->append_node(priority_two);
+  doc.append_node(ifc_priority_two);
+
+  // Mock out the function (which is tested elsewhere), that converts the list
```
KristaHyer: done

PR #1798, opened 2017-03-30
[Reviewer: Ellie] Include margin in default subscription expiry.

richardwhiuk:
TS 24.229 suggests the P-CSCF should send a higher subscription max expiry than
the registration. Thus we default the max subscription expiry to a higher value.

RFC 3680 suggests 3761 for an expiry of 3600, so we just add 161 seconds to the registration max expiry.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1559,7 +1559,7 @@ int main(int argc, char* argv[])
   if (opt.sub_max_expires == 0)
   {
     opt.sub_max_expires = opt.reg_max_expires + 161;
-    TRC_INFO("Subscription max expiry defaulted to %d based on registration max expiry %d",
+    TRC_INFO("Maximum subscription period defaulted to %d seconds, based maximum registration expiry %d seconds",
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -1552,6 +1552,17 @@ int main(int argc, char* argv[])
...
   }
 
+  // If sub_max_expires is unset, then allow a higher number than the
+  // maximum registration expiry, as required by TS 24.229 5.2.3.
+  // RFC 3680 suggests 3761 seconds for an expiry of 3600. We thus add
+  // 161 seconds for the maximum subscription expiry by default.
+  if (opt.sub_max_expires == 0)
+  {
+    opt.sub_max_expires = opt.reg_max_expires + 161;
+    TRC_INFO("Subscription max expiry defaulted to %d based on registration max expiry %d",
```
eleanor-merry: "Maximum subscription period (in seconds)" rather than "Subscription max expiry" (and likewise for the reg max expiry) to match the wording of the other logs?

PR #1799, opened 2017-04-07
[Reviewer: AJH] Update sproutlet selection algorithm

sathiyan-sivathas:
Alex, this PR implements the updated sproutlet selection algorithm that you wrote a design for. Everything is in line with the design, except that:
- There is a separate map for service names and ports. This made sense since the service name is a string, and the port is an int.

`make full_test` passes.


include/sproutletproxy.h
```diff
@@ -85,6 +85,14 @@ class SproutletProxy : public BasicProxy
   /// Create Sproutlet UAS transaction objects.
   BasicProxy::UASTsx* create_uas_tsx();
 
+  // Checks if a sproutlet service name is taken
+  bool is_service_taken(std::string service_name,
+                        Sproutlet* sproutlet);
+
+  // Checks if a sproutlet port is taken
+  bool is_port_taken(int port,
+                     Sproutlet* sproutlet);
```
AlexHockey: Why do these parameters have the second argument?

src/sproutletproxy.cpp
```diff
@@ -59,6 +59,10 @@ const pj_str_t SproutletProxy::STR_SERVICE = {"service", 7};
 
 const ForkState NULL_FORK_STATE = {PJSIP_TSX_STATE_NULL, NONE};
 
+const int SERVICE_NAME = 0;
+const int DOMAIN_PART = 1;
+const int USER_PART = 2;
```
AlexHockey: Should go in an `enum`.

src/sproutletproxy.cpp
```diff
@@ -109,6 +140,43 @@ BasicProxy::UASTsx* SproutletProxy::create_uas_tsx()
...
+bool SproutletProxy::is_service_taken(std::string service_name,
+                                      Sproutlet* sproutlet)
+{
+  std::map<std::string, Sproutlet*>::const_iterator it;
+  it = _services.find(service_name);
+  if (it != _services.end())
+  {
+    TRC_ERROR("Can't assign service name \"%s\" to sproutlet \"%s\" because it is taken by sproutlet \"%s\"",
+              service_name.c_str(),
+              sproutlet->service_name().c_str(),
```
AlexHockey: These calls to `c_str` result in dangling pointers.

src/sproutletproxy.cpp
```diff
@@ -109,6 +140,43 @@ BasicProxy::UASTsx* SproutletProxy::create_uas_tsx()
...
+              it->second->service_name().c_str());
+    return true;
+  }
+
+  return false;
+}
```
AlexHockey: To me it feels like "inserting an entry into the map" and "failing to insert and producing a log" are two sides of the same operation, and should probably go together, i.e. this subsection makes the code harder to follow not easier.

src/sproutletproxy.cpp
```diff
@@ -223,129 +279,131 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
...
+                                                    std::string& alias,
+                                                    SAS::TrailId trail)
 {
+  Sproutlet* sproutlet = NULL;
+
+  if (!PJSIP_URI_SCHEME_IS_SIP(uri))
+  {
+    // LCOV_EXCL_START
+    TRC_DEBUG("Sproutlets cannot match non-SIP URIs");
+    return false;
```
AlexHockey: nullptr, surely?

src/sproutletproxy.cpp
```diff
@@ -223,129 +279,131 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
...
+      {
+        sproutlet = it->second;
+        alias = service_name;
+        std::string uri_str = PJUtils::uri_to_string(PJSIP_URI_IN_ROUTING_HDR, uri);
+        SAS::Event event(trail, SASEvent::SPROUTLET_SELECTION_SERVICE_NAME, 0);
+        event.add_var_param(sproutlet->service_name());
+        event.add_static_param(SERVICE_NAME);
+        event.add_var_param(alias);
+        event.add_var_param(uri_str);
+        SAS::report_event(event);
```
AlexHockey: However this chunk of code would be ideal to subfunction.

src/sproutletproxy.cpp
```diff
@@ -223,129 +279,131 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
...
+      {
+        // Check if the part of the hostname before the first '.' matches
+        // a sproutlet.
+        it = _services.find(service_name);
+        if (it != _services.end())
+        {
+          sproutlet = it->second;
+          alias = service_name;
+          std::string uri_str = PJUtils::uri_to_string(PJSIP_URI_IN_ROUTING_HDR, uri);
+          SAS::Event event(trail, SASEvent::SPROUTLET_SELECTION_SERVICE_NAME, 0);
```
AlexHockey: Wrong enum value.

src/sproutletproxy.cpp
```diff
@@ -223,129 +279,131 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
...
-      return true;
+      // Check if the user part matches a sproutlet.
+      service_name = PJUtils::pj_str_to_string(&sip_uri->user);
+      it = _services.find(service_name);
+      if (it != _services.end())
+      {
+        sproutlet = it->second;
+        alias = service_name;
+        std::string uri_str = PJUtils::uri_to_string(PJSIP_URI_IN_ROUTING_HDR, uri);
+        SAS::Event event(trail, SASEvent::SPROUTLET_SELECTION_SERVICE_NAME, 0);
```
AlexHockey: Again, wrong enum value.

src/ut/sproutletproxy_test.cpp
```diff
@@ -65,10 +65,14 @@ template <class T>
 class FakeSproutlet : public Sproutlet
 {
 public:
-  FakeSproutlet(const std::string& service_name, int port, const std::string& uri, const std::string& service_host) :
+  FakeSproutlet(const std::string& service_name, int port, const std::string& uri, const std::string& service_host, bool alias = false) :
```
AlexHockey: It would be more flexible to have alias be a string and to register it only if it's non-empty.

src/ut/sproutletproxy_test.cpp
```diff
@@ -647,6 +651,7 @@ class SproutletProxyTest : public SipTest
     // Create a host alias.
     std::unordered_set<std::string> host_aliases;
     host_aliases.insert("proxy1.homedomain-alias");
+    host_aliases.insert("scscf.proxy1.homedomain");
```
AlexHockey: I doesn't feel like a domain with an `scscf` prefix should be a local hostname.

src/ut/sproutletproxy_test.cpp
```diff
@@ -2275,25 +2289,55 @@ TEST_F(SproutletProxyTest, LocalNonSubscribe)
...
+  //
+  // Check that this holds when we have conflicting information in the three
+  // parts.
+  std::string service_name;
+  std::string uri_str = "sip:b2bua@scscf.proxy1.homedomain";
+  pjsip_sip_uri* uri = (pjsip_sip_uri*)PJUtils::uri_from_string(uri_str, stack_data.pool, PJ_FALSE);
+  pjsip_param* param = PJ_POOL_ALLOC_T(stack_data.pool, pjsip_param);
+  pj_strdup(stack_data.pool, &param->name, &STR_SERVICE);
+  pj_strdup2(stack_data.pool, &param->value, "fwd");
+  pj_list_insert_after(&uri->other_param, param);
```
AlexHockey: Why don't you just put ";service=fwd` on the URI. That would be a lot clearer!

src/ut/sproutletproxy_test.cpp
```diff
@@ -2275,25 +2289,55 @@ TEST_F(SproutletProxyTest, LocalNonSubscribe)
...
-  names = extract_services("sip:scscf.proxy1.homedomain");
+  // The order of precedence for matching is:
+  // - service parameter
+  // - domain part
+  // - user part
+  //
+  // Check that this holds when we have conflicting information in the three
+  // parts.
+  std::string service_name;
+  std::string uri_str = "sip:b2bua@scscf.proxy1.homedomain";
```
AlexHockey: WE should have a port on all these UTIs to check that port gets lowest precedence.

src/ut/sproutletproxy_test.cpp
```diff
@@ -2275,25 +2289,55 @@ TEST_F(SproutletProxyTest, LocalNonSubscribe)
...
+// same service name or port.
+TEST_F(SproutletProxyTest, ConflictingSproutlets)
+{
+  Sproutlet* sproutlet = new FakeSproutlet<FakeSproutletTsxForwarder<false> >("dummy", 0, "sip:dummy.homedomain;transport=tcp", "");
 
-  names = extract_services("sip:alice@otherdomain");
+  // The service name fwd and the port 44444 are taken. Check that this is
+  // actually true.
+  ASSERT_EQ(_proxy->is_service_taken("fwd", sproutlet), true);
+  ASSERT_EQ(_proxy->is_port_taken(44444, sproutlet), true);
```
AlexHockey: I don't like this too much. These methods feel really internal to the proxy. Should we have a `register_sproutlet` method that returns true/false. I'd be happy with testing that in UT.

Hmmm, having said that, `register_sproutlet` would ideally act atomically - i.e. the registration either works completely or fails without making any changes. i.e. it should back out any previous changes that it's made. Not sure how hard that is or if it's worth it.

PR #1800, opened 2017-04-07
Fix extraneous NOTIFYs being sent to other subscribers when a subscriber re-SUBSCRIBEs.

metaswitch-colin-haywood:
- Track the latest CSeq on a SUBSCRIBE received for each subscription.
- Only send NOTIFYs if the bindings have changed, the CSeq has changed or this is a new subscription.

### DIFF
--- a/src/subscriber_data_manager.cpp
+++ b/src/subscriber_data_manager.cpp
@@ -63,6 +63,7 @@ static const char* const JSON_TO_URI = "to_uri";
 static const char* const JSON_TO_TAG = "to_tag";
 static const char* const JSON_ROUTES = "routes";
 static const char* const JSON_NOTIFY_CSEQ = "notify_cseq";
+static const char* const JSON_LAST_CSEQ = "last_cseq";
 
 /// Helper to delete vectors of bindings safely
 void delete_bindings(ClassifiedBindings& cbs)
@@ -1016,6 +1017,7 @@ void SubscriberDataManager::AoR::Subscription::
     writer.String(JSON_TO_URI); writer.String(_to_uri.c_str());
     writer.String(JSON_TO_TAG); writer.String(_to_tag.c_str());
     writer.String(JSON_CID); writer.String(_cid.c_str());
+    writer.String(JSON_LAST_CSEQ); writer.Int(_last_cseq);
 
     writer.String(JSON_ROUTES);
     writer.StartArray();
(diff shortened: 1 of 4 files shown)


src/subscriptionsproutlet.cpp
```diff
@@ -598,6 +598,9 @@ SubscriberDataManager::AoRPair* SubscriptionSproutletTsx::write_subscriptions_to
       // Update/create the subscription.
       subscription->_req_uri = contact_uri;
 
+      pjsip_cseq_hdr* cseq_hdr = (pjsip_cseq_hdr*)pjsip_msg_find_hdr(req, PJSIP_H_CSEQ, NULL);
+      subscription->_last_cseq = cseq_hdr->cseq;
```
richardwhiuk: I don't think we need to store the last CSeq in the story. I think we can just mark the subscription as refreshed locally (defaulting it to false, and resetting it to false when we send a NOTIFY) when we get a SUBSCRIBE.

Duplicate SUBSCRIBEs with the same CSeq will be handled for us by the transaction module.

src/subscriber_data_manager.cpp
```diff
@@ -1596,43 +1606,177 @@ void SubscriberDataManager::NotifySender::send_notifys(
                                int now,
                                SAS::TrailId trail)
 {
+  std::vector<std::string> expired_binding_uris;  
+  std::vector<NotifyUtils::BindingNotifyInformation*> all_bnis;  
+  bool bindings_changed = false;
```
metaswitch-colin-haywood: all_bnis is updated with every BNI we want to include on the NOTIFY, not just the unchanged bindings. bindings_changed is true if at least one entry in all_bnis is a changed binding (CREATED/SHORTENED/EXPIRED but not REGISTERED).

src/subscriber_data_manager.cpp
```diff
@@ -1596,43 +1606,177 @@ void SubscriberDataManager::NotifySender::send_notifys(
...
+
+    // Find the subscription in the original AoR to determine if the current subscription
+    // has been created.
+    SubscriberDataManager::AoR::Subscriptions::const_iterator orig_sub =
+      aor_pair->get_orig()->subscriptions().find(s_id);
+    bool sub_created = (orig_sub == aor_pair->get_orig()->subscriptions().end());
+
+    if (bindings_changed ||
+        sub_created ||
+        (orig_sub->second->_last_cseq != s->_last_cseq))
```
metaswitch-colin-haywood: I've replaced it with a bool but one has to be careful that orig_sub is not NULL/invalid ;) (in the if-statement form we wouldn't have that issue because the check comes after sub_created, which is equivalent to "orig_sub != NULL")

metaswitch-colin-haywood:
Hi Richard,

I've applied the markups - please re-review. I didn't change the UT to check the destination of the NOTIFY - probably one can do this by checking the To header, but untangling the check_OK_and_NOTIFY function looked like a bit of a nightmare.

PR #1804, opened 2017-04-20
[Reviewer: AJH] Sproutlet routing optimization

sathiyan-sivathas:
Alex, please could you review this PR for a Sproutlet routing optimization that should improve Sprout's performance. This should be largely as in the design:
- We swap the order in which we create the SproutletTsx and the SproutletWrapper so that we only create the second if we create the first.
- Sproutlets are able to return a next-hop URI when they aren't interested in a message.
- Existing Sproutlets are enhanced to use this optimization.

One thing that might be controversial is how we modify messages that are owned by the UASTsx. We were worried about modifying this message when retargeting a request to another Sproutlet. I've decided that this isn't an issue (so we can modify the route headers on this message) because:
- We're only modifying the route headers so that won't affect how any delayed 100 Tryings look.
- For any CANCELs the UASTsx sends, we ideally want the route header to match that of the original request (i.e. it should identify the Sproutlet we actually reach). I've changed the behaviour here slightly to make sure this works.
- We're already modifying this message in other ways.
- It's easier ;)


src/ut/registrar_test.cpp
```diff
@@ -672,6 +672,7 @@ TEST_F(RegistrarTest, NotOurs)
 {
   Message msg;
   msg._domain = "not-us.example.org";
+  add_host_mapping("not-us.example.org", "5.6.7.8");
```
AlexHockey: What have you changed that means this is needed? Is it the right change to make?

src/ut/subscription_test.cpp
```diff
@@ -331,6 +331,7 @@ TEST_F(SubscriptionTest, NotOurs)
 {
   SubscribeMessage msg;
   msg._domain = "not-us.example.org";
+  add_host_mapping("not-us.example.org", "5.6.7.8");
```
AlexHockey: Ditto.

src/mangelwurzel/ut/mangelwurzel_test.cpp
```diff
@@ -179,9 +179,11 @@ MATCHER_P(ReqUriEquals, req_uri, "")
 TEST_F(MangelwurzelTest, Rot13)
 {
   MangelwurzelTsx::Config config;
-  EXPECT_CALL(*_helper, original_request());
+  Message msg;
+  pjsip_msg* req = parse_msg(msg.get_request());
   EXPECT_CALL(*_helper, free_msg(_));
-  MangelwurzelTsx mangelwurzel_tsx(_helper, config);
+  MangelwurzelTsx mangelwurzel_tsx(config, req);
```
AlexHockey: Passing the request into the tsx has got to be wrong hasn't it?

src/mangelwurzel/ut/mangelwurzel_test.cpp
```diff
@@ -289,23 +289,22 @@ TEST_F(MangelwurzelTest, CreateRot13)
 {
   Mangelwurzel mangelwurzel("mangelwurzel", 5058, "sip:mangelwurzel.homedomain:5058;transport=tcp");
   Message msg;
+  msg._routes = "Route: <sip:mangelwurzel.homedomain;mangalgorithm=rot13>\r\n";
```
AlexHockey: How come this change?

src/mangelwurzel/ut/mangelwurzel_test.cpp
```diff
@@ -316,25 +315,24 @@ TEST_F(MangelwurzelTest, CreateInvalidMangalgorithm)
   CapturingTestLogger log;
   Mangelwurzel mangelwurzel("mangelwurzel", 5058, "sip:mangelwurzel.homedomain:5058;transport=tcp");
   Message msg;
+  msg._routes = "Route: <sip:mangelwurzel.homedomain;mangalgorithm=invalid>\r\n";
```
AlexHockey: Why do you need this change?

src/sproutletproxy.cpp
```diff
@@ -634,35 +632,25 @@ pj_status_t SproutletProxy::UASTsx::init(pjsip_rx_data* rdata)
...
-        // allow the request to be forwarded. The exception here is if we
-        // rejected the sproutlet selection - which means that we want to
-        // route to the same location but route externally to get the
-        // registrar/subscription modules
-        if (!force_external_routing)
-        {
-          TRC_INFO("Remove top Route header and forward request");
-          pj_list_erase(route);
-        }
-      }
```
sathiyan-sivathas: This is happening somewhere else now so I don't need it. This used to be used for the case where no sproutlet was interested in a message and we wanted to move remove the top route before routing on. That's now handled in get_sproutlet_tsx()

src/authenticationsproutlet.cpp
```diff
@@ -702,127 +842,6 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
...
-    }
-    else
-    {
-      // Unrecognized authentication mode - should never happen. LCOV_EXCL_START
-      assert(!"Unrecognized authentication mode");
-      return PJ_FALSE;
-      // LCOV_EXCL_STOP
-    }
-  }
-}
```
sathiyan-sivathas: I moved this method from the SproutletTsx to the Sproutlet so that's why it shows as a large change.

src/mangelwurzel/ut/mangelwurzel_test.cpp
sathiyan-sivathas: The route used to be provided by a mock call into route_hdr() but since I've replaced that with an actual lookup for the top route, I need to make sure the top route is actually on the message.

src/mangelwurzel/ut/mangelwurzel_test.cpp
sathiyan-sivathas: Ditto

src/mangelwurzel/ut/mangelwurzel_test.cpp
AlexHockey: Ah OK. Fair enough.

src/sproutletproxy.cpp
AlexHockey: Ok, sounds good.

PR #1803, opened 2017-04-19
[Reviewer: Rob] Validate json files in the upload_json script, and log when they fail

johadalin:
This now handles an empty file, and we have better, more consistent error messages.

@steelyvoid , you wanted notifications about this work

PR #1806, opened 2017-04-21
[Reviewer RKD]: Emit json.tool output to screen to aid problem determination

steelyvoid:
Hi Rob

Can you review this tweak to https://github.com/Metaswitch/sprout/pull/1803 to emit the error text from json.tool to the screen?

Tested live against a deliberately broken enum.json file, and the upload_enum_json tool now emits, e.g.

"The configuration file at /etc/clearwater/enum.json contains invalid JSON.
Expecting , delimiter: line 7 column 5 (char 173)
Please provide a valid configuration file, and re-run the command"

(I deleted one of the square brackets in the file.  line7 column 5 is the location of the mismatched remaining bracket)

### DIFF
--- a/sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
+++ b/sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_json
@@ -86,10 +86,11 @@ if [[ ! -s $CONFPATH/$FILENAME ]]
 then
   echo "The configuration file at $CONFPATH/$FILENAME is empty, uploading a blank file"
 else
-  python -m json.tool < $CONFPATH/$FILENAME > /dev/null 2>&1
+  jsoncheck=$(python -m json.tool < $CONFPATH/$FILENAME 2>&1)
   rc=$?
   if [ $rc != 0 ]; then
-    echo "The configuration file at $CONFPATH/$FILENAME contains invalid JSON"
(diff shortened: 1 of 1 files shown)


steelyvoid:
Better than nothing, surely?  At the least the tool has pointed us at exactly where it spotted the mistake.

steelyvoid:
Would you prefer some sort of qualifier around the text, e.g. "The following output may help locate the error in the file - '<text from json.tool>'"

rkd-msw:
That might help - I'm just worried that this error says a , is missing when it isn't, and that's going to confuse users. If you're not concerned, though, maybe I shouldn't be.

steelyvoid:
I think if we were to say

"The configuration file at /etc/clearwater/enum.json contains invalid JSON. The JSON parser reported the following error:
  - "Expecting , delimiter: line 7 column 5 (char 173)"
Correct /etc/clearwater/enum.json to use the documented syntax for this configuration file type, and re-run the command"

Then that reduces the risk that the user will just read the error text and make the change without thinking (though, in practice, if they do change the closing bracket to a comma in this case, they'll get a new error - "Expecting property name: line 8 column 1 (char 174)", so they'll be forced to think about what they're doing eventually :)).

Happy for me to make that change?

steelyvoid:
(sorry - github did something unexpected with the formatting above.  The error text only is supposed to be bulletted)

rkd-msw:
Yes, fine.

PR #1810, opened 2017-04-25
[Reviewer EM] Handle invalid associated identities in registrar

steelyvoid:
Hi Ellie

Can you review the fix to the "invalid Identity in IRS" crash?  I've tested this live with a subscriber with both wildcarded URIs (dropped from the PAU list as expected) and a duff URI (a tel URI with a domain, which was dropped and logged as expected).

I took a look at the other cases where we don't explicitly check whether the URI is parseable, but I think in all the other cases I looked at, the URI being parsed was from either from the message itself, or one of our configuration parameters.  My judgement was that URIs in messages are less likely to be invalid (perhaps bad URIs would be spotted by the PJSIP message parse code if this was the case?) than URIs in Diameter messages, so didn't think it was worth the risk of accidentally introducing a regression, and any crashes due to bad config are no more impactful than Sprout failing to initialise because its got bad config.


src/registrarsproutlet.cpp
```diff
@@ -723,18 +723,30 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
+      if (this_uri != NULL)
+      {
+        pjsip_routing_hdr* pau =
+                        identity_hdr_create(get_pool(rsp), STR_P_ASSOCIATED_URI);
+        pau->name_addr.uri = this_uri;
+        pjsip_msg_add_hdr(rsp, (pjsip_hdr*)pau);
+      }
+      else
+      {
+        TRC_WARNING("Bad associated URI %s", it->c_str());
```
eleanor-merry: Warning feels too high for this log - we'll hit this on every register for a misconfigured subscriber. It's also impossible to see what's gone wrong without the context of the other debug logs. I think this should be debug level and we rely the SAS event.

PR #1811, opened 2017-04-26
[Reviewer EM] Sas log expired bindings

steelyvoid:
Hi Ellie

Can you review this SAS logging enhancement?  I was debugging a problem which turned out to be a clock jump issue on Sprout, which was silently expiring bindings as a result.  It would have been easier to get to the bottom of what was going on if the implicit expiry was logged - hence this change.  I tested live by registering with a softphone, disabling network connectivity and then waiting for the timed dereg.

Note that I wasn't sure which were the best fields to log  - is the AoR, contact URI and reg call-id a sensible subset?

Also, I didn't SAS log expired subscriptions, as I wasn't sure which were the most meaningful fields in a subscriber to log (and its harder for me to test:)).  Would you like to see subscription expiries logged as well?  If so, which fields should be logged in your view?

Thanks!

Steve


src/subscriber_data_manager.cpp
```diff
@@ -465,6 +468,15 @@ int SubscriberDataManager::expire_bindings(AoR* aor_data,
 
     if (b->_expires <= now)
     {
+      if (trail != 0)
+      {
+        SAS::Event event(trail, SASEvent::REGSTORE_BINDING_EXPIRED, 0);
```
eleanor-merry: This SAS log will get raised whenever we look at this data, and we could end up looking at the data multiple times before writing to SAS. I think that's fine, but the wording of the SAS log should make clear that this only excludes the binding from any calculation, and doesn't remove it from the store.

src/subscriber_data_manager.cpp
```diff
@@ -465,6 +468,15 @@ int SubscriberDataManager::expire_bindings(AoR* aor_data,
...
+        event.add_var_param(b->_address_of_record);
+        event.add_var_param(b->_uri);
+        event.add_var_param(b->_cid);
```
eleanor-merry: Can you also include b->_expires and now in the SAS log?

PR #1812, opened 2017-04-27
[Reviewer: AJH] Increment sproutlet tsx failure stat on timeout or transport failure

GAB-MS:
This updates the `rx_fork_error` function so that we log a failure to the outgoing SIP transactions table.

This fixes an issue where we don't log transaction failures in the cases of hitting a transport error or the request timing out.

### DIFF
--- a/src/sproutletproxy.cpp
+++ b/src/sproutletproxy.cpp
@@ -1877,10 +1877,10 @@ void SproutletWrapper::rx_fork_error(pjsip_event_id_e event, int fork_id)
     if (status == PJ_SUCCESS)
     {
       if ((_sproutlet != NULL) &&
-          (_sproutlet->_incoming_sip_transactions_tbl != NULL))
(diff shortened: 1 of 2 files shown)


GAB-MS:
Yes, I intend on getting it reviewed by CW.

I'll look into writing a live test for us, but I intend to get this fix merged in ASAP to unblock @jack-atack's re-test.

GAB-MS:
@AlexHockey, could you please review this? If you're not the right person, could you please pass it on to someone appropriate? Thank you

GAB-MS:
Okay, I've updated it with a unit test that simulates a request timeout. Unfortunately the best way I could find to feed the stats tables in to the Sproutlet was to add them to the FakeSproutlet constructor, with default values of `NULL`, and to specify them in the sproutlet created for the test. If you think this is wrong, please let me know.

Sending back for the UT to be reviewed

AlexHockey:
Looks good to me :smile:

PR #1813, opened 2017-04-27
[Reviewer: Seb] Abort if a plugin fails to load in any way.

richardwhiuk:
Seb,

Currently there is a couple of scenarios where a Sprout plugin can fail to load, but Sprout still continues to run, without the plugin functionality.

This can make it possibly to upgrade a plugin (e.g. the BGCF sproutlet) and break the functionality if the new BGCF fails to load without any alarms being fired, and with monit reporting that everything is working correctly.

This fix resolves that:

- We now mark the plugins as having failed to load if `dlopen` fails on one of the .so files.

- We now mark the plugins as having failed if for a plugin we are unable to get a handle to the sproutlet plugin reference.

- We now mark the plugins as having failed to load if any of the sproutlets are using an unsupported API version.

There's no UTs for this code.

If we fail to load a plugin, we make no attempt to clear up. Given Sprout will simply exit when this happens, this isn't a big deal, but in theory we could attempt to tidy up. The old code didn't do so in any useful way, so I don't think this is a regression, but comments welcome.

On line 153->160, I use a `break; if (foo){ break; }` pattern. I could replace this with a `goto` or an exception, but this seemed a reasonable approach.

### DIFF
--- a/src/pluginloader.cpp
+++ b/src/pluginloader.cpp
@@ -92,66 +92,78 @@ bool PluginLoader::load(std::list<Sproutlet*>& sproutlets)
 
       TRC_STATUS("Attempt to load plug-in %s", p.name.c_str());
 
+      // Clear any dynamic load errors
       dlerror();
+
       p.handle = dlopen(p.name.c_str(), RTLD_NOW);
-      if (p.handle != NULL)
-      {
-        p.plugin = static_cast<SproutletPlugin*>(dlsym(p.handle, "sproutlet_plugin"));
 
-        if (p.plugin != NULL)
-        {
-          std::list<Sproutlet*> plugin_sproutlets;
-          plugins_loaded = p.plugin->load(_opt, plugin_sproutlets);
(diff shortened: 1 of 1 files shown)


src/pluginloader.cpp
```diff
@@ -92,66 +92,78 @@ bool PluginLoader::load(std::list<Sproutlet*>& sproutlets)
...
-            else
-            {
-              // The API version required by the sproutlet is not supported.
-              TRC_ERROR("Sproutlet %s requires unsupported API version %d",
-                        s->service_name().c_str(), s->api_version());
-            }
-          }
-        }
+      // Clear any dynamic load errors
+      dlerror();
```
sebrexmetaswitch: We've just done this above - what's the reasoning for doing this again here? If there is some, can you update the comment so it's more obvious why we need to do this again

richardwhiuk: It's probably overkill, but it's safest to simply call dlerror each time we call a dl*() call to prevent us seeing unrelated errors. To be honest, this is a horrible API.

PR #1814, opened 2017-04-28
[Reviewer: Mat] Don't observe HSS connection in registrar tests

richardwhiuk:
Mat,

https://github.com/Metaswitch/sprout/pull/1635 introduced a bunch of console spam in the logs, as for all of the tests we monitor the HSS connection, however we only care about observing the HSS connection in the multiple registrations tests.

I've split RegistrarTest into two - one which observes the HSS Connection, and one which doesn't.

This also fixes up a bunch of duplicated code in RegistrarRemoteSDMTest.


MatMeredith:
LGTM.

PR #1818, opened 2017-05-03
[Reviewer: SR2] Preserve the user part when routing between sproutlets

sathiyan-sivathas:
Turns out I found the right fix for this.

The live tests fail because the odi token is stripped when routing from the AS via the other sproutlets to the S-CSCF. We need to make sure we don't do that.

I also needed to disable a test that didn't make sense anymore.


sebrexmetaswitch:
Is it helpful to disable rather than remove a test? I think there are already some disabled, so maybe this is what we tend to do, but I'd be inclined to say we should either fix the test or delete it if it doesn't make sense any more (unless you're planning to fix it up later).

I'll take your word on the change itself, if it fixes the bug.

sathiyan-sivathas:
I also think we should delete them, but have disabled it because it's what we tend to do. Don't know why though - I'll ask if there's a good reason, and if not I'll delete it.

I'll merge this for now to unblock you.

PR #1819, opened 2017-05-04
[Reviewer: KH1] Pass through the sos flag on UAR

sathiyan-sivathas:
Converts the "sos" parameter in the contact header of a register to the "sos=true" parameter on the UAR HTTP GET request to homestead.

Together with Metaswitch/homestead#446 this adds support for signalling to the HSS on the UAR that we are handling an emergency REGISTER.

### DIFF
--- a/src/hssconnection.cpp
+++ b/src/hssconnection.cpp
@@ -802,6 +802,7 @@ HTTPCode HSSConnection::get_user_auth_status(const std::string& private_user_ide
                                              const std::string& public_user_identity,
                                              const std::string& visited_network,
                                              const std::string& auth_type,
+                                             const bool& emergency,
                                              rapidjson::Document*& user_auth_status,
                                              SAS::TrailId trail)
 {
@@ -826,6 +827,10 @@ HTTPCode HSSConnection::get_user_auth_status(const std::string& private_user_ide
   if (!auth_type.empty())
   {
     path += "&auth-type=" + Utils::url_escape(auth_type);
+  }
+  if (emergency)
+  {
+    path += "&sos=true";
   }
 
   HTTPCode rc = get_json_object(path, user_auth_status, trail);
--- a/src/ut/hssconnection_test.cpp
+++ b/src/ut/hssconnection_test.cpp
@@ -155,6 +155,7 @@ class HssConnectionTest : public BaseTest
       "</ClearwaterRegData>";
     fakecurl_responses_with_body[std::make_pair("http://10.42.42.42:80/impu/pubid44/reg-data", "{\"reqtype\": \"reg\", \"server_name\": \"server_name\"}")] = CURLE_REMOTE_FILE_NOT_FOUND;
     fakecurl_responses["http://10.42.42.42:80/impi/privid69/registration-status?impu=pubid44"] = "{\"result-code\": 2001, \"scscf\": \"server-name\"}";
+    fakecurl_responses["http://10.42.42.42:80/impi/privid69/registration-status?impu=pubid44&sos=true"] = "{\"result-code\": 2001, \"scscf\": \"server-name\"}";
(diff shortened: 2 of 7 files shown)


src/hssconnection.cpp
```diff
@@ -827,6 +828,10 @@ HTTPCode HSSConnection::get_user_auth_status(const std::string& private_user_ide
   {
     path += "&auth-type=" + Utils::url_escape(auth_type);
   }
+  if (emergency)
+  {
+    path += "&sos=true";
```
KristaHyer: Same comment as for other review request - split this string out and put the components into a shared header?

src/ut/hssconnection_test.cpp
```diff
@@ -526,12 +527,21 @@ TEST_F(HssConnectionTest, CorruptAuth)
...
   CapturingTestLogger log;
   rapidjson::Document* actual;
-  _hss.get_user_auth_status("privid_corrupt", "pubid44", "", "", actual, 0);
+  _hss.get_user_auth_status("privid_corrupt", "pubid44", "", "", false, actual, 0);
   ASSERT_TRUE(actual == NULL);
   EXPECT_TRUE(log.contains("Failed to parse Homestead response"));
   delete actual;
 }
 
+TEST_F(HssConnectionTest, EmergencyAuth)
```
KristaHyer: What are we actually testing here? - please leave a comment.
I think we're checking that the register is succeeding with the emergency flag set to true?

src/ut/icscfsproutlet_test.cpp
```diff
@@ -623,6 +623,72 @@ TEST_F(ICSCFSproutletTest, RouteRegisterHSSCaps)
 }
 
 
+TEST_F(ICSCFSproutletTest, RouteEmergencyRegister)
+{
+  // Tests routing of REGISTER requests when the "sos" flag is set. This test
```
KristaHyer: I love the comment :)

src/ut/icscfsproutlet_test.cpp
```diff
@@ -623,6 +623,72 @@ TEST_F(ICSCFSproutletTest, RouteRegisterHSSCaps)
...
+
+  // Check the RequestURI has been altered to direct the message appropriately.
+  ASSERT_EQ("sip:scscf1.homedomain:5058;transport=TCP", str_uri(tdata->msg->line.req.uri));
+
+  // Check no Route or Record-Route headers have been added.
+  string rr = get_headers(tdata->msg, "Record-Route");
+  string route = get_headers(tdata->msg, "Route");
+  ASSERT_EQ("", rr);
+  ASSERT_EQ("", route);
+
```
KristaHyer: In your checks here I don't see a check for the "sos=true".

Maybe put a comment highlighting it, as it is the purpose of the test.

src/ut/icscfsproutlet_test.cpp
```diff
@@ -623,6 +623,72 @@ TEST_F(ICSCFSproutletTest, RouteRegisterHSSCaps)
...
+  _hss_connection->delete_result("/impi/6505551000%40homedomain/registration-status?impu=sip%3A6505551000%40homedomain&visited-network=homedomain&auth-type=REG;sos=true");
+
+  delete tp;
+}
+
+
 TEST_F(ICSCFSproutletTest, RouteRegisterHSSCapsNoMatch)
 {
   // Tests routing of REGISTER requests when the HSS responses with
```
KristaHyer: I feel we need another test here.

Earlier you said that there could be multiple contact headers, and the emergency registration on any of them will trigger us adding the "sos=true" into the message to Homestead.
In the test above, it looks like there's only one?
I would add a test that has multiple contact headers, with the "sos=true" on only one of them, then make sure we still add it.

sathiyan-sivathas: I edited the existing test so that it had two contact headers where the second had the "sos" parameter. That should get both cases in one test.

PR #1821, opened 2017-05-05
[Reviewer: Richard] Perform PJUtils::send_request callbacks on a worker thread

sebrexmetaswitch:
When sending a request with PJUtils::send_request, the callback may be called on the main PJSIP transport thread (which happens if there is a transport error when trying to send the request).

Since the callback is specified by the sender, it may involve lots of work that we don't want done on the main thread e.g. when we send a register to an AS (if the register fails and the iFC has session terminate, then we deregister the user which is a blocking HTTP request to homestead).

So we want to push these callbacks onto the worker threads instead.

We do this by:

* Creating a `Callback` abstract class that represents a callback that we want to run on the worker threads (and must hold all the data it needs)
* implementing a `RegisterCallback` concrete subclass that we use for registers to an AS (the only case where we use this callback function at the moment)
    * this `RegisterCallback`'s run() method does the same as what was previously specified in the static `send_register_cb` function
* change the `send_request` function to take as a callback a function which build's a `Callback` object that can safely be run on a worker thread
* enhance the worker queue to take either messages or callbacks, and add a function to add callbacks to the queue

This is a fix for issue https://github.com/Metaswitch/clearwater-issues/issues/1452

#### Testing
* tested that register stress that repros the issue before doesn't repro it after the change
* make full_test passes
* live tests pass
* tested sending 400 registers a second through sprout to an AS, both with and without the change, and didn't see a noticeable performance impact (my sprout was between 60 and 70 percent CPU for both)
    * this is what you'd expect given that, in the mainline case, we're already on a worker thread and so the callback is just run directly

### DIFF
--- a/include/thread_dispatcher.h
+++ b/include/thread_dispatcher.h
@@ -63,4 +63,6 @@ void unregister_thread_dispatcher(void);
 pj_status_t start_worker_threads();
 pj_status_t stop_worker_threads();
 
+void add_callback_to_queue(PJUtils::Callback*);
+
 #endif
--- a/include/thread_dispatcher.h
+++ b/include/thread_dispatcher.h
@@ -63,6 +63,8 @@ void unregister_thread_dispatcher(void);
 pj_status_t start_worker_threads();
 pj_status_t stop_worker_threads();
 
+// Add a Callback object to the queue, to be run on a worker thread.
+// This MUST be called from the main PJSIP transport thread.
 void add_callback_to_queue(PJUtils::Callback*);
 
 #endif
--- a/src/registration_utils.cpp
+++ b/src/registration_utils.cpp
@@ -132,9 +132,9 @@ public:
         }
       }
       else
-      // Count all failed registration attempts, not just ones that result in user
-      // being unsubscribed.
       {
+        // Count all failed registration attempts, not just ones that result in
+        // user being unsubscribed.
         if (_reg_data->expires == 0)
         {
           third_party_reg_stats_tables->de_reg_tbl->increment_failures();
@@ -307,7 +307,7 @@ void RegistrationUtils::register_with_application_servers(Ifcs& ifcs,
 static PJUtils::Callback* build_register_cb(void* token, pjsip_event* event)
 {
   RegisterCallback* cb = new RegisterCallback(token, event);
-  return (PJUtils::Callback*)cb;
+  return cb;
 }
 
 static void send_register_to_as(SubscriberDataManager* sdm,
--- a/src/registration_utils.cpp
+++ b/src/registration_utils.cpp
@@ -77,6 +77,88 @@ struct ThirdPartyRegData
   bool is_initial_registration;
 };
 
+class RegisterCallback : public PJUtils::Callback
+{
+  int _status_code;
+  ThirdPartyRegData* _reg_data;
+  std::function<void(ThirdPartyRegData*, int)> _send_register_callback;
+
+public:
+  ~RegisterCallback() override
+  {
+    delete _reg_data; _reg_data = NULL;
+  }
+
+  void run() override
+  {
+    if ((_reg_data->default_handling == SESSION_TERMINATED) &&
+        ((_status_code == 408) ||
+         (PJSIP_IS_STATUS_IN_CLASS(_status_code, 500))))
+    {
+      std::string error_msg = "Third-party REGISTER transaction failed with code " + std::to_string(_status_code);
+      TRC_INFO(error_msg.c_str());
+
+      SAS::Event event(_reg_data->trail, SASEvent::REGISTER_AS_FAILED, 0);
+      event.add_var_param(error_msg);
+      SAS::report_event(event);
+
+      // 3GPP TS 24.229 V12.0.0 (2013-03) 5.4.1.7 specifies that an AS failure
+      // where SESSION_TERMINATED is set means that we should deregister "the
+      // currently registered public user identity" - i.e. all bindings
+      RegistrationUtils::remove_bindings(_reg_data->sdm,
+                                         _reg_data->remote_sdms,
+                                         _reg_data->hss,
+                                         _reg_data->public_id,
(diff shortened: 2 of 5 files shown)


src/registration_utils.cpp
```diff
@@ -77,6 +77,88 @@ struct ThirdPartyRegData
...
+          third_party_reg_stats_tables->init_reg_tbl->increment_successes();
+        }
+        else
+        {
+          third_party_reg_stats_tables->re_reg_tbl->increment_successes();
+        }
+      }
+      else
+      // Count all failed registration attempts, not just ones that result in user
+      // being unsubscribed.
```
richardwhiuk: An existing issue, but this comment should really be inside the brace.

src/registration_utils.cpp
```diff
@@ -222,71 +304,10 @@ void RegistrationUtils::register_with_application_servers(Ifcs& ifcs,
...
-      else
-      {
-        third_party_reg_stats_tables->re_reg_tbl->increment_failures();
-      }
-    }
-  }
-
-  delete tsxdata; tsxdata = NULL;
+  RegisterCallback* cb = new RegisterCallback(token, event);
+  return (PJUtils::Callback*)cb;
```
richardwhiuk: Shouldn't need this cast.

src/registration_utils.cpp
```diff
@@ -433,7 +454,7 @@ static void send_register_to_as(SubscriberDataManager* sdm,
   tsxdata->public_id = served_user;
   tsxdata->expires = expires;
   tsxdata->is_initial_registration = is_initial_registration;
-  pj_status_t resolv_status = PJUtils::send_request(tdata, 0, tsxdata, &send_register_cb);
+  pj_status_t resolv_status = PJUtils::send_request(tdata, 0, tsxdata, &build_register_cb);
```
richardwhiuk: Can we not just build the register callback here and pass it in?

src/thread_dispatcher.cpp
```diff
@@ -139,78 +155,86 @@ static int worker_thread(void* p)
...
 
-  while (rx_msg_q.pop(qe))
+  while (worker_thread_q.pop(qe))
   {
-    pjsip_rx_data* rdata = qe.rdata;
-
-    if (rdata)
+    if (qe.type == MESSAGE)
     {
-      TRC_DEBUG("Worker thread dequeue message %p", rdata);
```
richardwhiuk: This log seems to have got lost - it seems quite useful as well. Probably worth having one saying processing callback.

src/thread_dispatcher.cpp
```diff
@@ -79,13 +79,29 @@ extern "C" {
...
+// An Event on the queue is either a SIP message or a callback
+enum EventType { MESSAGE, CALLBACK };
+
+union Event
 {
-  pjsip_rx_data* rdata;    // received message
-  Utils::StopWatch stop_watch;    // stop watch for tracking message latency
+  pjsip_rx_data* rdata;
+  PJUtils::Callback* callback;
 };
```
richardwhiuk: Hmm. I wonder whether it'd be nice to have a clean std::function queue, and then have the function for SIP RX messages handle the stop watch etc.

src/thread_dispatcher.cpp
```diff
@@ -338,3 +364,18 @@ void unregister_thread_dispatcher(void)
   pjsip_endpt_unregister_module(stack_data.endpt, &mod_thread_dispatcher);
 
 }
+
+void add_callback_to_queue(PJUtils::Callback* cb)
+{
+  // Create an Event to hold the Callback
+  Event queue_event;
+  queue_event.callback = cb;
+  worker_thread_qe qe = { CALLBACK };
```
richardwhiuk: surely `qe = { CALLBACK, queue_event }` would be clearer than two lines.

include/thread_dispatcher.h
```diff
@@ -63,4 +63,6 @@ void unregister_thread_dispatcher(void);
 pj_status_t start_worker_threads();
 pj_status_t stop_worker_threads();
 
+void add_callback_to_queue(PJUtils::Callback*);
```
richardwhiuk: Should add a comment saying this MUST be called on the transport queue, as this function is not at all threadsafe.

src/registration_utils.cpp
richardwhiuk: That may well eliminate the need for a user token as well....

src/thread_dispatcher.cpp
```diff
@@ -79,13 +79,29 @@ extern "C" {
...
+struct worker_thread_qe
+{
+  // The type of the event
+  EventType type;
+
+  // The event itself
+  Event event;
+
+  // A stop watch for tracking SIP message latency
+  Utils::StopWatch stop_watch;
```
richardwhiuk: What's this set to if we are a callback event?

src/thread_dispatcher.cpp
richardwhiuk: Alternatively, it would be neater to have:

```
BaseClass
|
---- Callback
|
---- SIP Message
```

instead of needing the union and the virtual class.

sebrexmetaswitch:
As discussed, we considered changing the callback to be a std::function using closures to capture all the data we need, but we didn't immediately see that we gained anything from that approach.
We also discussed whether we should make the queue more generic, but thought that the queue still represents SIP messages, or callbacks which we have because we previously handled a SIP message. Given that, we didn't think it made sense to make the queue fully generic, and instead stuck with the approach of a `union` that can be either a message or a callback.

I made the other changes you suggested (e.g. moving the stopwatch to the struct that represents a message, since that's the only one of the two that uses it, and now having the union be either a Callback* or a MessageEvent*)

Can you have a look at the newest changes again, and check you think they make sense?

src/thread_dispatcher.cpp
sebrexmetaswitch: I've now reinstated it below

PR #1822, opened 2017-05-08
[Reviewer: EM] Modify Sproutlet SAS logging

sathiyan-sivathas:
Corresponding changes for the SAS PR.

The interesting change here is the one to the SCSCFSproutlet. I've removed the check `is_uri_reflexive()` because it's completely pointless and causes us to duplicate Sproutlet selection SAS logs for in-dialog requests. I think it's pointless because the top route header on a request (if it exists) is guaranteed to be reflexive since that's the URI that caused us to route to the Sproutlet. The SAS log duplication comes because `is_uri_reflexive()` calls through to the Sproutlet selection.

I think we should just remove this API completely because it spams logs that don't make sense except when the Sproutlet is first being selected, and I just don't see why someone would need it. The API comments talk about matching against a URI further down in the routeset, but why would you want to do that? Thoughts?


sathiyan-sivathas:
The latest commit fixes 1405 by passing in the routing number.

PR #1823, opened 2017-05-10
Authenticate non-REGISTERs from UEs that use SIP digest authentication.

AlexHockey:
Ellie, as discussed, please could you review this PR. It adds the ability to authenticate non-REGISTER requests from UEs that authenticate using SIP digest authentication. At the moment this PR does not cover making this work with GR or nonce counts - that will be in separate PRs.

There are some co-req cpp-common and SAS PRs that I'm about to send to you. 

Tested in UT. `make full_test` runs cleanly.


src/main.cpp
```diff
@@ -1131,22 +1133,29 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
...
-        {
-          TRC_INFO("Non-REGISTER authentication set to 'if_proxy_authorization_present'");
-          options->non_register_auth_mode =
-            NonRegisterAuthentication::IF_PROXY_AUTHORIZATION_PRESENT;
-        }
-        else
+        std::vector<std::string> tokens;
+        Utils::split_string(this_arg, ',', tokens);
+
+        for (const std::string& tok: tokens)
```
eleanor-merry: token rather than tok?

src/main.cpp
```diff
@@ -1131,22 +1133,29 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
...
+              NonRegisterAuthentication::IF_PROXY_AUTHORIZATION_PRESENT;
+          }
+          else if (tok == "inital_req_from_reg_digest_endpoint")
+          {
+            options->non_register_auth_mode |=
+              NonRegisterAuthentication::INITIAL_REQ_FROM_REG_DIGEST_ENDPOINT;
+          }
+          else
+          {
+            TRC_ERROR("Invalid token in non-REGISTER authentication field: %s",
```
eleanor-merry: Is exiting the right choice? This will make upgrade harder if we add more options here.

AlexHockey: This is only an issue if we add a new value and somebody sets that value before upgrading their deployment, which shouldn't happen. The downside to not doing any policing is that someone could silently mis-configure their deployment. In the past we've decided to be strict in what we accept (we already have config options that take a certain set of values and if you specify a bad value sprout fails to start) and I'd rather be consistent with those existing options. OK with you @eleanor-merry?

PR #1824, opened 2017-05-10
[Reviewer: Sathiyan] Add default dummy ifc support

eleanor-merry:
This PR adds support for using default IFCs if there are no matching IFCs, rejecting requests if there are no matching IFCs, and adding IFCs with dummy application servers.

Tested live and in UT.


src/scscfsproutlet.cpp
```diff
@@ -561,6 +574,7 @@ void SCSCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
     }
     else
     {
+      // TODO - decide what to do here?
```
sathiyan-sivathas: Do you need to address this?

src/ut/aschain_test.cpp
```diff
@@ -137,53 +187,289 @@ TEST_F(AsChainTest, Basics)
...
+  pjsip_tx_data* tdata = NULL;
+  create_invite(&tdata);
+  std::string server_name;
+
+  pjsip_status_code rc = as_chain_link.on_initial_request(tdata->msg, server_name, 1u);
+  EXPECT_EQ(server_name, "");
+  EXPECT_EQ(rc, PJSIP_SC_OK);
+}
+
+// There are no matching standard IFCs, we're not using default IFCs, and we're
```
sathiyan-sivathas: I think this comment is wrong since you do match a standard IFC.

src/ut/scscf_test.cpp
```diff
@@ -9371,3 +9377,367 @@ TEST_F(SCSCFTest, TestAddStoredPathURI)
...
+  poll();
+  ASSERT_EQ(2, txdata_count());
+
+  // 100 Trying goes back to bono
+  pjsip_msg* out = current_txdata()->msg;
+  RespMatcher(100).matches(out);
+  free_txdata();
+
+  // Check that there's a 404 - no attempt to send via an application server
+  // (it's a 404 the terminating subscriber isn't registered to make this UT
```
sathiyan-sivathas: Missing word

PR #1825, opened 2017-05-12
add SAS logs to barring code

KristaHyer:


src/scscfsproutlet.cpp
```diff
@@ -591,6 +591,16 @@ void SCSCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
         TRC_INFO("Served user is barred so reject the request");
         status_code = _session_case->is_originating() ? PJSIP_SC_FORBIDDEN : PJSIP_SC_NOT_FOUND;
         pjsip_msg* rsp = create_response(req, status_code);
+
+        SAS::Event event(trail(), SASEvent::REJECT_CALL_INVOLVING_BARRED_USER, 0);
+        std::string sess_case = _session_case->is_originating() ? "originating" : "terminating";
```
eleanor-merry: Can you create two SAS logs here, one for originating and one for terminating, rather than passing it as a variable?

include/sproutsasevent.h
```diff
@@ -227,6 +227,8 @@ namespace SASEvent
 
   const int AMBIGUOUS_WILDCARD_MATCH = SPROUT_BASE + 0x0130;
 
+  const int REJECT_CALL_INVOLVING_BARRED_USER = SPROUT_BASE + 0x0135;
```
eleanor-merry: Can this be 140 rather than 135?

src/scscfsproutlet.cpp
```diff
@@ -591,6 +591,16 @@ void SCSCFSproutletTsx::on_rx_initial_request(pjsip_msg* req)
...
+        event.add_var_param(sess_case);
+        std::string served_user = served_user_from_msg(req);
+        event.add_var_param(served_user);
+        std::string st_code = std::to_string(status_code);
+        event.add_var_param(st_code);
```
eleanor-merry: You can pass the status_code in as is by using add_static_param.

src/scscfsproutlet.cpp
KristaHyer: done

src/scscfsproutlet.cpp
KristaHyer: done

PR #1826, opened 2017-05-17
[Reviewer EM] Add (regardless of cache state) clear 'No DNS records returned' SAS log to SIP resolver

steelyvoid:
Hi Ellie

Can you sign off on this small SAS log fix?  This introduces an unconditional SAS log made at the end of the sipresolver "resolve" code (logically paired with the SIPRESOLVE_START log at the start) to log in the clear the condition that there are no DNS records.  The existing logs don't necessarily (depending on cache conditions) jump out at you to say "no DNS records returned", which can make it hard to track down problems whose underlying issue was that expected SIP domain records weren't in DNS.

Tested live


include/sproutsasevent.h
```diff
@@ -89,6 +89,7 @@ namespace SASEvent
   const int SIPRESOLVE_SRV_LOOKUP = SPROUT_BASE + 0x000037;
   const int SIPRESOLVE_A_LOOKUP = SPROUT_BASE + 0x000038;
   const int SIPRESOLVE_IP_ADDRESS = SPROUT_BASE + 0x000039;
+  const int SIPRESOLVE_NO_RECORDS = SPROUT_BASE + 0x000040;
```
eleanor-merry: Can this be 3A, not 40?

steelyvoid:
Doh! Yes it could (you’d have thought I’d been in the industry long enough to spot the difference between a hex and a decimal number).  Ta!

PR #1828, opened 2017-05-18
[Reviewer: EM] URI classification

sathiyan-sivathas:
This PR improves our URI classification to fix Metaswitch/clearwater-issues#1470.

There was a bug in the code that meant that if we didn't have enforce user=phone, and didn't enforce global only lookups, then we would always do an enum lookup on a URI, even if it was clearly not a phone number.

This PR improves our checking of the user part of SIP URIs to make sure that we actually have a phone number before classifying the URI as one. Please can you review.

### DIFF
--- a/src/uri_classifier.cpp
+++ b/src/uri_classifier.cpp
@@ -38,6 +38,11 @@
 #include "stack.h"
 #include "constants.h"
 
+// Regexes that match global and local numbers:
+// - A global number starts with "+" followed by a combination of digits "0-9"
+//   and visual separators ",-()".
+// - A local number can contain a combination of hexdigits "0-9A-F", "*#" and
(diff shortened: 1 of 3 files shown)


src/uri_classifier.cpp
```diff
@@ -33,10 +33,14 @@
  */
 
 #include <vector>
+#include <boost/regex.hpp>
 #include "uri_classifier.h"
 #include "stack.h"
 #include "constants.h"
 
+static const boost::regex CHARS_ALLOWED_IN_GLOBAL_NUM = boost::regex("\\+[0-9,\\-\\(\\)]*");
```
eleanor-merry: Can you comment what these regexes actually match?

PR #1830, opened 2017-05-18
Add GR and NC support to non-REGISTER authentication

AlexHockey:
Hi Ellie, please could you review this PR that adds GR and nonce count support to non-REGISTER digest authentication. I've tested this in UT.


src/authenticationsproutlet.cpp
```diff
@@ -1371,3 +1274,154 @@ void AuthenticationSproutletTsx::on_rx_response(pjsip_msg* rsp, int fork_id)
...
+
+      if (backup_impi_obj != NULL)
+      {
+        if (!backup_impi_obj->auth_challenges.empty())
+        {
+          // We found an IMPI in a backup store that has some challenges. Copy
+          // them over and return (remembering to delete the backup IMPI object).
+          TRC_DEBUG("Found IMPI in backup store");
+          impi_obj->auth_challenges = std::move(backup_impi_obj->auth_challenges);
+          delete backup_impi_obj; backup_impi_obj = NULL;
```
eleanor-merry: Do you not want to break here?

src/main.cpp
```diff
@@ -2198,19 +2237,44 @@ int main(int argc, char* argv[])
...
-  // relevant challenge is sent.
+  // sproutlet.
   if (impi_store_location != "")
   {
-    impi_memstore = (Store*)new TopologyNeutralMemcachedStore(impi_store_location,
-                                                              astaire_resolver,
-                                                              astaire_comm_monitor);
-    impi_store = new ImpiStore(impi_memstore, opt.impi_store_mode);
+    // Use memcached store.
+    TRC_STATUS("Using memcached compatible store with binary protocol");
```
eleanor-merry: I don't understand this TRC statement...

src/main.cpp
```diff
@@ -2198,19 +2237,44 @@ int main(int argc, char* argv[])
...
+    local_impi_data_store = (Store*)new TopologyNeutralMemcachedStore(impi_store_location,
+                                                                      astaire_resolver,
+                                                                      false,
+                                                                      astaire_comm_monitor);
+    local_impi_store = new ImpiStore(local_impi_data_store, opt.impi_store_mode);
+
+    // Only set up remote IMPI stores if some has been configured, and we need
```
eleanor-merry: *have

src/main.cpp
```diff
@@ -2198,19 +2237,44 @@ int main(int argc, char* argv[])
...
+        Store* remote_data_store = (Store*)new TopologyNeutralMemcachedStore(*it,
+                                                                             astaire_resolver,
+                                                                             true,
+                                                                             remote_astaire_comm_monitor);
+        remote_impi_data_stores.push_back(remote_data_store);
+        remote_impi_stores.push_back(new ImpiStore(remote_data_store, opt.impi_store_mode));
+      }
+    }
   }
   else
```
eleanor-merry: Does this else branch make sense? Isn't local store purely test code?

include/cfgoptions.h
```diff
@@ -170,15 +170,21 @@ struct options
   bool                                 disable_tcp_switch;
   std::string                          chronos_hostname;
   std::string                          sprout_chronos_callback_uri;
+  int                                  non_reg_chal_lifetime_s;
```
AlexHockey: I'm about to remove this option, so I'll leave as is for the moment.

src/authenticationsproutlet.cpp
```diff
@@ -1089,78 +1037,30 @@ void AuthenticationSproutletTsx::on_rx_initial_request(pjsip_msg* req)
...
-            TRC_DEBUG("Storing challenge in order to challenge non-REGISTER requests");
-            new_expiry = calculate_challenge_expiration_time(req);
-          }
+          TRC_DEBUG("Storing challenge because nonce counts are supported");
+          auth_challenge->expires = calculate_challenge_expiration_time(req);
         }
-
-        Store::Status store_status;
-        do
+        else if (is_register &&
```
AlexHockey: I thought that we didn't use brackets around single identifiers (as there is no way for operator precedence to trip you up)?

src/authenticationsproutlet.cpp
AlexHockey: Yep, that would make sense!

src/main.cpp
AlexHockey: I think that was copied from somewhere else (maybe where we set up the reg-stores), but I agree it's confusing so I'll fix it up.

src/authenticationsproutlet.cpp
eleanor-merry: It means it doesn't line up with the other options at the same level, so it's slightly harder to parse.

src/main.cpp
AlexHockey: I think we've always allowed sprout to run without memcached and just store stuff locally. I'm not sure I've ever actually made use of it, but the reg-store does the same thing...

src/authenticationsproutlet.cpp
AlexHockey: Ok fair enough.

PR #1831, opened 2017-05-19
Fix missing include

huws:
Recent changes to `uri_classifier` broke the Houdini build, due to a missing include.  The sprout build wasn't broken, so presumably `string` got included elsewhere when building sprout.

@sathiyan-sivathas - FYI


sathiyan-sivathas:
Sounds right - thanks for making the fix!

PR #1834, opened 2017-05-23
[Reviewer: Richard] Getting orig-served-user is a read-only operation

dimbleby:
Self-explanatory. 

(Some of us like to scatter out sproutlet code with `const` at every opportunity.  I ran into this method when it stopped me from doing so.)


dimbleby:
Thanks.  (re authorship: we're sharing a VM and occasionally forget what's currently in our `.gitconfig`!)

PR #1837, opened 2017-05-24
[Reviwer: Krista] Allow IMPUs that are part of wildcards to be barred

eleanor-merry:
This PR allows IMPUs that are part of wildcards to be barred. Tested in UT.


src/associated_uris.cpp
```diff
@@ -48,6 +49,14 @@ void AssociatedURIs::add_uri(std::string uri,
                              bool barred)
 {
   _associated_uris.push_back(uri);
+  update_barring_status(uri, barred);
```
KristaHyer: I don't like this name so much, as it tripped me up in the code.
"update" makes it sound like the uri has already been added, so I spent a while looking for where that had been done.
I'd like to change the name to something more like "add_barring_status_to_map" (even if that is a bit of a mouthful).

src/ut/scscf_test.cpp
```diff
@@ -1288,35 +1416,122 @@ TEST_F(SCSCFTest, TestBadScheme)
...
+                                   "call",
+                                   "REGISTERED",
+                                   IMS_SUB_BARRED_WILDCARD);
+  Message msg;
+  msg._route = "Route: <sip:homedomain;orig>";
+  doSlowFailureFlow(msg, 403);
+}
+
+// Test that a call from an IMPU that belongs to a barred wildcarded public
+// identity, when multiple wildcarded public identities are present in the
```
KristaHyer: I'd change this from "when multiple wildcarded.." to "when other unbarred wildcarded...", to make this comment clearer.
(same for other test using IMS_SUB_BARRED_MULTIPLE_WILDCARD)

PR #1838, opened 2017-05-25
[Reviewer: Krista] Add ifcs scripts

eleanor-merry:
Updates to the shared/default IFCs to match the spec.


src/ut/difcservice_test.cpp
```diff
@@ -130,12 +141,13 @@ TEST_F(DIFCServiceTest, ReloadInvalidDIFCFile)
   delete root_reload; root_reload = NULL;
 }
 
-// Test that reloading a Default iFC file with an invalid file doesn't cause the
-// valid entries to be lost.
+// Test that reloading a Default iFC file with valid file doesn't destroy any
```
KristaHyer: This doesn't seem to describe the test....
You say here that you're reloading a valid file, but on line 156 you say you're reloading an invalid file.

eleanor-merry: The comment was wrong (I'd fixed the test description but missed the comment within it).

PR #1840, opened 2017-05-30
[Reviewer: KH1] Add diagnostics to match call barring spec

sathiyan-sivathas:
Adds the 2 sas logs, 2 pd logs and 1 stat that support asked for.

I had to change what variables we passed around for this to work so there's quite a large diff.


include/sproutsasevent.h
```diff
@@ -208,6 +208,8 @@ namespace SASEvent
 
   const int REJECT_CALL_FROM_BARRED_USER = SPROUT_BASE + 0x0140;
   const int REJECT_CALL_TO_BARRED_USER = SPROUT_BASE + 0x0141;
+  const int OMIT_ID_FROM_P_ASSOC_URI = SPROUT_BASE + 0x0142;
```
KristaHyer: I'd like to get the word "barred" in here somewhere.
Maybe "OMIT_BARRED_ID_FROM_P_ASSOC_URI"?
That is a bit of a mouthful though - I'll leave it up to you whether or not you choose to make this change.

include/subscriber_data_manager.h
```diff
@@ -571,14 +572,14 @@ class SubscriberDataManager
   /// succeeds, this returns true.
   ///
   /// @param aor_id               The AoR to retrieve
-  /// @param unbarred_irs_impus   The unbarred IMPUs in the Implicit Registration
+  /// @param associated_uris      The associated IMPUs in the Implicit Registration
```
KristaHyer: Change the description of "associated_uris" to match the other descriptions of this param?

src/notify_utils.cpp
```diff
@@ -118,7 +120,28 @@ pj_xml_node* notify_create_reg_state_xml(
...
-  // Iterate over the IRS, inserting a registration element for each one
+  // Log any URIs that have been left out of the P-Associated-URI because they
+  // are barred.
+  std::vector<std::string> barred_uris = associated_uris->get_barred_uris();
+  if (!barred_uris.empty())
+  {
+    std::stringstream ss;
+    std::copy(barred_uris.begin(), barred_uris.end(), std::ostream_iterator<std::string>(ss, ","));
+    std::string list = ss.str();
+    if (!list.empty())
```
KristaHyer: Can we ever hit the case where the list is empty?
If we can, shouldn't we not log in that case?
(same comment in the other SAS log)

PR #1843, opened 2017-06-08
[Reviewer: Mat] Update PD logs for split-storage

sebrexmetaswitch:


include/sprout_pd_definitions.h
```diff
@@ -78,7 +78,7 @@ static const PDLog CL_SPROUT_NO_SI_CSCF
   "Neither P-CSCF, S-CSCF nor I-CSCF functionality is enabled on this node.",
```
MatMeredith: There are still a few references to node in this file that I think we could easily avoid?   E.g. here: "The Sprout process is running but not providing I-CSCF, S-CSCF or P-CSCF functionality". (or maybe "is not running because ..."?)

Are there not more major issues with this log though?   E.g. it's terribly written, the grammar sucks and it doesn't mention BGCF?  Also is Weatherwax or something based on Sprout?   What happens when people write other applications based on Sprout?

sebrexmetaswitch: I'm not convinced that we need to remove the "node" here, since the log is accurate, but no harm in doing so.

As for not allowing for other applications based on Sprout - indeed, but the log is raised when we detect that none of those 3 are enabled, so the description is accurate.

I'll fix up the grammar since I'm changing this one anyway

PR #1844, opened 2017-06-09
[Reviewer=HJS, AJH]  Add ability for the SCSCF to apply MMF to messages routed to an Application Server

jack-atack:
Add MMF service infrastructure, accessible by the SCSCFSproutlet to invoke MMF before, and after, routing a message to an Application Server.

Points of note:

- There are placeholders in place for the adding of the Route headers required to invoke MMF before and after routing a message to an Application Server.   This will be added in a follow-up Pull Request early next week.  This code will only ever be hit if MMF is turned on, and a valid mmf_targets.json config file is present.
- Locking consists of a write lock on config update, and a read lock on querying whether to invoke SCSCF pre-AS and post-AS.
- The overall structure and shape of this code was based largely off the existing DIFCService class.
- Methods defined in the .h files are marked 'inline' for consistency with the rest of the Sprout codebase.
- The _mmf_config member variable may have several keys pointing to the same MMFTarget value.  This is to enable easy access to the MMF config (and whether we have any) for a given Application Server.


include/mmfservice.h
```diff
@@ -0,0 +1,72 @@
...
+    return _mmf_config.count(address);
+  }
+
+  Alarm* _alarm;
+  std::map<std::string, MMFTarget::ptr> _mmf_config;
+  std::string _configuration;
+  Updater<void, MMFService>* _updater;
+
+  // Mark as mutable to flag that this can be modified without affecting the
+  // external behaviour of the calss, allowing for locking in 'const' methods.
```
huws: "class"

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_mmf_targets_json
```diff
@@ -0,0 +1,15 @@
...
+# of the source code repository by which you are accessing this code, then
+# the license outlined in that COPYING file applies to your use.
+# Otherwise no rights are granted except for those provided to you by
+# Metaswitch Networks in a separate written agreement.
+
+FILENAME=mmf_targets.json
+KEY=mmf_targets_json
+
+# Pass optional argument directly to upload_json, which checks for --allow-large
+# option that allows file upload larger than advisable
```
huws: Do we really want to allow very large files here?  I think we should avoid it, as we do for the `s-cscf.json`, etc.

sprout-base.root/usr/share/clearwater/infrastructure/alarms/sprout_alarms.json
```diff
@@ -298,6 +298,29 @@
...
+        },
+        {
+            "index": 1013,
+            "name": "SPROUT_MMF_STATUS",
+            "cause": "UNDERLYING_RESOURCE_UNAVAILABLE",
+            "levels": [
+                {
+                    "severity": "CLEARED",
+                    "details": "The configuration file for MMF is valid.",
+                    "description": "sprout: The MMF configuration file is valid.",
```
huws: Isn't it a bit misleading to talk about the "MMF configuration file".  I would assume that this referred to the `mmf_rules.json` file.  Why not call it the "MMF targets file"?

include/mmfservice.h
```diff
@@ -0,0 +1,72 @@
+/**
+ * @file mmfservice.h Support for MMF function.
+ *
+ * Copyright (C) Metaswitch Networks
```
huws: Missing copyright year.

src/mmf.cpp
```diff
@@ -0,0 +1,102 @@
...
+    TRC_ERROR("Invalid 'name' field in MMF configuration.  The 'name' "
+              "field must be present, and must be a string");
+    JSON_FORMAT_ERROR();
+  }
+}
+
+void MMFTarget::parse_pre_as(const rapidjson::Value& config)
+{
+  TRC_DEBUG("Reading pre-AS");
+  if (config.HasMember("pre-AS") && config["pre-AS"].IsBool())
```
huws: I thought that Chris had changed the capitalisation of these parameters.  Please check that you're using strings that match Chris's current opinion (same comment for "post-AS", obviously).

src/mmfservice.cpp
```diff
@@ -0,0 +1,194 @@
+/**
+ * @file mmfservice.cpp The MMF Config handler.
+ *
+ * Copyright (C) Metaswitch Networks
```
huws: Copyright year

src/mmfservice.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    TRC_DEBUG("Taking write lock on mmf config");
+
+    // Take a write lock on the houdini config, in RAII style.
+    boost::lock_guard<boost::shared_mutex> write_lock(get_mmf_rw_lock());
+
+    // Now that we have the mmf config lock, free the memory from the old
+    // mmf config objects, and start pointing at the new ones.
+    TRC_DEBUG("Delete old MMF config.");
+    _mmf_config = mmf_config;
```
huws: Your comments suggest that you haven't understood what the code you've written is doing.  This line is performing a copy of the map - nothing gets deleted.  Perhaps you intended to store a smart pointer to the map, so that pointing at a new value would result in the old map being implicitly freed?

Doing it with a smart pointer seems like a much nicer way of managing this, otherwise you need to concern yourself about whether the contents of the map is being copied, or just pointed to, and whether that is OK.

src/mmfservice.cpp
```diff
@@ -0,0 +1,194 @@
...
+  for (rapidjson::Value::ConstValueIterator mmf_it = mmf_targets.Begin();
+       mmf_it != mmf_targets.End();
+       ++mmf_it)
+  {
+    // Throws a JsonFormatError if the config is invalid
+    MMFTarget::ptr config(new MMFTarget(*mmf_it));
+
+    for (std::string address : config->get_addresses())
+    {
+      if (mmf_config.count(address))
```
huws: `count()` returns an integer value - I'd prefer to see an explicit numerical comparison here - i.e. 
```cpp
if (mmf_config.count(address) > 0)
```

src/mmfservice.cpp
```diff
@@ -0,0 +1,194 @@
...
+  boost::lock_guard<boost::shared_mutex> read_lock(get_mmf_rw_lock());
+
+  if (has_config_for_address(address) && get_address_config(address)->apply_pre_as())
+  {
+    return true;
+  }
+  else
+  {
+    return false;
+  }
```
huws: You've effectively written "if true, return true".  Save yourself 7 lines of code, by just returning the value of the Boolean expression.

Ah - or maybe this is supposed to represent some code that you'll add in future - in which case, you should put in a comment to explain the bizarre-looking code.

src/mmfservice.cpp
```diff
@@ -0,0 +1,194 @@
...
+  boost::lock_guard<boost::shared_mutex> read_lock(get_mmf_rw_lock());
+
+  if (has_config_for_address(address) && get_address_config(address)->apply_post_as())
+  {
+    return true;
+  }
+  else
+  {
+    return false;
+  }
```
huws: Ditto

src/ut/mmfservice_test.cpp
```diff
@@ -0,0 +1,230 @@
+/**
+ * @file MMFservice_test.cpp Tests support of MMF configuration
+ *
+ * Copyright (C) Metaswitch Networks
```
huws: Year

huws:
Jack - I've just noticed, it looks like you haven't included the new plugin file for `mmf_targets.json`.  The key thing I'm interested in here is what default config will be provided by the plugin.

jack-atack:
the plugin will be in the etcd-plugins repository (hence why it's not included in the PR), although I've built debians with it in.

The default value will be :  ```{} ```

include/mmfservice.h
jack-atack: I copied this from difcservice.h - this is also missing it!

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_mmf_targets_json
jack-atack: We do the same here as the s-cscf.json file - the 2 upload scripts are identical aside from the FILENAME and KEY

sprout-base.root/usr/share/clearwater/infrastructure/alarms/sprout_alarms.json
jack-atack: yep.  The effect was actually incorrect too for the CRITICAL case - we'll reuse old config if it was previously valid so i'll change that too

src/mmfservice.cpp
jack-atack: I think the comment is wrong.  The MMFService is intentionally initiated, stored, and deleted identically to how existing config (such as the DIFCService)  is stored on the SCSCF

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_mmf_targets_json
huws: OK - fine.

src/mmfservice.cpp
jack-atack: We spoke, I'll change the member variable such that it is a shared pointer to the original planned map.  We want to avoid calling the copy constructor where possible.

jack-atack:
Thanks for offering to review this Alex.

As mentioned above, there are placeholder comments in the route_to_as() method for future code to add the relevant route headers if we wish to invoke MMF for a request before and/or after an Application Server.

include/mmfservice.h
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  Alarm* _alarm;
+  std::shared_ptr<MMFService::MMFMap> _mmf_config;
+  std::string _configuration;
+  Updater<void, MMFService>* _updater;
+
+  // Mark as mutable to flag that this can be modified without affecting the
+  // external behaviour of the class, allowing for locking in 'const' methods.
+  mutable boost::shared_mutex _mmf_rw_lock;
```
AlexHockey: Erm... this is slightly scary. Why do you need interior mutability?

include/mmfservice.h
```diff
@@ -0,0 +1,72 @@
...
+
+  /// Updates the MMF Config
+  void update_config();
+
+  std::shared_ptr<MMFService::MMFMap> read_config(rapidjson::Document& doc);
+
+  const bool apply_mmf_pre_as(std::string address);
+  const bool apply_mmf_post_as(std::string address);
+
+  inline boost::shared_mutex& get_mmf_rw_lock() {return _mmf_rw_lock;};
```
AlexHockey: You only call this method from within the class, so what is the point of it?

src/mmfservice.cpp
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+  }
+
+  return mmf_config;
+}
+
+const bool MMFService::apply_mmf_pre_as(std::string address)
+{
+  boost::lock_guard<boost::shared_mutex> read_lock(get_mmf_rw_lock());
```
AlexHockey: Aren't you taking the lock exclusively here (i.e. for writing)?

include/mmfservice.h
```diff
@@ -0,0 +1,72 @@
...
+
+private:
+  MMFService(const MMFService&) = delete;  // Prevent implicit copying
+
+  // These methods are private to prevent access to the mmf config without
+  // first taking a read_lock
+  inline const MMFTarget::ptr get_address_config(std::string address)
+  {
+    return _mmf_config->at(address);
```
AlexHockey: Ah I see - you're relying on the caller to have called `has_config_for_address` first. That's a paradigm called "look before you leap" and it is generally better to try to do the thing you want (i.e. get the config) and cope with it not being there. Given you're returning a pointer that just falls out (I think)

src/scscfsproutlet.cpp
```diff
@@ -1562,11 +1570,30 @@ void SCSCFSproutletTsx::route_to_as(pjsip_msg* req, const std::string& server_na
...
 
+    // If we are configured to apply MMF prior to forwarding the request to
+    // the AS, add the necessary route header
+    if (_scscf->mmfservice()->apply_mmf_pre_as(server_name))
+    {
+      // This is a placeholder for code which is currently being written.
+      // The will involve the addition of route header to invoke MMF on the
+      // way to the AS
+    }
+
```
AlexHockey: I initially thought these were in the wrong order, but then I spotted that we're adding headers to the top of the request. Worth a comment?

src/ut/mmfservice_test.cpp
```diff
@@ -0,0 +1,230 @@
...
+  EXPECT_TRUE(MMF.apply_mmf_pre_as("as.test.domain"));
+  EXPECT_TRUE(MMF.apply_mmf_post_as("second.test.domain"));
+  EXPECT_FALSE(MMF.apply_mmf_post_as("10.10.0.2"));
+}
+
+
+// In the following tests we have various invalid/unexpected MMF Json
+// files.
+// These tests check that the correct logs are made in each case; this isn't
+// ideal as it means the tests are quite fragile, but it's the best we can do.
```
AlexHockey: Could you check the ENT logs that are generated? These all have unique IDs.

src/mmf.cpp
```diff
@@ -0,0 +1,102 @@
...
+  {
+    TRC_DEBUG("Read pre-as: %d", config["pre-as"].GetBool());
+    _pre_as = config["pre-as"].GetBool();
+  }
+  else
+  {
+    TRC_ERROR("Invalid 'pre-as' field in MMF configuration.  The 'pre-as' "
+              "field must be present, and must be a boolean");
+    JSON_FORMAT_ERROR();
+  }
```
AlexHockey: If we're going to extend the possible scopes in which MMF gets invoked, I don't think we should be making `post-as` and `pre-as` mandatory. Otherwise you have to specify them each time you add a new entry. Let's default them to false instead.

AlexHockey: In fact, it they are just bools, why not have a `scopes` property that is an array of scope strings that apply to this target?

src/scscfsproutlet.cpp
```diff
@@ -1562,11 +1570,30 @@ void SCSCFSproutletTsx::route_to_as(pjsip_msg* req, const std::string& server_na
       pj_strdup2(get_pool(req), &orig_param->value, "");
       pj_list_insert_after(&odi_uri->other_param, orig_param);
     }
+
     PJUtils::add_top_route_header(req, odi_uri, get_pool(req));
 
+    // If we are configured to apply MMF on requests forwarded on by the AS
+    // add the necessary route header
+    if (_scscf->mmfservice()->apply_mmf_post_as(server_name))
```
AlexHockey: It looks like the server name is a URI. Does that mean that MMF should be doing URI comparisons rather than plain string comparisons (e.g. `sip:alice@exmaple.com;lr;orig` and `sip:alice@exmaple.com;orig;lr` are the same URI but different strings.

include/mmfservice.h
AlexHockey: These boost classes are mostly in the C++ standard library now. Please can you use the `std::` versions?

include/mmf.h
```diff
@@ -0,0 +1,49 @@
+/**
+ * @file mmf.h  MMF target configuration options
```
jack-atack: You're right, this should be either be called mmftarget.h  (and have an associated mmftarget.cpp file) or be included as part of the mmfservice files

include/mmfservice.h
```diff
@@ -0,0 +1,72 @@
...
+  ~MMFService();
+
+  typedef std::map<std::string, MMFTarget::ptr> MMFMap;
+
+  /// Updates the MMF Config
+  void update_config();
+
+  std::shared_ptr<MMFService::MMFMap> read_config(rapidjson::Document& doc);
+
+  const bool apply_mmf_pre_as(std::string address);
```
jack-atack: It means "Do I want to apply MMF before/ after routing to the Application Server?".  Agree this needs a comment

include/mmfservice.h
```diff
@@ -0,0 +1,72 @@
...
+    return _mmf_config->at(address);
+  }
+
+  inline const bool has_config_for_address(std::string address)
+  {
+    return _mmf_config->count(address);
+  }
+
+  Alarm* _alarm;
+  std::shared_ptr<MMFService::MMFMap> _mmf_config;
```
jack-atack: The use of shared pointer is for both memory management, and to prevent us ever 'copying' the MMFMap object, or any MMFTarget object.

On config update - we read in the new config file, creating a new MMFMap, and then update the config used by the MMFService by pointing the _mmf_config member variable at the new MMFMap we have created (and away from the MMFMap representing the old config).   

The rw_lock is to ensure that the SCSCF  is unable to read the _mmf_config (to work out whether to route a request to an AS or not) at the same time that the MMFService is updating the _mmf_config member variable

include/sprout_pd_definitions.h
```diff
@@ -588,4 +588,35 @@ static const PDLog1<const char *> CL_SPROUT_TERM_PARTY_BARRED
   "None"
 );
 
+
+static const PDLog CL_SPROUT_MMF_FILE_INVALID
+(
+  PDLogBase::CL_SPROUT_ID + 67,
+  LOG_NOTICE,
+  "The file detailing when to apply MMF is invalid.",
+  "The /etc/clearwater/mmf_targets.json file, defining when to apply MMF, is invalid due to invalid JSON or missing elements.",
```
jack-atack: Yep, there's a few things wrong with these PDLogs as you've pointed out

src/mmfservice.cpp
jack-atack: We only need read access - these functions are essentially "read the mmf_targets.json file.  Is there any config for the passed in address?  If so, should I invoke MMF before/after routing to the address?"

src/scscfsproutlet.cpp
jack-atack: Yep, I think you're right

include/mmfservice.h
```diff
@@ -0,0 +1,72 @@
...
+  const bool apply_mmf_pre_as(std::string address);
+  const bool apply_mmf_post_as(std::string address);
+
+  inline boost::shared_mutex& get_mmf_rw_lock() {return _mmf_rw_lock;};
+
+private:
+  MMFService(const MMFService&) = delete;  // Prevent implicit copying
+
+  // These methods are private to prevent access to the mmf config without
+  // first taking a read_lock
```
jack-atack: The SCSCF Sproutlet can only access the  _mmf_config  via the public `apply_mmf_pre_as()` and `apply_mmf_post_as()` methods, both of which take the `rw_lock`

include/mmf.h
```diff
@@ -0,0 +1,49 @@
...
+  void parse_pre_as(const rapidjson::Value& config);
+
+  void parse_post_as(const rapidjson::Value& config);
+
+  inline const bool apply_pre_as() const {return _pre_as;};
+  inline const bool apply_post_as() const {return _post_as;};
+  inline const std::vector<std::string>& get_addresses() {return _addresses;};
+
+private:
+  MMFTarget(const MMFTarget&) = delete;  // Prevent implicit copying
```
jack-atack: We don't want to call the copy constructor on non-trivial and potentially large objects such as MMFTarget, or MMFService; we want to be passing around references or pointers instead.

Accidentally calling the copy constructor on non-trivial objects has caused weird bugs and unexpected behaviour for us in the past, and this enables us to catch unintended copying at compile time.

src/mmf.cpp
```diff
@@ -0,0 +1,102 @@
...
+    TRC_DEBUG("Read post-as: %d", config["post-as"].GetBool());
+    _post_as = config["post-as"].GetBool();
+  }
+  else
+  {
+    TRC_ERROR("Invalid 'post-as' field in MMF configuration.  The 'post-as' "
+              "field must be present, and must be a boolean");
+    JSON_FORMAT_ERROR();
+  }
+}
```
jack-atack: It would save a lot of code, but you'd lose the precise error output as to why the file is invalid, which is really beneficial (particular as the mmf_targets.json file could become quite large if you have lots of Application Servers)

The JsonFormatError object only contains details of where in the codebase the error was hit, rather than where in the config file the error was hit

include/sprout_pd_definitions.h
```diff
@@ -588,4 +588,35 @@ static const PDLog1<const char *> CL_SPROUT_TERM_PARTY_BARRED
...
+  "Sprout will not apply MMF to any calls.",
+  "Follow the documentation to create this file correctly."
```
jack-atack: This is a copy of the last sentence of the `PDLog CL_SPROUT_SCSCF_FILE_INVALID` log.  Should this also be amended?

include/mmfservice.h
huws: I think the point that @AlexHockey was making was that we could just drop the lock altogether.

While updates to `shared_ptr` values aren't guaranteed to be immediately synchronised out between all threads (I think), they are atomic, so any thread reading the value will either get the entire old map or the entire new map (it's fine not to know which of the two a given request will get while we're changing between two sets of config).  There is no risk of getting half a map, because we complete the construction of the new map before we assign it to the member variable, and we won't start destroying the old map until every `shared_ptr` instance has stopped pointing to it.

Alex - please could you confirm that I've correctly understood what you were saying?

Jack - apologies - I've previously given you some incorrect advice, here.

include/mmf.h
AlexHockey: I agree about preventing it for `MMFService` but `MMFTarget` is pretty much a plain-old-data type so I don't see the benefit of preventing copying.

include/mmfservice.h
AlexHockey: Ah OK. In that case they are private because they are simply not part of the API, and I think what your really trying to say is that "these methods assume the lock is already held". Might be worth rewording the comment, as it confused me at least! :-)

include/mmfservice.h
AlexHockey: Yep, what @huws said.

include/sprout_pd_definitions.h
AlexHockey: Apparently we have loads of logs with the same wording :-(. I'm happy for you to leave this as is.

src/mmf.cpp
AlexHockey: Right but your only logging the error to `/var/log/sprout` which customers don't look at anyway. So the log will only be consumed by Metaswitch staff who have access to the code (or know someone that does). 

If you're really set on having the extra information in the logs, I think it would be better to extend `json_parse_utils.h` to allow the user to specify a string if the error operation failed that gets stuffed into the exception.

src/mmfservice.cpp
AlexHockey: I get that you only need the lock for reading, but I think the code you've written takes it for writing. [boost::lock_guard](http://www.boost.org/doc/libs/1_51_0/doc/html/thread/synchronization.html#thread.synchronization.locks.lock_guard) calls `lock()` on the underlying [boost::shared_mutex](http://www.boost.org/doc/libs/1_51_0/doc/html/thread/synchronization.html#thread.synchronization.mutex_types.shared_mutex) which gets exclusive access. You want to use [boost::shared_lock](http://www.boost.org/doc/libs/1_51_0/doc/html/thread/synchronization.html#thread.synchronization.locks.shared_lock) to get read access. 

Thought this is all a moot point if you can do away with the lock entirely :-)

include/mmf.h
```diff
@@ -0,0 +1,49 @@
...
+#include <iostream>
+#include <memory>
+#include "utils.h"
+#include "json_parse_utils.h"
+
+/// A representation of an entry in the mmf_targets.json file
+class MMFTarget
+{
+public:
+  typedef std::shared_ptr<MMFTarget> ptr;
```
jack-atack: I think it is much nicer, and more readable, to refer to a `MMFTarget::ptr` than to say an `MMFService::MMFTargetptr`.  I'm happy to change it if you insist however

include/mmf.h
jack-atack: I don't think there is a use case currently where we would want to copy it though, instead of returning a const reference or a pointer (especially given the size of the `MMFTarget` object may increase if we add more function and use cases).   Being prompted by the compiler that you are doing a copy forces you to think about whether it is the right thing to be doing (and if it was I'd be happy for this line to be removed)

Again, I'm happy to change it if you insist however

include/mmfservice.h
jack-atack: The locking is being scrapped - we're utilising the shared pointer's atomic properties, and a comment will be added to the `_mmf_config` variable to ensure future coders are aware of how to correctly use/access the config lock-free

include/mmfservice.h
jack-atack: Agree.  I'll add an appropriate comment, and add try/catch statements around where this is currently used

include/mmfservice.h
jack-atack: This is being scrapped

include/sprout_pd_definitions.h
jack-atack: Ok

src/mmfservice.cpp
jack-atack: Ah!  Good spot.  Although as you say the lock is being done away with 🎉

include/mmf.h
AlexHockey: I think that pragmatism should win out, but given that `MMFTargets` are not used outside the `MMFService`, you never have to type the `MMFService::` qualifier. This means you're going from `MMFTarget::ptr` to `MMFTargetPtr` which seems roughly the same from a readability perspective.

include/mmf.h
AlexHockey: I don't care too much about this.

src/mmf.cpp
jack-atack: Seems I've gotten the wrong end of the stick here - on discussion with Chris/Huw we **do** want these to be optional, and to default to false if they are not present.  I'll change this

In terms of the `scopes` idea - this is more of a HLD question.  I personally slightly prefer the current config; I like the idea of having `pre-as` and `post-as` as keys rather than values, plus it allows you to toggle MMF on and off without having to delete and re-enter a magic string

AlexHockey: If they're optional I care less about the idea about having a `scopes` property.

src/scscfsproutlet.cpp
jack-atack: The MMF should be doing a domain comparison.  I think we want to be extracting the domain from the server_name (using the `domain_from_uri` PJUtils method) and passing this value in to the MMF Service functions

sprout-base.root/usr/share/clearwater/infrastructure/alarms/sprout_alarms.json
```diff
@@ -298,6 +298,29 @@
...
+                    "cause": "The configuration file at /etc/clearwater/mmf_targets.json is valid.",
+                    "effect": "The S-CSCF will be able to apply any configured MMF rules to calls.",
+                    "action": "No action."
+                },
+                {
+                    "severity": "CRITICAL",
+                    "details": "There is at least one invalid entry in the MMF targets file.",
+                    "description": "sprout: The MMF targets file is invalid.",
+                    "cause": "The configuration file at /etc/clearwater/mmf_targets.json has at least one invalid entry.",
+                    "effect": "New config will not be loaded.  If good config was previously loaded, the SPN will continue to use it.",
```
jack-atack: If we make valid MMF config invalid, the updater will ensure that we continue to use the old, valid, config (this was an intentional design decision).  As you say however, if Sprout is restarted it reloads all config, and there will be no MMF applied to any requests.   

I'll update the alarm text accordingly

src/mmf.cpp
jack-atack: The mmf_targets.json file is something we would expect customers to be editting themselves as they go (as opposed to setting it up on commissioning and then running with it).  

We definitely want a way for them to work out what was wrong with the json file if it is invalid - I'd previously assumed this would be through ERROR logs, but if this isn't the case then perhaps we need to have a rethink

sprout-base.root/usr/share/clearwater/infrastructure/alarms/sprout_alarms.json
jack-atack: How does the below sound?

```
"effect": "New config will not be loaded.  If good config was previously loaded, the S-CSCF will continue to use it.  However, if Sprout is restarted, the S-CSCF will lose previously loaded config, and will not apply MMF to any requests.",
```

include/sprout_pd_definitions.h
```diff
@@ -588,4 +588,35 @@ static const PDLog1<const char *> CL_SPROUT_TERM_PARTY_BARRED
...
+);
+
+static const PDLog CL_SPROUT_MMF_FILE_MISSING
+(
+  PDLogBase::CL_SPROUT_ID + 68,
+  LOG_NOTICE,
+  "The file detailing when to apply MMF is not present.",
+  "The /etc/clearwater/mmf_targets.json file, defining when to apply MMF, does not exist.",
+  "Sprout will not apply MMF to any calls.",
+  "If you are expecting clearwater-config-manager to be managing this file, check that it is running and that there are no ENT logs relating to it or clearwater-etcd.  If you are managing /etc/clearwater/mmf_targets.json manually, follow the documentation to create it."
```
jack-atack: The line about `clearwater-config-manager` appears in almost every invalid config file PDLog in the file.  

The logs also all have the same '4 lines' syntax, which seem to vaguely map to the `details`, `description`, `effect`, `action` fields in the alarm that gets raised, so I'm a bit hesitant to split it out into a 5th sentence without knowing anything about how PDLogs work (a lot of the invalid config logs suffer from a very long final sentence)

sprout-base.root/usr/share/clearwater/infrastructure/alarms/sprout_alarms.json
jack-atack: I'll also be using this text in the PDLogs (which previously incorrectly said that Sprout will not apply MMF to any requests)

src/ut/mmfservice_test.cpp
jack-atack: They wouldn't check that the correct ERROR log is raised (which is important as it gives the necessary information as to what is wrong with the json file, and exactly how to resolve it). 

However, as mentioned above I think there is a discussion to be had as to whether ERROR trace is the right place to include this information

sprout-base.root/usr/share/clearwater/infrastructure/alarms/sprout_alarms.json
AlexHockey: :+1:

include/sprout_pd_definitions.h
AlexHockey: If the existing alarms refer tot he config manager, just leave it in (sorry, I did't realize how much of the PR was based on existing stuff). When I talked about splitting the line I meant just splitting the text, not adding  a new field. In C 

```c
char* x = "Hello world";
```
is the same as 
```c
char* x = "Hello "
          "world";
```

But I don't care that much.

AlexHockey:
@jack-atack you're good to go.

PR #1845, opened 2017-06-09
Optionally log Ralf HTTP bodies to SAS

huws:
In order to reduce the volume of data logged to SAS, we've made the logging of Ralf HTTP bodies optional.  By default, the bodies will not be logged (the messages appear in SAS with the generic string `<Body present but not logged>`).

If required, the parameter `http_acr_logging=Y` can be set in `shared_config` to turn this logging back on.

This PR covers the Sprout-side changes.  There are some similar changes to Ralf - see [here](https://github.com/Metaswitch/ralf/pull/278).


huws:
Alex - please could you review these changes to turn off SAS logging of Ralf HTTP bodies by default?

src/main.cpp
```diff
@@ -446,6 +448,8 @@ static void usage(void)
        "     --dummy-app-server <app server URI>\n"
        "                            If any IFC has an application server that matches the one defined here, \n"
        "                            then the IFC is skipped over.\n"
+       "     --http_acr_logging     Whether to include the bodies of ACR HTTP requests when they are logged \n"
```
AlexHockey: The option has hyphens but this help text uses underscores.

PR #1849, opened 2017-06-13
[Reviewer: Ellie] Add a new option to control whether Sprout should send to Ralf

richardwhiuk:
As discussed in, https://github.com/Metaswitch/ralf/pull/276, we think we should add a new config option to separate Ralf being configured, and whether Sprout should actually send to Ralf.

It can be set to Y or N to control the behavior, and is defaulted so that the existing behavior is fine over upgrade.

### DIFF
--- a/sprout-base.root/etc/init.d/sprout
+++ b/sprout-base.root/etc/init.d/sprout
@@ -86,6 +86,9 @@ get_settings()
         signaling_dns_server=127.0.0.1
         scscf_node_uri=""
 
+        # Should we send to Ralf? Y, N, or X => base on ralf hostname
+        sprout_ralf_enabled=X
+
         # Enable no-ENUM TEL URI translation fallback by default for backwards compatibility
         default_tel_uri_translation="Y"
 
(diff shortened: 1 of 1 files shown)


sprout-base.root/etc/init.d/sprout
```diff
@@ -131,7 +134,11 @@ get_daemon_args()
           [ -z "$xdms_hostname" ] || xdms_hostname_arg="--xdms=$xdms_hostname"
         fi
 
-        [ -z "$ralf_hostname" ] || ralf_arg="--ralf=$ralf_hostname"
+        if [ "$sprout_ralf_enabled" == "Y" ] || \
+           [ "$sprout_ralf_enabled" == "X" -a -n "$ralf_hostname" ]
+        then
+          ralf_arg="--ralf=$ralf_hostname"
```
eleanor-merry: As discussed, this will do odd things if sprout_ralf_enabled is set but ralf_hostname isn't.

Why not default sprout_ralf_enabled to Y, then only set ralf_arg if sprout_ralf_enabled is Y and ralf_hostname is set to something. This still works over upgrade and I think it's simpler (and we don't add a third value to an option that should be Y/N).

You've agreed to set sprout_ralf_enabled to N in shared config in systems that don't have ralf enabled.

rkd-msw:
Do we need this in V11.0.2?

PR #1852, opened 2017-06-15
[Reviewer: AJH] Fix leaked UACTsx

sathiyan-sivathas:
Mark the UAC as ready for deletion when a client is unresponsive. This fixes a issue I saw where the the UTs would leak memory when run with JUSTTEST="SCSCFTest*".

I've also removed the coverage exclusion statements since this code has clearly been covered since it caused a leak.


AlexHockey:
LGTM. I'm going to merge this now so that we can get the fix back to ST.

PR #1853, opened 2017-06-16
[Reviewer=HJS, AJH] Add Route headers to invoke MMF function

jack-atack:
Pull Request which fills in the placeholders left in by PR #1844 

**Points of note:**
- There has been a slight change to the MMF config API - the scscfsproutlet now obtains a shared pointer to the config for the particular application server. This change is for locking reasons; the initial intention was for the scscfsproutlet code to query the mmfservice to obtain details about the AS's config.  However, there would be a short time window between queries during which the mmf config may change, which could cause unexpected behaviour and exceptions.  Taking a shared_ptr to the AS's config at the time gets around this

- As per the HLD, when doing MMF prior to routing the request to an AS we want to route to the MMF server using the sprout cluster domain and the MMF port.  When doing MMF after routing the request to an AS we want to route to the MMF server using the private IP of the node, and the MMF port.  The cleanest way to do this was to set these uri's up in the SCSCF plugin, and pass these in to the SCSCFSproutlet constructor.  

- Currently, the MMF server will be on the same port as the BGCF, and we will route to it using the sprout hostname and the bgcf port.  This isn't ideal, and there are plans to add MMF configuration options to shared_config in future

- The constants.h changes have been merged in early in the PR #1855


include/mmftargets.h
```diff
@@ -23,20 +23,21 @@ class MMFTarget
 public:
   MMFTarget(const rapidjson::Value& config);
 
+  inline const std::string get_mmfcontext() const {return _name;};
+  inline const bool apply_mmf_pre_as() const {return _pre_as;};
+  inline const bool apply_mmf_post_as() const {return _post_as;};
```
huws: I worry slightly about the naming here.  A method called `apply_blah()` sounds like it contains the processing required to apply "blah".  I'm struggling to come up with a better alternative, though.  `is_mmf_pre_as()`?  `should_apply_mmf_pre_as()`?  `needs_mmf_pre_as()`?  I don't really like them much  better.  Up to you if you want to change.

src/scscfsproutlet.cpp
```diff
@@ -2407,3 +2415,96 @@ pjsip_msg* SCSCFSproutletTsx::get_base_request()
...
+void SCSCFSproutletTsx::add_mmf_pre_as_route_header(pjsip_msg* req,
+                                                    std::string mmfcontext,
+                                                    pj_str_t as_transport_param)
+{
+  // Route the message to the MMF service running on this Sprout node
+  // Use the MMF port as the identifier, as we do not have a separate
+  // MMF sproutlet.
+  // A lot of this is currently hard-coded; this will change in future
+  pj_str_t home_domain = stack_data.default_home_domain;
+  std::string mmf_uri = "sip:sprout." + PJUtils::pj_str_to_string(&home_domain) + ":5053";
```
huws: You shouldn't hard-code this number - that's a configurable port.  We default to 5053, but a customer can change this in `shared_config` via the `bgcf` parameter.

src/scscfsproutlet.cpp
```diff
@@ -2407,3 +2415,96 @@ pjsip_msg* SCSCFSproutletTsx::get_base_request()
...
+  // A lot of this is currently hard-coded; this will change in future
+  std::string node_host(stack_data.local_host.ptr, stack_data.local_host.slen);
+
+  if (Utils::parse_ip_address(node_host) == Utils::IPV6_ADDRESS)
+  {
+    node_host = "[" + node_host + "]";
+  }
+
+  // Hard code the MMF port for the meantime
+  std::string mmf_node_uri = "sip:" + node_host + ":5053";
```
huws: Again - should be read from config.

src/scscfsproutlet.cpp
```diff
@@ -2407,3 +2415,96 @@ pjsip_msg* SCSCFSproutletTsx::get_base_request()
...
+                                pjsip_uri_clone(get_pool(req), mmf_pj_uri);
+
+  // Use same transport as AS, in case it can only cope with one.
+  post_as_uri->transport_param = as_transport_param;
+
+  add_mmf_uri_parameters(post_as_uri, "mmf", mmfcontext, "post-as", req);
+
+  TRC_DEBUG("Adding top route header for post-as MMF");
+  PJUtils::add_top_route_header(req, post_as_uri, get_pool(req));
+}
```
huws: Does this method need to be separate from the `add_mmf_pre_as_route_header()`.  It is definitely the case that this one needs to route to the IP address of this particular node, but that would also work for the pre-AS case.

include/mmftargets.h
jack-atack: I've changed to `should_apply_mmf_xx_as()`, as you're right and I think it's more intuitive.  There were explanatory comments alongside the previous MMFService methods, which I've also moved to this file.

src/scscfsproutlet.cpp
jack-atack: Yep, the mmf_uri will be moved to the scscfplugin, and passed in as an argument to the SCSCFConstructor, as that's the right long-term approach, and we might as well do it now

src/mmfservice.cpp
```diff
@@ -147,29 +147,15 @@ std::shared_ptr<MMFService::MMFMap> MMFService::read_config(rapidjson::Document&
...
-{
-  try
-  {
-    return (get_address_config(address)->apply_post_as());
-  }
-  catch (std::out_of_range)
-  {
-    TRC_DEBUG("No MMF config for the address: %s", address.c_str());
-    return false;
+    return nullptr;
```
AlexHockey: We generally try to avoid using try/except in our C++ code. As C++ doesn't have checked exceptions its easy for an exception to escape the code that is able to handle it, which means it goes unhandled and can cause a crash. That can't happen here but I's rather we didn't try/excepts unless they provide a big ergonomic win, which I don't think we get here.

src/scscfsproutlet.cpp
```diff
@@ -2407,3 +2472,36 @@ pjsip_msg* SCSCFSproutletTsx::get_base_request()
...
   }
 }
+
+
+void SCSCFSproutletTsx::add_mmf_uri_parameters(pjsip_sip_uri* mmf_uri,
+                                               pj_str_t as_transport_param,
+                                               std::string namespace_param,
+                                               std::string mmfscope_param,
+                                               std::string mmfcontext_param,
+                                               pjsip_msg* req)
```
AlexHockey: It would be better to pass the pool in as the last argument rather than the request. Doing this proves to the reader that the function is only going to modify the URI that's been passed in.

src/scscfsproutlet.cpp
```diff
@@ -2407,3 +2472,36 @@ pjsip_msg* SCSCFSproutletTsx::get_base_request()
...
+  pj_strdup(get_pool(req), &mmfcontext_p->name, &STR_MMFCONTEXT);
+  pj_list_insert_before(&mmf_uri->other_param, mmfcontext_p);
+  pj_strdup2(get_pool(req), &mmfcontext_p->value, mmfcontext_param.c_str());
+
+  // Insert the mmfscope parameter
+  TRC_DEBUG("Adding mmfscope parameter %s", mmfscope_param.c_str());
+  pjsip_param* mmfscope_p = PJ_POOL_ALLOC_T(get_pool(req), pjsip_param);
+  pj_strdup(get_pool(req), &mmfscope_p->name, &STR_MMFSCOPE);
+  pj_list_insert_before(&mmf_uri->other_param, mmfscope_p);
+  pj_strdup2(get_pool(req), &mmfscope_p->value, mmfscope_param.c_str());
```
AlexHockey: Adding a parameter must be a really common operation and I was surprised to find there isn't already a utility function for it. But this code is the straw that broke the camels back. I think you should refactor this code into a subfunction in `PJUtils` and just call it from this function.

src/scscfsproutlet.cpp
```diff
@@ -2407,3 +2472,36 @@ pjsip_msg* SCSCFSproutletTsx::get_base_request()
     return original_request();
   }
 }
+
+
+void SCSCFSproutletTsx::add_mmf_uri_parameters(pjsip_sip_uri* mmf_uri,
+                                               pj_str_t as_transport_param,
+                                               std::string namespace_param,
```
AlexHockey: Why do you need to pass the `namespace` parameter into the function if it's always set to "mmf"?

src/ut/scscf_test.cpp
```diff
@@ -416,6 +416,8 @@ class SCSCFTest : public SipTest
                                           "sip:127.0.0.1:5058",
                                           "",
                                           "sip:bgcf@homedomain:5058",
+                                          "sip:11.22.33.44:5053;transport=tcp",
+                                          "sip:44.33.22.11:5053;transport=tcp",
```
AlexHockey: I wouldn't have expected the plugin to specify the transport, so it's a bit weird for the UTs to do so.

src/ut/scscf_test.cpp
```diff
@@ -10129,3 +10131,406 @@ TEST_F(SCSCFTest, MixedRealAndDummyApplicationServer)
...
+  // MMF sends a 100 Trying
+  string fresp = respond_to_txdata(current_txdata(), 100);
+  inject_msg(fresp, &tpMMFpreAS);
+
+  const pj_str_t STR_ROUTE = pj_str("Route");
+
+  // Ensure the pre-as header was added as expected
+  pjsip_hdr* preas_hdr = (pjsip_hdr*)pjsip_msg_find_hdr_by_name(out, &STR_ROUTE, NULL);
+  std::string preas_uri = PJUtils::get_header_value(preas_hdr);
+  EXPECT_EQ(preas_uri, "<sip:11.22.33.44:5053;transport=UDP;lr;namespace=mmf;mmfcontext=pre-as;mmfscope=PreASOnly>");
```
AlexHockey: A small thing: this check is not resistant to various things like the parameters being in a different order. It would be better to split this into several checks for each parameter, by doing something like `EXPECT_THAT(preas_uri, MatchesRegex(";namespace=mmf[;>]")`

src/ut/scscf_test.cpp
```diff
@@ -10129,3 +10131,406 @@ TEST_F(SCSCFTest, MixedRealAndDummyApplicationServer)
...
+  pj_list_erase(preas_hdr);
+
+  // Ensure the AS header was added as expected
+  pjsip_hdr* as_hdr = (pjsip_hdr*)pjsip_msg_find_hdr_by_name(out, &STR_ROUTE, NULL);
+  std::string as_uri = PJUtils::get_header_value(as_hdr);
+  EXPECT_EQ(as_uri, "<sip:pre.as.only.mmf.test.server:56789;transport=UDP;lr>");
+  pj_list_erase(as_hdr);
+
+  // Simulate the request being routed to the AS, and then routed from the AS
+  // back to the SCSCF
```
AlexHockey: Don't you need to strip some route headers to simulate this?

AlexHockey: Ditto below.

src/ut/scscf_test.cpp
AlexHockey: Ditto below.

include/constants.h
```diff
@@ -111,6 +111,9 @@ const pj_str_t STR_CONTENT_LENGTH = pj_str((char*)"Content-Length");
 const pj_str_t STR_SERVICE = pj_str((char*)"service");
 const pj_str_t STR_USERNAME = pj_str((char*)"username");
 const pj_str_t STR_NONCE = pj_str((char*)"nonce");
+const pj_str_t STR_NAMESPACE = pj_str((char*)"namespace");
+const pj_str_t STR_MMFCONTEXT = pj_str((char*)"mmfcontext");
+const pj_str_t STR_MMFSCOPE = pj_str((char*)"mmfscope");
```
AlexHockey: I know these match the HLD, but I saw some discussion over whether the MMF parameters were too verbose. I didn't see the resolution, so could you check with Chris where that discussion ended up before you merge the code?

jack-atack: I spoke to Chris - we're going with the current design.  However, you're right they are too verbose, and it's something that is going to be looked at in the future

src/ut/scscf_test.cpp
jack-atack: The route headers we're stripped - by the pj_list_erase() function.  The comment is just out of place

src/ut/scscf_test.cpp
jack-atack: I wasn't aware of that syntax.  That's a nicer way of implementing this

src/ut/scscf_test.cpp
jack-atack: You're right, copy-and-paste oversight

src/ut/scscf_test.cpp
AlexHockey: Lol! I missed that completely. Sorry!

include/constants.h
AlexHockey: That's fine, thanks for getting back to me.

src/mmfservice.cpp
jack-atack: Yep, I can take a shared_ptr to the MMF config, and then use `find`.  

The reason for the try/ catch was to prevent the admittedly microscopic window between checking for a key, and then extracting the value (during which config may have been updated).  However, taking a fresh shared_ptr to the MMFMap gets around this

src/scscfsproutlet.cpp
jack-atack: Have done.  I'll add a  `generic add_parameter_to_sip_uri`, which can either add a value-less parameter, or a param: value pair to a URI

PR #1854, opened 2017-06-16
[Reviewer: Richard] Fail the sprout build if plugins fail to build. 

AlexHockey:
Richard, please can you review this fix that makes the sprout build fail if one of the plugins fails to build. I've done this in a way that means plugin building plays nicely with `-k`.

Tested as follows:

* Clone `greeter` into the plugins directory and break it. 
* Run `make` from the top level. It exists with error code 2. 
* Run `make deb` from the top level. It exists with error code 2. 
* Clone another `greeter` into `greeter2`
* Run `make plugins-build`. This fails without building `greeter2`
* Run `make plugins-build -k`. This fails but builds `greeter2`
* Fix greeter. Run `make plugins-build`. This succeeds (exits with 0 error code)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,26 +58,49 @@ distclean: $(patsubst %, %_distclean, ${SUBMODULES}) sprout_distclean
 	rm -rf ${ROOT}/usr
 	rm -rf ${ROOT}/build
 
+#
+# Plugin handling.
+#
+
+# Generate a list of all the plugins available.
+PLUGINS := $(patsubst plugins/%,%,$(shell find plugins -mindepth 1 -maxdepth 1 -type d ))
+
+# Macro to define rules for building plugins.
+#
+# $1 - The name of the plugin to build.
+# $1 - The make target (e.g build, test, clean, ...)
+define plugin_template
+.PHONY: plugin-$1-$2
+
+# Note that the plugins are not required to have a 'build' target, but assume
+# the default target performs a build.
+plugin-$1-$2:
+	make -C plugins/$1 $$(filter-out build,$2)
+endef
+
 .PHONY: plugins-build
-plugins-build:
-	find plugins -mindepth 1 -maxdepth 1 -type d -exec ${MAKE} -C {} \;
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -58,26 +58,54 @@ distclean: $(patsubst %, %_distclean, ${SUBMODULES}) sprout_distclean
...
+
+# Macro to define the rule for building a single target for a single plugin.
+#
+# Parameters:
+#   $1 - The name of the plugin to build.
+#   $2 - The make target (e.g build, test, clean, ...)
+#
+# Note that the plugins are not required to have a 'build' target, but assume
+# the default target performs a build.  For this reason we never call "make
+# build" for a plugin, and always just call "make" instead.
```
richardwhiuk: I know this is what old code did, but is this actually sensible? Normally make on it's own defaults to `all` not `build`.

Makefile
```diff
@@ -58,26 +58,54 @@ distclean: $(patsubst %, %_distclean, ${SUBMODULES}) sprout_distclean
...
-plugins-deb:
-	find plugins -mindepth 1 -maxdepth 1 -type d -exec ${MAKE} -C {} deb \;
-
-.PHONY: plugins-deb-only
-plugins-deb-only:
-	find plugins -mindepth 1 -maxdepth 1 -type d -exec ${MAKE} -C {} deb-only \;
-
+#
+# Plugin handling.
+#
```
richardwhiuk: I'm undecided whether I like this style of heading. It's quite a good separator, but it also eats quite a lot of vertical space.

Makefile
AlexHockey: We discussed this and your top-level comment. The conclusion was that long-term the right thing to is come up with a build pipeline that means plugins can be built without needing a sprout checkout. Once this happens the sprout Makefile won't need knowledge of plugins at all. When/if this happens this code will be deleted so there isn't much point finessing it, especially as it's unlikely to change much.

Makefile
AlexHockey: Well with that resounding opinion I'm going to leave it as it is :-)

PR #1858, opened 2017-06-21
[Reviewer=HJS] Fix following ST of mmf_targets JSON file

jack-atack:
The fixes following ST of the new mmf_targets.json file.  They include:

- Amend the previous ERROR trace printed with every invalid config, to better explain the behaviour

- Only allow A-Z, a-z, 0-9, -, and _ characters in the 'name' field.  The value of this field is sent as the 'mmfcontext' URI parameter when requests are forwarded to the MMF server, so the characters allowed here must be restricted

- Fix up some STATUS and DEBUG logs

### DIFF
--- a/src/mmftargets.cpp
+++ b/src/mmftargets.cpp
@@ -1,5 +1,5 @@
 /**
- * @file mmf.cpp  class representing MMF Target configuration options
+ * @file mmftargets.cpp  class representing MMF Target configuration options
  *
  * Copyright (C) Metaswitch Networks 2017
  * If license terms are provided to you in a COPYING file in the root directory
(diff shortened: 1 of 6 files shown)


src/mmftargets.cpp
```diff
@@ -25,10 +25,26 @@ MMFTarget::MMFTarget(const rapidjson::Value& config):
...
-    TRC_DEBUG("Read name: %s", config["name"].GetString());
-    _name = config["name"].GetString();
+    std::string name = config["name"].GetString();
+    TRC_DEBUG("Read name: %s", name.c_str());
+
+    // The _name is used as the mmfcontext URI parameter in requests sent for
+    // MMF processing.  We only allow A-Z, a-z, 0-9, - and _.
+    for(char& s : name)
+    {
+      if(!(isalnum(s) || s == '_' || s == '-'))
```
huws: Feels like this might have been easier to do with a regex.  Also, it looks like you'll allow zero-length values - is that intentional?

jack-atack: Good spot, zero-length values will result in a value-less parameter `;mmfcontext;` which may cause issues on the other end

PR #1860, opened 2017-06-23
[Reviewer: Graeme] Save off original request for in-dialog transactions too

rkd-msw:
Currently we only initialise `_unmodified_request` for initial requests, which causes carnage on in-dialog requests when we try and use `_unmodified_request` in processing the response.

@GAB-MS


rkd-msw:
Reverting - this broke MangelwurzelTest.InDialogReq. I'll investigate and fix on Monday.

PR #1862, opened 2017-06-26
[Reviewer: Alex] plugins-deb prior to deb-only

richardwhiuk:
This means that plugins-deb-only won't get called before plugins-deb when performing make deb.

This only works with `make -j 1` (as we need to guarantee dependency ordering), but the full fix requires rotating our make file targets to ensure Debian build targets correctly depend on their dependencies.


AlexHockey:
:+1:

PR #1869, opened 2017-06-29
[Reviewer: Alex] Test we can route * and # in To headers

richardwhiuk:
Alex,

This validates that the I-CSCF can route messages with a * or a # in the To header. It requires in order to pass https://github.com/Metaswitch/pjsip-upstream/pull/66.


richardwhiuk:
Alex, I agreed that it didn't make much sense in I-CSCF Sproutlet test, and we decided that a better home would be to have a suite of SIP Parser tests. I've renamed the custom headers UT as discussed, and added a test in there for expediency.

Passing back to confirm that you are happy with the updated approach.

PR #1870, opened 2017-07-04
rename mmf_context to mmf_target

ChrisRees:
Update mmf_context to be called mmf_target, to reduce confusion between config files.
Also fixing up the calls to add_mmf_uri_parameters, to pass the correct parameters in in the correct order (previously context and scope were mixed up) see https://github.com/Metaswitch/houdini/issues/812

### DIFF
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -1632,15 +1632,15 @@ void SCSCFSproutletTsx::route_to_as(pjsip_msg* req, const std::string& server_na
 
       add_mmf_uri_parameters(post_as_uri,
                              as_uri->transport_param,
-                             server_mmf_config->get_mmfcontext(),
                              "post-as",
+                             server_mmf_config->get_target_name(),
                              get_pool(req));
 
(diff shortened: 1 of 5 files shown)


src/scscfsproutlet.cpp
```diff
@@ -1632,15 +1632,15 @@ void SCSCFSproutletTsx::route_to_as(pjsip_msg* req, const std::string& server_na
 
       add_mmf_uri_parameters(post_as_uri,
                              as_uri->transport_param,
-                             server_mmf_config->get_mmfcontext(),
                              "post-as",
+                             server_mmf_config->get_target_name(),
```
benritchie: Chris - why wasn't this bug caught by the UT's?
I think we should try to understand that, and improve the UT's so they do catch this, and similar bugs.

ChrisRees: The UT's were checking the incorrect values were set, hence the extra switch in the UT's rather than just renaming mmf_context.

They will now spot this failure.

PR #1871, opened 2017-07-04
[Reviewer: Alex] Validate config files before uploading

Ying-metaswitch:
A python script to validate the json file against the schema, and print errors.
ENUM, S-CSCF and BGCF upload scripts in this repo, and DNS upload script in /clearwater-etcd, all call into this validator. 

Ran mainline test in Sprout, successfully printed two errors:
```
Uploading bgcf.json is aborted as the file fails the following format check:
               4 is not of type u'string'
               u'++2466' does not match u'^\\+{0,1}[0-9]*$'
```

Schema is written in https://github.com/Metaswitch/clearwater-etcd-plugins/pull/36


sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/validate_config.py
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# Copyright (C) Metaswitch Networks 2017
+# If license terms are provided to you in a COPYING file in the root directory
+# of the source code repository by which you are accessing this code, then
+# the license outlined in that COPYING file applies to your use.
+# Otherwise no rights are granted except for those provided to you by
+# Metaswitch Networks in a separate written agreement.
+
+import jsonschema
```
AlexHockey: jsonschema is not part of the python standard library so how are you making sure that it is installed? In fact, does this script need to be in it's own virtual environment?

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/upload_enum_json
```diff
@@ -9,9 +9,13 @@
 
 FILENAME=enum.json
 KEY=enum_json
+SCHEMA=/usr/share/clearwater/clearwater-config-manager/scripts/config_validation/enum_schema.json
 
 . /usr/share/clearwater/utils/check-root-permissions 1
 
+# Validate config json against schema for syntactic errors
+python /usr/share/clearwater/clearwater-config-manager/scripts/validate_config.py $SCHEMA $FILENAME
```
AlexHockey: Same comment goes for all the other upload scripts.

AlexHockey:
One more comment - if you have large JSON file it might be hard to figure out where in the file an error refers to. Is there any way to print file and line numbers?

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/validate_config.py
Ying-metaswitch: python-jsonschema is the debian package and needs to be put in as dependency in the installation file. The live test should preferably create a deployment using with that.

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/validate_config.py
```diff
@@ -0,0 +1,34 @@
...
+import json
+import sys
+
+schema_file = sys.argv[1]
+config_file = sys.argv[2]
```
Ying-metaswitch: Given that it's always called from existing bash script, I thought input argument checking wouldn't be needed. I would add in a check for file not existing.

sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/validate_config.py
AlexHockey: Doesn't that actually mean that one of our existing debian packages should depend on `python-jsonschema`?

Ying-metaswitch:
Line number isn't readily available, but I added the following 'schema_path' to the error message to pinpoint the error. Filename is always printed out in the "Abort uploading $filename"

```
deque([u'properties', u'routes', u'items', u'properties', u'name', u'type'])
```

AlexHockey:
That's great. Could we format the schema path before printing though. If you path is stored in the variable `schema_path`, `print " -> ".join(schema_path)` will format the text like this, which is more readable in my opinion:

> properties -> routes -> items -> properties -> name -> type

Ying-metaswitch:
Work to do: put `python-jsonschema` as a debian package for installation dependency. Ellie knows about this. You may create a deployment with this installed, and do the live tests.

PR #1872, opened 2017-07-07
[Reviewer: Ellie] Remove port from the default Sproutlet URIs

richardwhiuk:
As discussed, we think this is the right change.

### DIFF
--- a/include/sproutlet_options.h
+++ b/include/sproutlet_options.h
@@ -135,7 +135,6 @@
                            opt.prefix_##NAME_LOWER +                           \
                            "." +                                               \
                            opt.sprout_hostname +                               \
-                           ":" + std::to_string(opt.port_##NAME_LOWER) +       \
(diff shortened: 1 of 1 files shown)


pdavies:
@eleanor-merry @rkd-msw Does this need to wait until @richardwhiuk is back from holiday, or can I merge?

eleanor-merry:
I say merge it

rkd-msw:
@eleanor-merry , @richardwhiuk - we've hit test failures after taking this, which are because you need corresponding changes in Homestead. See https://github.com/Metaswitch/homestead/blob/dev/debian/homestead.init.d#L96-L98. Do you want me to raise an issue?

@graemerobertson

eleanor-merry:
I've hopefully made the change in https://github.com/Metaswitch/homestead/pull/463/files - I don't think there needs to be a separate issue.

PR #1874, opened 2017-07-12
[Reviewer: Mat] Tweak AS comm failure text

sebrexmetaswitch:
The alarm and PD log now warn that call setup time may also be increased if an
AS with session continued default handling is uncontactable.

Tested live, and verified that the correct text is seen

### DIFF
--- a/include/sprout_pd_definitions.h
+++ b/include/sprout_pd_definitions.h
@@ -420,7 +420,7 @@ static const PDLog2<const char *, const char*> CL_SPROUT_SESS_CONT_AS_COMM_FAILU
   LOG_ERR,
   "Sprout is currently unable to successfully communicate with an Application Server that uses session continued default handling. The server's URI is %s. Failure reason: %s",
   "Communication is failing to an Application server",
-  "Probable minor degradation of service, or loss of a supplemental service. The precise impact will vary depending on the role of the Application Server in the deployment.",
+  "Probable minor degradation of service, loss of a supplemental service or increased call setup time. The precise impact will vary depending on the role of the Application Server in the deployment.",
   "Investigate why communication to this Application Server is failing. It might be due to failure of the AS, misconfiguration of Initial Filter Criteria, or network / DNS problems"
 );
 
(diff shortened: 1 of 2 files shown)


MatMeredith:
Looks OK as far as it goes, but I wonder whether we need to be more explicit about the impacts?   IIRC, IFCs can specify that either the call should continue if the app server is unavailable, or should fail.   In the latter case, is this still the log / alarm we will hit?    If so, that's feels like more than a minor service impact -- depending on  the app server, this might even mean that ALL calls are failing?   When I recently updated the alarms following split storage I significantly expanded the descriptions to be more explicit.   https://github.com/Metaswitch/sprout/commit/3b10d08fd9e79149d7a9a4325059b325e08670b9   Should we do the same here?

sebrexmetaswitch:
We have a different log and alarm for an app server with session terminated default handling (i.e. the type that causes calls to fail when it's unavailable), so this one is specifically for the type that allows calls to succeed. (The other one is marked as a MAJOR alarm, since it means call failures.)

The problem with being more explicit is that we don't know the role of the app server, hence the existing placeholder text `The precise impact will vary depending on the role of Application Servers in the deployment`

MatMeredith:
OK -- that's less of an issue then and I don't care that much.   

I'm still not convinced by the description though -- it seems very vague and I'd rather that we elaborated a bit if necessary to more clearly describe the impact, or range of possible impacts?   Someone reading this alarm needs to be able to work out how serious this issue is and whether they need to "go and wake someone up to fix it".

"Probable minor degradation of service, loss of a supplemental service or increased call setup time. The precise impact will vary depending on the role of the Application Server in the deployment"

- Why "minor degradation of service"?  (and what does that mean anyway).   Are we not sure that whatever service being provided by the app server will be lost?   Or if not why not -- might there be e.g. a backup app server?

- Can we not say something more definite like "The service(s) provided by this app server will be unavailable until this alarm clears.  If the app server is not responding at all, then call setup times for all subscribers configured to use this app server will be substantially increased."

PR #1876, opened 2017-07-14
[Reviewer: Mat] Better logging for AS comm error alarms

sebrexmetaswitch:
This is a fix for https://github.com/Metaswitch/clearwater-issues/issues/2293

As per the issue, we think the behaviour here is sensible so we just want to make it more transparent. We already say that the alarm will take 5-10 minutes to clear (https://github.com/Metaswitch/sprout/blob/dev/sprout-base.root/usr/share/clearwater/infrastructure/alarms/sprout_alarms.json#L252), so this just adds a warning log whenever we check if the servers are healthy again and decide that they're not. We log the servers that aren't healthy, and when the next check will be.

We still don't clear the alarm on startup - we require at least 5 minutes to pass without errors. We could easily set the `_next_check_time` to 0 so that the very first successful communication with an AS triggers it to clear if you think we should be clearing it sooner.

Tested live and in UTs that the warning log generates the correct text:
```
14-07-2017 14:02:33.519 UTC Warning as_communication_tracker.cpp:135: Not clearing the alarm as failures detected for the  following ASs: sip:as.cwc.openstack.obu-enfield.test;transport=TCP. Next
recheck will be in 300000ms.
```

### DIFF
--- a/src/as_communication_tracker.cpp
+++ b/src/as_communication_tracker.cpp
@@ -20,6 +20,11 @@ AsCommunicationTracker::AsCommunicationTracker(Alarm* alarm,
   _as_ok_log(as_ok_log)
 {
   pthread_mutex_init(&_lock, NULL);
+
+  // Clear the alarm on startup so we don't get alarms hanging over from a
+  // previous run. If an AS is actually in error, we'll alarm as soon as we
+  // detect a failure.
+  _alarm->clear();
 }
 
 
--- a/src/as_communication_tracker.cpp
+++ b/src/as_communication_tracker.cpp
@@ -101,6 +101,9 @@ void AsCommunicationTracker::check_for_healthy_app_servers()
       std::map<std::string, int>::iterator curr_as = _as_failures.begin();
       std::map<std::string, int>::iterator next_as;
 
+      // We build this string for logging which ASs are in failure.
+      std::string failed_as_string;
+
       while (curr_as != _as_failures.end())
       {
         next_as = std::next(curr_as);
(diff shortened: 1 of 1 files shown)


MatMeredith:
Thoughts...
- I'm not sure how valuable this is -- at least in the scenario where the issue was hit...   If Sprout starts up and all app servers are healthy then the alarm will still stay stuck for 5 minutes, and there won't be any logs in the meantime explaining why.   I'm not sure that this new log helps?   
- When an App server fails or is deemed to have recovered we make an ENT log.  If there is value in the new log then should it actually be an ENT log?   I suspect there is an issue here in that testers / developers take more notice of WARNING / ERROR logs than ENT logs, whereas for a customer it will be the other way round?
- I notice that we actually only do the check for healthy app servers when we get notified that a request has succeeded or failed.   In the real world that will probably happen often enough that everything is OK, but in theory at least you could end up with an alarm stuck forever -- e.g. if a sess term app server is failed at the point that you reconfigure your IFCs to only use sess cont app servers, then the sess term app server will remain alarmed forever, even across Sprout restarts?

I think I might still be tempted to clear the alarms on Sprout restart.   If there's an app server that is unhealthy we'll raise the alarm again almost instantly.   Alarms flickering on and off isn't great, but hopefully (!) Sprout restart is rare enough for it not to be issue here?

sebrexmetaswitch:
Fair points all - I was thinking mostly in terms of the alarm getting "stuck" in the first place, not so much what happens on restart (since that was the original issue). But I think you've convinced me that we may as well clear it on restart, and it will re-alarm as soon as we have a failure, which seems sensible. So I'll make that change.

As for why this is a warning log - there is a PD log when an AS changes state, and I don't think we want to pollute the PD logs with a bunch of "we're not doing anything logs", so I think warning logs are better here (there's more info available if they want to dig deeper into why the AS hasn't changed state).

MatMeredith:
LGTM

PR #1878, opened 2017-07-18
Include contact parameters on NOTIFY bodies

AlexHockey:
(Section 5.1 of RFC3680)[https://tools.ietf.org/html/rfc3680#section-5.1] states that parameters from the contact header of a binding should be included in the `unknown-param` elements in the reg state NOTIFY. This PR makes that the case. 

Tested in UT, and also live (by @johadalin).

### DIFF
--- a/src/ut/subscription_test.cpp
+++ b/src/ut/subscription_test.cpp
@@ -112,7 +112,11 @@ public:
     b1->_params["+sip.instance"] = "\"<urn:uuid:00000000-0000-0000-0000-b4dd32817622>\"";
     b1->_params["reg-id"] = "1";
     b1->_params["+sip.ice"] = "";
+    b1->_params["+g.3gpp.smsip"] = "";
     b1->_emergency_registration = false;
+
+    // Save off the contact params for testing later.
+    _contact_params = b1->_params;
 
     // Add the AoR record to the store.
     std::string aor = "sip:6505550231@homedomain";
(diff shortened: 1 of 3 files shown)


src/ut/subscription_test.cpp
```diff
@@ -1027,6 +1042,60 @@ TEST_F(SubscriptionTest, NoDefaultID)
...
+  rapidxml::xml_node<>* reg_info = doc.first_node("reginfo");
+  rapidxml::xml_node<>* registration = reg_info->first_node("registration");
+  rapidxml::xml_node<>* contact = registration->first_node("contact");
+
+  std::map<std::string, std::string> params;
+  for (rapidxml::xml_node<>* unknown_param = contact->first_node("unknown-param");
+       unknown_param != nullptr;
+       unknown_param = unknown_param->next_sibling("unknown-param"))
+  {
+    params[unknown_param->first_attribute("name")->value()] = unknown_param->value();
```
eleanor-merry: Should you check the value of the parameters as well? I don't think you currently have an unknown parameter that has a value in the UTs as well.

AlexHockey: The UT has contact parameters that have values and some that are valueless. I check that the unknown-params on the NOTIFY exactly match those on the binding. That sounds sufficient - what am I missing?

AlexHockey: As discussed, all the parameters that were already defined in the UT are unknown parameters. I only added another (valueless) parameter as that was the one that prompted the original issue and I wanted to confirm that parameter worked.

PR #1879, opened 2017-07-19
[Reviewer: Richard] Move sproutlet timer pop handling to worker thread

sebrexmetaswitch:
This is a fix for issue https://github.com/Metaswitch/clearwater-issues/issues/2299, where the PJSIP transport thread was busy handling AS timeout timer pops and so couldn't respond to a SIP OPTIONS poll, causing monit to kill it.

PJSIP timers pop on the transport thread, but the sproutlets may need to do actual work in response to a timer pop. In the case of an AS, we either continue with the originating/terminating processing or we terminate the call (session continued vs session terminated). If an AS becomes unresponsive when Sprout is under load, we'll get loads of timers popping at basically the same time, so it's important that we don't process each one in serial on the transport thread.

The `SproutletTimerCallbackData` and timer get deleted when we delete the `UASTsx`, so we don't need to free them up when deleting the `SproutletTimerCallback`.

### Testing
* Verified `make full_test` passes
* Live tested that running load through sprout and restarting an app server triggers these timer pops, and added logs to verify that the timers pop on the main thread and the actual work is done on a worker thread.


include/sproutletproxy.h
```diff
@@ -114,6 +115,16 @@ class SproutletProxy : public BasicProxy, SproutletHelper
     void* context;
   };
 
+  // The timer callback object, which is run on a worker thread
+  class SproutletTimerCallback : public PJUtils::Callback
```
richardwhiuk: Given this is only ever used by the UASTsx, would it be neater to just put it inside the UASTsx class? That'll mean you don't need the friend declaration.

richardwhiuk: Also, given it's inside the `SproutletProxy` class, calling it a `SproutletTimerCallback` is redundant - it's full name is `SproutletProxy::SproutletTimerCallback` - surely `SproutletProxy::TimerCallback` or `SproutletProxy::UASTsx::TimerCallback` is sufficient.

src/sproutletproxy.cpp
```diff
@@ -998,7 +1009,22 @@ void SproutletProxy::UASTsx::on_timer_pop(pj_timer_heap_t* th,
...
   TRC_DEBUG("Sproutlet timer popped, id = %ld", (TimerID)tentry);
-  ((SproutletTimerCallbackData*)tentry->user_data)->uas_tsx->process_timer_pop(tentry);
+
+  SproutletTimerCallback* callback = new SproutletTimerCallback(tentry);
+#ifndef UNIT_TEST
+  if (is_pjsip_transport_thread())
+  {
+    // Timer pops happen on the main pjsip transport thread, but we want to
+    // handle them on a worker thread
+    add_callback_to_queue(callback);
```
richardwhiuk: I'd have thought this would need scoping as PJUtils::add_callback_to_queue....

src/sproutletproxy.cpp
```diff
@@ -998,7 +1009,22 @@ void SproutletProxy::UASTsx::on_timer_pop(pj_timer_heap_t* th,
                                           pj_timer_entry* tentry)
 {
   TRC_DEBUG("Sproutlet timer popped, id = %ld", (TimerID)tentry);
```
richardwhiuk: Can you move this comment into the if block, and put it so it either reads:

`TRC_DEBUG("Sproutlet timer popped, id = %ld, adding to queue", (TimerID)tentry);`

or

`TRC_DEBUG("Sproutlet timer popped, id = %ld, running immediately", (TimerID)tentry);`

Alternatively, you could just pass in the value of `is_pjsip_transport_thread()`, but I think the above is tidier, and clearer to read.

We possibly should also have a debug statement inside the callback, indicating that we are now running the callback.

src/sproutletproxy.cpp
```diff
@@ -998,7 +1009,22 @@ void SproutletProxy::UASTsx::on_timer_pop(pj_timer_heap_t* th,
...
-  ((SproutletTimerCallbackData*)tentry->user_data)->uas_tsx->process_timer_pop(tentry);
+
+  SproutletTimerCallback* callback = new SproutletTimerCallback(tentry);
+#ifndef UNIT_TEST
+  if (is_pjsip_transport_thread())
```
richardwhiuk: It's slightly irritating we repeat this if UNIT_TEST, blah blah model. Not obvious what we can do....

Maybe `add_callback_to_queue` should handle this for us....

Would probably need to rename it to `schedule_worker_thread_callback` to make it less obvious that it might not get queued.

PR #1880, opened 2017-07-19
[Reviewer: Richard] SAS log if authentication fails due to stale nonce

sebrexmetaswitch:
This is a fix for https://github.com/Metaswitch/clearwater-issues/issues/2266, and it adds a SAS log if an authentication attempt fails because the nonce is stale.

The actual SAS log is in the linked PR

Tested live that it generates the log correctly.


src/authenticationsproutlet.cpp
```diff
@@ -1157,6 +1157,12 @@ void AuthenticationSproutletTsx::on_rx_initial_request(pjsip_msg* req)
     TRC_DEBUG("No authentication information in request or stale nonce, so reject with challenge");
     pj_bool_t stale = (status == PJSIP_EAUTHACCNOTFOUND);
 
+    if (stale)
```
richardwhiuk: Can we make the debug statement three lines above specific about which case it is?

The comment also says 'no authentication vector found in the store (so request is likely stale)', so I think we need to be careful how definitive we are here.

PR #1882, opened 2017-07-21
[Reviewer: AJH] Change behaviour of S-CSCF URIs

sathiyan-sivathas:
Alex, this PR implements the changes to how the S-CSCF URIs are used. This means that the server name on the SAR matches the original route header that was on the request that entered the node. The service route header should also match the route header that was originally received.

I believe I've got the mainline tests working for the S-CSCF and Registrar, but I don't have full coverage yet. I've also disabled a test temporarily because it doesn't make sense. Please could you review.


src/authenticationsproutlet.cpp
```diff
@@ -877,6 +885,16 @@ void AuthenticationSproutletTsx::on_rx_initial_request(pjsip_msg* req)
   int sc = unauth_sc;
   status = PJ_SUCCESS;
 
+  // Construct the S-CSCF URI for this transaction. Use the configured S-CSCF
+  // URI as a starting point.
+  pjsip_sip_uri* scscf_uri = (pjsip_sip_uri*)pjsip_uri_clone(get_pool(req), stack_data.scscf_uri);
+  pj_str_t unused_local_hostname;
+  SCSCFUtils::get_scscf_uri(req,
+                            scscf_uri,
+                            &unused_local_hostname,
```
AlexHockey: If you have output parameters that the caller might not care about, it's a good idea to let the caller pass them as NULL. That way you don't have to define dummy values like you have here.

src/scscf_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+void SCSCFUtils::get_scscf_uri(pjsip_msg* req,
+                               pjsip_sip_uri* scscf_uri,
+                               pj_str_t* local_hostname,
+                               SproutletTsxHelper* tsx)
+{
+  // Get the local hostname part of the URI that routed to this Sproutlet. We
+  // will use this in the S-CSCF URI.
+  //
+  // This is so that we preserve the URI of the S-CSCF that we originally tried
+  // to route to so that we then send it on the SAR to the HSS.
```
AlexHockey: It would be useful to give some examples of the possible URI combinations and how they would get transformed.

src/scscf_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+    original_uri = (pjsip_sip_uri*)original_route->name_addr.uri;
+  }
+  else
+  {
+    original_uri = (pjsip_sip_uri*)req->line.req.uri;
+  }
+
+  pj_pool_t* pool = tsx->get_pool(req);
+  pj_str_t unused_service_name;
+  bool success = tsx->get_local_hostname(original_uri, local_hostname, &unused_service_name, pool);
```
AlexHockey: Again, allow `get_local_hostname` to take NULL pointers and you can deleted these temporary values.

src/scscf_utils.cpp
```diff
@@ -0,0 +1,67 @@
...
+      if (pj_strcmp2(&service_name, ""))
+      {
+        pj_str_t period = pj_str((char*)".");
+        pj_strdup(pool, &hostname, &service_name);
+        PJUtils::pj_str_concatenate(&hostname, &period, pool);
+        PJUtils::pj_str_concatenate(&hostname, local_hostname, pool);
+      }
+      else
+      {
+        pj_strdup(pool, &hostname, local_hostname);
```
AlexHockey: Looking at all of this I think it would have been easier to have returned the local hostname and service name as `std:string`s and only converted back to `pj_str`s after the manipulations.

src/hssconnection.cpp
```diff
@@ -701,7 +710,7 @@ HTTPCode HSSConnection::update_registration_state(const std::string& public_user
   std::string json_wildcard =
         (wildcard != "") ? ", \"wildcard_identity\": \"" + wildcard + "\"" : "";
   std::string req_body = "{\"reqtype\": \"" + type + "\"" +
-                          ", \"server_name\": \"" +_scscf_uri + "\"" +
+                          ", \"server_name\": \"" + server_name + "\"" +
```
AlexHockey: The `HSSConnection` still has an `_scscf_uri` parameter. Should this be removed?

src/impistore.cpp
```diff
@@ -135,6 +137,7 @@ ImpiStore::AuthChallenge* ImpiStore::AuthChallenge::from_json(rapidjson::Value*
       JSON_SAFE_GET_UINT_MEMBER(*json, JSON_NONCE_COUNT, auth_challenge->nonce_count);
       JSON_SAFE_GET_INT_MEMBER(*json, JSON_EXPIRES, auth_challenge->expires);
       JSON_SAFE_GET_STRING_MEMBER(*json, JSON_CORRELATOR, auth_challenge->correlator);
+      JSON_SAFE_GET_STRING_MEMBER(*json, JSON_SCSCF_URI, auth_challenge->scscf_uri);
```
AlexHockey: Do we need to worry about upgrade at all here? What if an uplevel node reads a down-level record that doesn't have the URI filled in?

src/registrarsproutlet.cpp
```diff
@@ -322,13 +326,23 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
     SAS::report_event(event);
   }
 
+  // Construct the S-CSCF URI for this transaction. Use the configured S-CSCF
+  // URI as a starting point.
+  pjsip_sip_uri* scscf_uri = (pjsip_sip_uri*)pjsip_uri_clone(get_pool(req), stack_data.scscf_uri);
+  SCSCFUtils::get_scscf_uri(req,
+                            scscf_uri,
+                            &_local_hostname,
```
AlexHockey: It took me ages to realize that this was what set up the `_local_hostname`. The fact that calling `get_scscf_uri` happens to store off the local hostname for just his sproutlet seems like a massive side-effect and suggests the interface to `get_scscf_uri` isn't very clean.

src/registrarsproutlet.cpp
```diff
@@ -722,6 +737,29 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
+      PJUtils::pj_str_concatenate(&hostname, &period, get_pool(rsp));
+      PJUtils::pj_str_concatenate(&hostname, &_local_hostname, get_pool(rsp));
+    }
+    else
+    {
+      pj_strdup(get_pool(rsp), &hostname, &_local_hostname);
+    }
+
+    sr_uri->host = hostname;
+  }
```
AlexHockey: This is so similar to some of the processing in `get_scscf_uri` that it suggests some of that function should be further subfunctioned, and that smaller subfunction be used here.

src/sproutletproxy.cpp
```diff
@@ -390,9 +390,19 @@ Sproutlet* SproutletProxy::match_sproutlet_from_uri(const pjsip_uri* uri,
...
 {
-  pjsip_sip_uri* base_uri = (pjsip_sip_uri*)(route ? route->name_addr.uri : nullptr);
+  pjsip_sip_uri* base_uri = NULL;
+  if (route != NULL)
+  {
+    base_uri = (pjsip_sip_uri*)route->name_addr.uri;
+  }
+  else
+  {
+    base_uri = (pjsip_sip_uri*)req->line.req.uri;
```
AlexHockey: There is no need to pass in the request and a route header. It looks like you could pass in just the request.

src/sproutletproxy.cpp
```diff
@@ -430,13 +440,28 @@ pjsip_sip_uri* SproutletProxy::create_sproutlet_uri(pj_pool_t* pool,
...
 {
   TRC_DEBUG("Creating URI for service %s", name.c_str());
 
   pjsip_sip_uri* base_uri = ((existing_uri != nullptr) ? existing_uri : _root_uri);
   pjsip_sip_uri* uri = (pjsip_sip_uri*)pjsip_uri_clone(pool, base_uri);
-  pj_strdup(pool, &uri->host, &_root_uri->host);
+
+  // Replace the hostname part of the base URI with the local hostname part of
+  // the URI that routed to us. If this doesn't work, then fall back to using
+  // the root URI.
```
AlexHockey: Again, some examples would be useful.

src/sproutletproxy.cpp
```diff
@@ -496,6 +521,41 @@ bool SproutletProxy::is_uri_local(const pjsip_uri* uri)
   //LCOV_EXCL_STOP
 }
 
+bool SproutletProxy::get_local_hostname(const pjsip_sip_uri* uri,
+                                        pj_str_t* hostname,
+                                        pj_str_t* service_name,
+                                        pj_pool_t* pool)
+{
+  pj_strdup(pool, hostname, &uri->host);
+  pj_strdup(pool, service_name, &uri->host);
```
AlexHockey: If hostname is local, surely it's not right to fill in the service name as the hostname??

src/sproutletproxy.cpp
```diff
@@ -496,6 +521,41 @@ bool SproutletProxy::is_uri_local(const pjsip_uri* uri)
...
+      // Set the service name.
+      service_name->slen = sep - hostname->ptr;
+
+      // Remove the service name part and the period from the hostname.
+      hostname->slen -= (sep - hostname->ptr + 1);
+      hostname->ptr = sep + 1;
+
+      if (!is_host_local(hostname))
+      {
+        pj_strdup2(pool, service_name, "");
```
AlexHockey: Why only clear out the service name - surely the `hostname` is invalid as well if we hit this code.

src/sproutletproxy.cpp
```diff
@@ -496,6 +521,41 @@ bool SproutletProxy::is_uri_local(const pjsip_uri* uri)
...
+    }
+    else
+    {
+      pj_strdup2(pool, service_name, "");
+      return false;
+    }
+  }
+
+  return true;
+}
```
AlexHockey: I don wonder if there's a risk in having this implementation be different from the code in the sproutlet selection algorithm. Should the selection code be using this function?

include/sproutlet.h
```diff
@@ -252,11 +252,28 @@ class SproutletTsxHelper
...
+
+  /// Get the local hostname and (possible) service name part of a SIP URI.
+  ///
+  /// @returns            - Whether this operation was successful.
+  ///
+  /// @param uri          - The SIP URI.
+  /// @param hostname     - Return parameter that contains the local hostname
+  ///                       part of the URI.
+  /// @param service_name - Return parameter that contains the service name
+  ///                       part of the URI (optional).
```
AlexHockey: It's not clear to me whether a URI like `sip:local;service=scscf` would return a service name of "scscf".

src/sproutletproxy.cpp
AlexHockey: Yeah, assume that `local` and `scscf.local` were both considered local domains for some reason. When presented with the URI `sip:scscf.local` this function says the hostname is "scscf.local" but the service name is "". Whereas I think the sproutlet selection algorithm would say that the hostname is "local" and the service name is "scscf". It's really crucial that these two bits of code agree, which makes me think they should be the same bit of code.

src/subscriber_data_manager.cpp
```diff
@@ -1398,6 +1400,10 @@ SubscriberDataManager::AoR* SubscriberDataManager::JsonSerializerDeserializer::
          ((doc.HasMember(JSON_TIMER_ID)) && ((doc[JSON_TIMER_ID]).IsString()) ?
                                              (doc[JSON_TIMER_ID].GetString()) :
                                               "");
+    aor->_scscf_uri =
+         ((doc.HasMember(JSON_SCSCF_URI)) && ((doc[JSON_SCSCF_URI]).IsString()) ?
+                                                (doc[JSON_SCSCF_URI].GetString()) :
+                                                 "");
```
AlexHockey: `JSON_SAFE_GET_STRING`?

src/ut/handlers_test.cpp
```diff
@@ -1689,7 +1691,7 @@ TEST_F(DeleteImpuTaskTest, WritingToRemoteStores)
       EXPECT_CALL(*store, set_aor_data(impu, _, EmptyAoR(), _, _))
         .WillOnce(DoAll(SetArgReferee<4>(true), // All bindings expired
                         Return(Store::OK)));
-      EXPECT_CALL(*mock_hss, update_registration_state(impu, _,_, _, _, _))
+      EXPECT_CALL(*mock_hss, update_registration_state(impu, _, _, _, _, _, _))
```
AlexHockey: I was expecting at least some of the tests in this file to check the correct SCSCF name is being passed to the HSS?

src/ut/hssconnection_test.cpp
```diff
@@ -1160,7 +1162,7 @@ TEST_F(HssWithSifcTest, ComplexSifcIfcMix)
   // 6 distinct iFCs with priorities 1, 1, 2, 3, 3 and 4 should be returned.
   // Additionally 3 iFCs from shared iFCs with priorities 1, 2 and 2 should be
   // returned.
-  _sifc_hss.update_registration_state("sifcifcmix", "", HSSConnection::REG, regstate, ifcs_map, uris, 0);
+  _sifc_hss.update_registration_state("sifcifcmix", "", HSSConnection::REG, regstate, "server_name", ifcs_map, uris, 0);
```
AlexHockey: You need a test to check that you can specify a different server name and have that be passed through on the HTTP request.

src/ut/mock_hss_connection.h
```diff
@@ -30,39 +30,43 @@ class MockHSSConnection : public HSSConnection
                                       &SNMP::FAKE_EVENT_ACCUMULATOR_TABLE,
                                       &SNMP::FAKE_EVENT_ACCUMULATOR_TABLE,
                                       NULL,
-                                      "sip:scscf.sprout.homedomain:5058;transport=TCP",
+                                      "sip:scscf.homedomain;transport=TCP",
```
AlexHockey: We need to delete this URI right?

src/ut/registrar_test.cpp
```diff
@@ -113,7 +113,7 @@ string Message::get()
                "Allow: PRACK, INVITE, ACK, BYE, CANCEL, UPDATE, SUBSCRIBE, NOTIFY, REFER, MESSAGE, OPTIONS\r\n"
                "%9$s"
                "%7$s"
-               "Route: <sip:%12$s;transport=tcp;lr>\r\n"
+               "Route: <sip:%12$s;transport=tcp;lr;service=registrar>\r\n"
```
AlexHockey: Just checking, this presumably matches the URI that would be added by the authentication module, correct?

src/ut/registrar_test.cpp
```diff
@@ -155,7 +155,7 @@ class RegistrarTest : public SipTest
   static void SetUpTestCase()
   {
     SipTest::SetUpTestCase();
-    SipTest::SetScscfUri("sip:all.the.sprout.nodes:5058;transport=TCP");
+    SipTest::SetScscfUri("sip:scscf.sprout.homedomain:5058;transport=TCP");//"sip:scscf.sprout.homedomain:5058;transport=TCP");
```
AlexHockey: Why the trailing comment.

src/ut/registrar_test.cpp
```diff
@@ -240,10 +240,14 @@ class RegistrarTest : public SipTest
...
 
+    std::unordered_set<std::string> additional_home_domains;
+    additional_home_domains.insert("sprout.homedomain");
+    additional_home_domains.insert("sprout-site2.homedomain");
+
     _registrar_proxy = new SproutletProxy(stack_data.endpt,
                                           PJSIP_MOD_PRIORITY_UA_PROXY_LAYER,
                                           "homedomain",
-                                          std::unordered_set<std::string>(),
+                                          additional_home_domains,
```
AlexHockey: FYI you might be able to just pass `{"foo", "bar"}` in instead of constructing the set in advance.

src/ut/registrar_test.cpp
```diff
@@ -970,7 +1026,7 @@ TEST_F(RegistrarObservedHssTest, MultipleRegistrations)
 
 // Check that a sensible result (no bindings) is returned on a "fetch bindings"
 // REGISTER if the subscriber isn't actually registered
-TEST_F(RegistrarTest, FetchBindingsUnregistered)
+TEST_F(RegistrarTest, DISABLED_FetchBindingsUnregistered)
```
AlexHockey: If the test is useless let's just delete it. But why do you think it's useless?

src/ut/registrar_test.cpp
```diff
@@ -826,7 +830,59 @@ TEST_F(RegistrarTest, SimpleMainlineAuthHeaderWithTelURI)
...
+  out = pop_txdata()->msg;
+  EXPECT_EQ(200, out->line.status.code);
+  EXPECT_EQ("OK", str_pj(out->line.status.reason));
+  EXPECT_EQ("Supported: outbound", get_headers(out, "Supported"));
+  EXPECT_EQ("Contact: <sip:f5cc3de4334589d89c661a7acf228ed7@10.114.61.213:5061;transport=tcp;ob>;expires=300;+sip.ice;+sip.instance=\"<urn:uuid:00000000-0000-0000-0000-b665231f1213>\";reg-id=1;pub-gruu=\"sip:6505550231@homedomain;gr=urn:uuid:00000000-0000-0000-0000-b665231f1213\"",
+            get_headers(out, "Contact"));
+  EXPECT_EQ("Require: outbound", get_headers(out, "Require")); // because we have path
+  EXPECT_EQ(msg._path, get_headers(out, "Path"));
+  EXPECT_EQ("P-Associated-URI: <sip:6505550231@homedomain>", get_headers(out, "P-Associated-URI"));
+  EXPECT_EQ("Service-Route: <sip:scscf.sprout-site2.homedomain:5058;transport=TCP;lr;orig>", get_headers(out, "Service-Route"));
```
AlexHockey: This test is too noisy TBH. There are only about three things this test cares about:

* That the route URI is for site 2. 
* That the service route says site 2.
* That the S-CSCF URI sent to the HSS says site 2. 

You should cut out all the other extraneous checks so you're only checking what you actually care about.

src/ut/scscf_test.cpp
```diff
@@ -209,6 +210,7 @@ namespace SP
       _first_hop(false),
       _via("10.83.18.38:36530"),
       _branch(""),
+      _route("Route: <sip:scscf.sprout.homedomain>"),
```
AlexHockey: You've changed the registrar test so the route URI looks more realistic - but you've not done the same for the S-CSCF. Why not?

src/ut/scscf_test.cpp
```diff
@@ -1385,6 +1402,27 @@ TEST_F(SCSCFTest, TestSimpleMainline)
...
+  // Make sure that the HTTP request sent to homestead contains the correct S-CSCF URI.
+  EXPECT_TRUE(_hss_connection->url_was_requested("/impu/sip%3A6505551234%40homedomain/reg-data", "{\"reqtype\": \"call\", \"server_name\": \"sip:scscf.sprout-site2.homedomain:5058;transport=TCP\"}"));
+
+  // This is a terminating call so should not result in a session setup time
+  // getting tracked.
+  EXPECT_EQ(0, ((SNMP::FakeEventAccumulatorTable*)_scscf_sproutlet->_audio_session_setup_time_tbl)->_count);
+  EXPECT_EQ(0, ((SNMP::FakeEventAccumulatorTable*)_scscf_sproutlet->_video_session_setup_time_tbl)->_count);
+
+  // It also shouldn't result in any forked INVITEs
+  EXPECT_EQ(0, ((SNMP::FakeCounterTable*)_scscf_sproutlet->_forked_invite_tbl)->_count);
```
AlexHockey: Can this test also check that the RR is for site 1?

include/subscriber_data_manager.h
```diff
@@ -245,6 +245,10 @@ class SubscriberDataManager
     // Chronos Timer ID
     std::string _timer_id;
 
+    /// S-CSCF URI name for this AoR. This is used on the SAR if the
+    /// registration expires.
+    std::string _scscf_uri;
```
AlexHockey: Am I right in thinking that this should never change? Is it worth saying something like that in the comment? In fact, is it worth policing this in some way (even if it's just as an engineering log or something)?

include/impistore.h
```diff
@@ -100,6 +101,10 @@ class ImpiStore
     /// Correlator between original challenge and responses.
     std::string correlator;
 
+    /// URI of the SCSCF that issued the challenge. This is the server name
+    /// used on the SAR if the authentication times out.
+    std::string scscf_uri;
```
AlexHockey: Again, can this ever change after the challenge has been created? I think it probably can but that's just an artifact of our implementation. Worth a comment?

include/sproutletproxy.h
```diff
@@ -52,7 +52,16 @@ class SproutletProxy : public BasicProxy, SproutletHelper
...
-                              pj_pool_t* pool) const;
+                              const pjsip_msg* req,
+                              pj_pool_t* pool);
+
+  enum SPROUTLET_SELECTION_TYPES
+  {
+    SERVICE_NAME=0,
+    DOMAIN_PART,
+    USER_PART,
+    UNKNOWN,
```
AlexHockey: Why do we need an unknown value?

src/sproutletproxy.cpp
```diff
@@ -496,6 +492,32 @@ bool SproutletProxy::is_uri_local(const pjsip_uri* uri)
...
+                                  unused_selection_type);
+
+  if (!local_hostname.empty())
+  {
+    pj_strdup2(pool, hostname, local_hostname.c_str());
+  }
+  else
+  {
+    // This code is only executed if:
+    //  - We receive a request that doesn't match a Sproutlet, or
```
AlexHockey: It's not just receiving requests right? It's about the URI that's passed in. You're assuming that the URI passed to this function will route back to the sproutlet. That is why it's correct to default the hostname to the URI's host if `match_sproutlet_from_uri` returns NULL.

include/sproutletproxy.h
AlexHockey: Ah, I see why. It's not nice but it's probably OK. I would use a large numeric value though - this gives us space to add new "valid" values but doesn't leave us with unknown being in the middle of them. Also I think `NONE_SELECTED` would be a better name than `UNKNOWN` (it's not that we don't know why a sproutlet was selected, it's that a sproutlet hasn't even *been* selected).

PR #1883, opened 2017-07-24
[Reviewer: Seb] Ignore libc double free during test exit

richardwhiuk:
Seb,

This adds a valgrind suppression (in the same fashion as we have for Homestead) for a valgrind warning we've seen on certain machines.

We don't have the exact scenario for this, but I'm happy that it's fairly safe to suppress it because:

- We only see it during UT
- It always happens when the UTs are exiting rather than during execution

Short term plan is to suppress it and keep investigating.


sebrexmetaswitch:
Just noticed you have raised the issue, and that you've written quite a lot in it so ignore the above :)

PR #1890, opened 2017-08-02
Rework sprout UTs

KristaHyer:
As part of my work reworking the sprout UTs, I've created a new common testing script.
I've added functionality to this script, so that it can create a string which shows the XML of an IMS subscription.

I replaced only some of the hardcoded XML in scscf_test.cpp
I will replace the rest if you approve these changes.

Make full_test passes with these changes.


src/ut/testingcommon.h
```diff
@@ -0,0 +1,234 @@
...
+       ++it)
+  {
+    if (it->first == identity)
+    {
+      it->second = barring_indication;
+    }
+  }
+}
+
+// Add the text for an iFC to the list of iFCs.
```
KristaHyer: I'm thinking of editing this function.
Currently this creates the text for an iFC, and adds that to a list.
I'm thinking of making this more like "addIdentity", so it will add the details of an iFC to a list, and then the text will be built up later.
Let me know what you think.

AlexHockey:
This looks a lot nicer and easier to work with. Nice one! I did have a few comments/suggestions about how to make this *even* better. These are below, in order of highest priority to lowest. I'm not saying we have to all of these - after all we're making ergonomic changes so it's sometimes hard to say whcich of two options is easier to use. 

Note that I've not done a full review at this stage - I will do that when we decide to stop making improvements. 

### Code Structure

Lastly, you have put everything in a header file. you should be distinguishing between the interface to your new classes (which go in the header) and the implementation (which should go in a cpp file. 

### ServiceProfileBuilder

It's common to have builder methods return themselves by reference as this means that instead of:

```cpp
ServiceProfileBuilder service_profile_1 = ServiceProfileBuilder();
service_profile_1.addIdentity("sip:6505551000@homedomain");
service_profile_1.addIfc(2, {"<Method>INVITE</Method>"}, "sip:4.2.3.4:56788;transport=UDP");
service_profile_1.addIfc(1, {"<Method>INVITE</Method>"}, "sip:1.2.3.4:56789;transport=UDP");
```

you can write the following which is more concise, easier to follow, and less bug prone (you're less likely to accidentally add a field to the object you were working on previously)

```cpp
ServiceProfileBuilder service_profile_1
  .addIdentity("sip:6505551000@homedomain");
  .addIfc(2, {"<Method>INVITE</Method>"}, "sip:4.2.3.4:56788;transport=UDP");
  .addIfc(1, {"<Method>INVITE</Method>"}, "sip:1.2.3.4:56789;transport=UDP");
```

### SubscriptionBuilder

I don't see why `SubscriptionBuilder` needs to be a method. Couldn't it just be a function?

### SPT Builders?

It might be worth having `SPT` helpers. For example, instead of writing:

```cpp
service_profile.addIfc(1, {"<Method>INVITE</Method>", "<SessionCase>1</SessionCase><!-- terminating-registered -->"}, "sip:1.2.3.4:56789;transport=UDP", 1);
```

it would be nice to add 

```cpp
service_profile.addIfc(1, {SPT::method("INVITE"), SPT::sess_case(TERM)}, "sip:1.2.3.4:56789;transport=UDP", 1);
```

or maybe even:

```cpp
service_profile.addIfc(1, SPTs().with_method("INVITE").with_sess_case(TERM), "sip:1.2.3.4:56789;transport=UDP", 1);
```

KristaHyer:
I applied most of the changes you suggested, except:
- I didn't change SubscriptionBuilder to a function, to save time.
-I changed "addIfc", which meant your suggestion for an improvement no longer applied.

I went through and changed a lot of the hardcoded XML in scscf_test.cpp.
Unfortunately, there's so much that I didn't get through it!

But this should be enough to review, then I'll change the rest later.

KristaHyer:
I've commonised the Message classes from the following files:
- scscf_test.cpp
- bono_test.cpp
 -icscfsproutlet_test.cpp
- basicproxy_test.cpp
- common_sip_processing_test.cpp

The Message class was also present in several other files, but had diverged too much to be worth combining. These files are:
- registrar_test.cpp
- session_expires_helper_test.cpp
- options_test.cpp
- sproutletproxy_test.cpp
- mangelwurzel_test

src/ut/scscf_test.cpp
```diff
@@ -45,307 +46,6 @@ using testing::NiceMock;
...
-      _from("6505551000"),
-      _fromdomain("homedomain"),
-      _to("6505551234"),
-      _todomain("homedomain"),
-      _content_type("application/sdp"),
-      _forwards(68),
-      _first_hop(false),
-      _via("10.83.18.38:36530"),
-      _branch(""),
-      _route("Route: <sip:sprout.homedomain;service=scscf>"),
```
KristaHyer: Setting this broke many tests in icscfsproutlet_test.cpp, basicproxy_test.cpp, and common_sip_processing_test.cpp.
I removed it and all the tests, including all the tests in this file, all passed.
I don't believe it's necessary, but thought it was worth flagging to you.

src/ut/scscf_test.cpp
```diff
@@ -45,307 +46,6 @@ using testing::NiceMock;
...
-      if (!_todomain.empty())
-      {
-        target.append("@").append(_todomain);
-      }
-
-      // If there's no route, the target goes in the request
-      // URI. Otherwise it goes in the Route:, and the route goes in the
-      // request URI.
-      //string requri = _route.empty() ? target : _route;
-      //string route = _route.empty() ? "" : string("Route: ").append(target).append("\r\n");
```
KristaHyer: This comment describes the old, commented out code, so I've removed it.

src/ut/testingcommon.cpp
```diff
@@ -0,0 +1,431 @@
...
+  _unique = unique;
+  unique += 10; // leave room for manual increments
+}
+
+Message::~Message()
+{
+}
+
+// Set the route field on a message.
+void Message::set_route(pjsip_msg* msg)
```
sathiyan-sivathas: This isn't your code, but can we rename this function? What this function is really doing is turning all the record-route headers into route headers for in-dialog requests. I think something like `convert_routeset` would be better.

src/ut/testingcommon.h
```diff
@@ -0,0 +1,208 @@
...
+
+    void set_route(pjsip_msg*);
+    std::string get_request();
+    std::string get_response();
+  };
+
+}
+
+#endif
+
```
sathiyan-sivathas: Unnecessary blank line.

src/ut/testingcommon.cpp
```diff
@@ -0,0 +1,431 @@
...
+                     );
+
+  EXPECT_LT(n, (int)sizeof(buf));
+
+  std::string ret(buf, n);
+  // cout << ret <<endl;
+  return ret;
+}
+
+
```
sathiyan-sivathas: Unnecessary blank lines.

src/ut/testingcommon.cpp
```diff
@@ -0,0 +1,431 @@
...
+                   /* 12 */ _first_hop ? "" : "Via: SIP/2.0/TCP 10.114.61.213:5061;received=23.20.193.43;branch=z9hG4bK+7f6b263a983ef39b0bbda2135ee454871+sip+1+a64de9f6\r\n",
+                   /* 13 */ _cseq,
+                   /* 14 */ _via.c_str(),
+                   /* 15 */ branch.c_str()
+                     );
+
+  EXPECT_LT(n, (int)sizeof(buf));
+
+  std::string ret(buf, n);
+  // cout << ret <<endl;
```
sathiyan-sivathas: And this.

src/ut/testingcommon.cpp
```diff
@@ -0,0 +1,431 @@
...
+                   /* 14 */ route.c_str(),
+                   /* 15 */ _cseq,
+                   /* 16 */ branch.c_str(),
+                   /* 17 */ (_in_dialog) ? ";tag=10.114.61.213+1+8c8b232a+5fb751cf" : ""
+                     );
+
+  EXPECT_LT(n, (int)sizeof(buf));
+
+  std::string ret(buf, n);
+  // cout << ret <<endl;
```
sathiyan-sivathas: You can delete this now.

src/ut/scscf_test.cpp
sathiyan-sivathas: This would have been because you have `service=scscf` which doesn't match a sproutlet in the I-CSCF tests. The tests might all pass, but I think it's more realistic if the request has a route header and the sproutlet being invoked is identified in the service parameter. I think in reality sproutlets are almost always invoked using a route header, so I think the tests should match this behaviour.

KristaHyer: We have spoken about this.
The plan is to create subclasses for each test file, which will each set the appropriate _route value.
Since this is not a minor markup I will fix it in a new review.

PR #1891, opened 2017-08-03
[Reviewer: HJS] 'Allowed host state' interface change for DNS resolvers

GAB-MS:
To go along with cpp-common and cpp-common-test PRs, which I will be creating shortly.

**Please note that I have not run a `make full_test` against this to check for memory errors, due to an issue that means it takes hours to run on my dev VM. I am currently leaving one running in the background.**

I won't go into detail on every change, as this was mostly by-the-design. However, I will list interesting things to note, and interesting deviations from the design (where we've discussed these previously - let me know if you see anything unexpected!)

* `basicproxy.h` needs to include `sproutlet.h` for `ForkErrorState` - as a result, the C imports were no longer needed, as they also appear in `sproutlet.h`
* `sproutlet.h` needs to include `baseresolver.h` for the `BaseResolver::ALL_LISTS` constant - it no longer needs to include `sas.h` as a result, as it gets this from baseresolver
* The `fork_err_to_str` function had to be declared as inline, because it is a full function defined in a header file. If this is wrong, we likely need to move this function elsewhere - but I'm not sure where would be appropriate
* The new `SendRequest` struct had to be defined in `SproutletProxy` instead of `SproutletWrapper`, because it is used in the former and forward-declaration was not possible in this case
* I'm not sure about my use of the shorthand 'if' statement at line 1742 in `basicproxy.cpp`. I think it reads clear enough and saves a decent chunk of space over a full if/else block, shout if you disagree
* On line 1954 in `basicproxy.cpp`, fork_error has to be initialised to a default value because (as far as the compiler is aware) it is not guaranteed to be set in the if/else if block below it. The function should never be called outside of those two cases, though.
* For the host state implementation in `SIPResolver::resolve`: instead of performing a bitwise check on the allowed host states every time we use them (and including appropriate logging), I found that it was much cleaner to add a function to the `BaseResolver` class (which will be in the cpp-common PR) that takes two bool references and the allowed host state bit, and sets the bools as appropriate to indicate which states are allowed.
* I added the above bools to the SAS event when 'resolving' IP addresses, with the idea being that the SAS event will vary its text to indicate that whitelisted/blacklisted addresses are not allowed, and whether the IP address was rejected for this reason.
* On line 1570 in `sproutletproxy.cpp` - I'm not too sure whether my formatting here is appropriate. I did this to avoid having to separately initialise the `SendRequest` struct and then feed it in, which seemed unnecessary
* On line 1913 in `sproutletproxy.cpp`, the if/else if block was no longer necessary, as we cna just directly set the `fork_error` to whatever's passed in.
* In `scscf_test.cpp`, the substrings that it expects for some failures had to be changed, as those failures now raise "no addresses" errors instead of the "transport error" that was previously raised when we had no addresses.
* I only added one unit test here - to check the `allowed_host_state` behaviour for IP addresses. More unit tests are added for the cpp-common changes.


include/sproutlet.h
```diff
@@ -35,7 +35,30 @@ class SproutletProxy;
...
+    case ForkErrorState::TRANSPORT_ERROR:
+      err_str = "TRANSPORT_ERROR";
+      break;
+    case ForkErrorState::NO_ADDRESSES:
+      err_str = "NO_ADDRESSES";
+      break;
+  }
+
+  return err_str;
+};
```
huws: This lot could be reduced to something like shown below.  This also returns `char*` pointers rather than copies of `string` objects, which makes sense given that the only invocations of this method immediately call `c_str()` on the returned `string` object.

```cpp
const char* const FORK_ERROR_STATE_VALUES[] = {"NONE", "TIMER", "TRANSPORT_ERROR", "NO_ADDRESS"};
inline const char* fork_error_to_str(const ForkErrorState fork_error)
  {  return FORK_ERROR_STATE_VALUES[fork_error]; }
```

include/sproutlet.h
```diff
@@ -152,7 +175,7 @@ class SproutletTsxHelper
   /// @param  req          - The request message to use for forwarding.  If NULL
   ///                        the original request message is used.
   ///
-  virtual int send_request(pjsip_msg*& req) = 0;
+  virtual int send_request(pjsip_msg*& req, int allowed_host_state) = 0;
```
huws: Missing `@param` documentation for the `allowed_host_state` parameter.

include/sproutlet.h
```diff
@@ -474,8 +497,9 @@ class SproutletTsx
   /// @returns             - The ID of this forwarded request
   /// @param  req          - The request message to use for forwarding.
   ///
-  int send_request(pjsip_msg*& req)
-    {return _helper->send_request(req);}
+  int send_request(pjsip_msg*& req,
+                   int allowed_host_state=BaseResolver::ALL_LISTS)
+    {return _helper->send_request(req, allowed_host_state);}
```
huws: Again - missing `@param` documentation for the allowed_host_state parameter.

include/sproutletproxy.h
```diff
@@ -63,6 +63,14 @@ class SproutletProxy : public BasicProxy, SproutletHelper
...
 
 protected:
+  // A struct to wrap tx_data and allowed_host_state in a convenient bundle
+  // to pass over interfaces when sending a request.
+  typedef struct
+  {
+    pjsip_tx_data* tx_data;
+    int allowed_host_state;
+  } SendRequest;
+
```
huws: I'd prefer it if this `struct` was defined next to the other `struct` at the same scope (`SproutletTimerCallbackData`), rather than mixed in with the methods.

src/basicproxy.cpp
```diff
@@ -1736,7 +1738,12 @@ void BasicProxy::UACTsx::send_request()
       // Remove the top Via from the request before reporting the error in
       // case the request is used to build an error response.
       PJUtils::remove_top_via(_tdata);
-      _uas_tsx->on_client_not_responding(this, PJSIP_EVENT_TRANSPORT_ERROR);
+
+      ForkErrorState fork_error = (_servers.size() == 0)
+                                  ? ForkErrorState::NO_ADDRESSES
+                                  : ForkErrorState::TRANSPORT_ERROR;
```
huws: I would prefer it if the `?` and `:` characters appeared at the end of the previous lines, to make it more obvious that the statement continues onto the next line.

src/sipresolver.cpp
```diff
@@ -76,18 +77,40 @@ void SIPResolver::resolve(const std::string& name,
     // The name is already an IP address, so no DNS resolution is possible.
     // Use specified transport and port or defaults if not specified.
     TRC_DEBUG("Target is an IP address - default port/transport if required");
+
+    // Check which host states are permitted.
+    bool whitelisted_allowed, blacklisted_allowed;
```
huws: Don't do this - each variable gets its own line.

src/sipresolver.cpp
```diff
@@ -76,18 +77,40 @@ void SIPResolver::resolve(const std::string& name,
...
+    get_allowed_host_states(
+                allowed_host_state, whitelisted_allowed, blacklisted_allowed);
```
huws: I think I would prefer that you just did the bit-wise logic inline.  I think it is clearer, and probably no more verbose than what you've got here.  To be clear, I'm saying keep the `bool`s, just directly give set them to the value of the correct bit-wise expression.

GAB-MS: So the motivating factor for the helper function was that we set these `bool`s in three different places. This wouldn't be a problem if we were just setting the bool values inline when we declare them, but the duplication is more of a problem when we're also throwing in debug-level trace.

I think having that trace is necessary. I could reduce the total number of `TRC_DEBUG` lines from four to two by just including the bool in the trace line, but we're still duplicating those trace lines in three places. Thoughts?

GAB-MS:
Besides the one that I've responded to above, I'll go ahead and address the rest of your comments - I think they're all reasonable changes.

On my build system - it happens on a completely fresh Sprout checkout. It's not restricted to my VM either - AJH is also affected. It only applies when running the tests under Valgrind, and it's only two specific groups of tests that are hitting issues, but those groups are taking over an hour. I've no idea what to do to diagnose this or try to fix it.

GAB-MS:
Passing this PR, along with the others, to @AlexHockey for external review.

include/sproutlet.h
```diff
@@ -151,8 +155,10 @@ class SproutletTsxHelper
   ///
   /// @param  req          - The request message to use for forwarding.  If NULL
   ///                        the original request message is used.
+  /// @param  allowed_host_state - Permitted state of hosts when resolving
+  ///                        addresses. Values are defined in BaseResolver.
```
sathiyan-sivathas: Can you fix the alignment in this comment and the one below.

src/sipresolver.cpp
```diff
@@ -76,18 +77,39 @@ void SIPResolver::resolve(const std::string& name,
...
+    const bool blacklisted_allowed = allowed_host_state & BaseResolver::BLACKLISTED;
+
     ai.transport = (transport != -1) ? transport : IPPROTO_UDP;
     ai.port = (port != 0) ? port : 5060;
-    targets.push_back(ai);
+
+    bool addr_blacklisted = blacklisted(ai);
+    bool addr_rejected = false;
+
+    if ( (!addr_blacklisted && whitelisted_allowed ) ||
```
sathiyan-sivathas: Some unnecessary spaces in this if statement.

include/basicproxy.h
```diff
@@ -13,23 +13,16 @@
...
-#include <pjlib.h>
-#include <stdint.h>
-}
-
 #include <utility>
 #include <vector>
 #include <list>
 #include <set>
 
+#include "sproutlet.h"
```
sathiyan-sivathas: I don't think it's right that you've had to include sproutlet.h here just to get the definition of `ForkErrorState` - the basic proxy shouldn't need to know about the Sproutlet interface. I don't have any solutions though. @AlexHockey might though?

AlexHockey: Agreed, especially as the `SproutletProxy` is a subclass of the `BasicProxy`. Unfortunately I think the "right" solution is to move `ForkErrorState` into a common header file that both sproutlet.h and basicproxy.h include. :-(

include/sproutlet.h
GAB-MS: Is it okay to have these comments out of alignment compared to the rest of the file's comments? The `allowed_host_state` parameter name is too long to have them all nicely in one line.

sathiyan-sivathas: What we normally do is start the comment on the next line e.g. https://github.com/Metaswitch/sprout/blob/4d9e5be63a1983988abcaad85ba90d0ecefc5744/include/subscriber_data_manager.h#L498-L499

GAB-MS: Okay, that makes sense and looks neater. I'll do that.

include/basicproxy.h
GAB-MS: I've moved `ForkErrorState` to a separate header file. @sathiyan-sivathas, could you please take another look and make sure you're happy with the split? Thanks

sathiyan-sivathas: Yep, looks good.

PR #1892, opened 2017-08-03
[Reviewer: Ellie] Fix contact header parameters in 3rdparty registrations

Neil-Bruce:
When the SCSCF URI contains a transport parameter this is located outside the SIP URI in 3rd-party registrations.  e.g.

Contact: <sip:scscf.sprout.internal:5054>;transport=TCP

Problem is caused by passing a bare URI without <> into _endpt_create_request which has an interface which states that parameters like that will be treated as other parameters rather than URI parameters.

The fix is just to add <> roudn the string we pass into this call for the contact header.


Neil-Bruce:
As well as the UT enhancements I've run the Clearwater Live tests over this fix and they pass without problem.

PR #1895, opened 2017-08-08
[Reviewer: Ying] Improve the icscfsproutlet coverage to 100%

eleanor-merry:
This PR takes the icscfsproutlet code to 100% code coverage, and removes it from the coverage exclusion list.

I've added tests for wildcards and S-CSCF selection failure when retrying a non-register request.
I've removed the on_cancel method when processing REGISTER requests, as this won't ever be hit.
I've excluded from coverage branches that are only hit when pjsip is unable to construct a URI, or methods to deal with in dialog requests.

Tested by running make full_test


eleanor-merry:
The I-CSCF *should* never receive in-dialog requests, but we've seen cases where it has so we need to have code to cope with it.
The I-CSCF *will* never receive CANCEL requests, so we don't need code for it.

PR #1896, opened 2017-08-14
Refactor timer interface

MatMeredith:


include/handlers.h
```diff
@@ -77,7 +82,35 @@ class AoRTimeoutTask : public HttpStackUtils::Task
 
 protected:
   const Config* _cfg;
-  std::string _aor_id;
+};
+
+/// Base AuthTimeoutTask class for tasks that implement authentication timeout
+/// callbacks from specific timer services.
+class AuthTimeoutTask : public HttpStackUtils::Task
```
richardwhiuk: Curious why you've moved this up the file, for now apparent reason - makes it harder to review.

include/chronoshandlers.h
```diff
@@ -0,0 +1,50 @@
+/**
+ * @file chronoshandlers.h
```
richardwhiuk: chronostasks surely - there are no handlers in this file?

src/ut/handlers_test.h
```diff
@@ -0,0 +1,148 @@
...
+using namespace std;
+using ::testing::_;
+using ::testing::Return;
+using ::testing::SetArgReferee;
+using ::testing::SetArgPointee;
+using ::testing::SaveArg;
+using ::testing::SaveArgPointee;
+using ::testing::InSequence;
+using ::testing::ByRef;
+using ::testing::NiceMock;
```
richardwhiuk: Please don't use `using` in a header file - it get's inherited by everything below it, which makes life much harder. I don't think any of this is needed by this file anyway.

src/ut/handlers_test.h
```diff
@@ -0,0 +1,148 @@
...
+#include "localstore.h"
+#include "fakehssconnection.hpp"
+#include "fakechronosconnection.hpp"
+#include "test_interposer.hpp"
+#include "mock_subscriber_data_manager.h"
+#include "mock_impi_store.h"
+#include "mock_hss_connection.h"
+#include "rapidjson/document.h"
+
+using namespace std;
```
richardwhiuk: Our coding standards forbid this - see http://clearwater.readthedocs.io/en/stable/Clearwater_CPP_Coding_Guidelines.html#c-feature-use

include/handlers.h
MatMeredith: To put the two very similar TimeoutTasks next to each other...

include/chronoshandlers.h
MatMeredith: I guess.  Maybe.  But I think the codebase uses the term "handlers" to refer to Tasks,   I.e. Tasks effectively are handlers.   See e.g. handlers.h which defines all of the sprout "handlers".   There is a TimerHandler in there but all of the other "handlers" that it defines are just "tasks".   I'm inclined to leave the naming as is for consistency, but I'll change if you are confident that it will make it less confusing.

src/ut/handlers_test.h
MatMeredith: Re. this point and the next...
- FWIW, I didn't write these -- just copied them from https://github.com/Metaswitch/sprout/blob/dev/src/ut/handlers_test.cpp when I moved the task definitions from the c file to the header.
- You're right they aren't necessary here though so I've removed them (and I take your point about now using using in a header file).

PR #1897, opened 2017-08-14
[Reviewer: HJS]  Change MMF URI's

jack-atack:
Changes to the MMF URI's now that we no longer wish to route to MMF using the BGCF port

- For pre-as, use the mmf prefix
- For post-as, use the mmf service parameter (as we need to specify the exact node to route back to)

This also removes the now redundant `namespace` parameter.

NB:  This is likely a short-term fix, overridden by future changes following further high level thinking

### DIFF
--- a/src/scscfplugin.cpp
+++ b/src/scscfplugin.cpp
@@ -132,7 +132,7 @@ bool SCSCFPlugin::load(struct options& opt, std::list<Sproutlet*>& sproutlets)
     std::string mmf_cluster_uri = "sip:mmf." + opt.sprout_hostname;
 
     // As there is not currently a shared_config option for this, form the
-    // MMF node uri manually, specifying the port (currently the BGCF port)
+    // MMF node uri manually, directly specifying the mmf service
     std::string node_host(stack_data.local_host.ptr, stack_data.local_host.slen);
 
     if (Utils::parse_ip_address(node_host) == Utils::IPV6_ADDRESS)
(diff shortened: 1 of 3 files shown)


src/scscfplugin.cpp
```diff
@@ -128,20 +128,19 @@ bool SCSCFPlugin::load(struct options& opt, std::list<Sproutlet*>& sproutlets)
       icscf_uri = opt.external_icscf_uri;
     }
 
-    // Create the MMF cluster uri.  As this is not a separate sproutlet, use
-    // the sprout hostname, and specify the port (currently the BGCF port)
-    std::string mmf_cluster_uri = "sip:" + opt.sprout_hostname + ":" + std::to_string(opt.port_bgcf);
+    // Create the MMF cluster uri
+    std::string mmf_cluster_uri = "sip:mmf." + opt.sprout_hostname;
```
AlexHockey: Why do you use `mmf.` for this URI, but `;service=mmf` for the node URI? Why not use `;service=mmf` for both?

jack-atack: There was no reason at all other than I thought using the mmf prefix is 'nicer'.

Happy to change the both to `;service=mmf` for consistency

jack-atack:
I don't think there is anything here which is going to cause problems on upgrade.  The new MMF URI's will work for routing in both the old and new world, and arguably these should changes should have been made at the same time as the alias changes.

The `namespace` parameter is overly verbose, and while this removal means that we cannot do route to old world MMF using the new world URI, we only ever route to the local MMF server so there shouldn't be any issue with upgrading one but not the other.

I don't think there is anything to be worried about here, aside from the known existing issues which I haven't attempted to tackle

PR #1901, opened 2017-08-18
[Reviewer: Seb/Richard] Sdm store rework

johadalin:
Lots of stuff has just been moved or renamed, so there's not much new to check. key points are:

* AoR has moved out of SubscriberDataManager, to stand on its own two feet
    * This has led to the mass renaming of `SubsciberDataManager::AoR...` to just `AoR...`
    * Covers AoR and AoRPair
* We now have an abstract `AoRStore` class, in aorstore.h
* We have a Memcached/Astaire implementation of this in the AstaireAoRStore, in astaire_aor_store.[h/cpp]
    * This wraps around the Connector, and the JsonSerializer classes that were in SubscriberDataManager before
* main.cpp now creates AstaireAoRStores for each of the stores that we create, and passes these into the SubscriberDataManagers constructors.

Work we should do:
* tidy up main to have a separate `get_stores` function rather than it all being messy in line. This will make it easier to implement other stores in future.
* [Maybe] create a mock AoRStore for the UTs. I'm not sure if we actually need/want this; we should discuss at some point. 

## Testing
Code builds, passes `make full_test`, and passes the live tests, so i think it's all in a good state.


include/astaire_aor_store.h
```diff
@@ -0,0 +1,120 @@
...
+{
+public:
+  /// Constructor.
+  AstaireAoRStore(Store* store);
+
+  /// Destructor.
+  virtual ~AstaireAoRStore();
+
+  // Called through to from handlers code.
+  virtual bool has_servers() { return _connector->underlying_store_has_servers(); }
```
sebrexmetaswitch: Just a style thing, but I like override methods to use the `override` keyword, so that the compiler complains if it doesn't override (which helps prevent someone breaking it in future)

include/astaire_aor_store.h
```diff
@@ -0,0 +1,120 @@
...
+    ///                 deserialized (e.g. because it is corrupt).
+    AoR* deserialize_aor(const std::string& aor_id,
+                         const std::string& s);
+  };
+
+  /// Provides the interface to the data store. This is responsible for
+  /// updating and getting information from the underlying data store. The
+  /// classes that call this class are responsible for retrying the get/set
+  /// functions in case of failure.
+  class Connector
```
sebrexmetaswitch: I don't really understand waht this `Connector` class adds beyond the store (why can't the AstaireAorStore do this stuff?) but I wasn't part of the design discussions so fine to leave it if you have a good reason to. Regardless, I don't think it hurts to have it

sebrexmetaswitch:
```/// Provides the interface to the data store. This is responsible for
 /// updating and getting information from the underlying data store.
```

I thought that's what the AstaireAorStore was for?

src/astaire_aor_store.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+AstaireAoRStore::AstaireAoRStore(Store* store) : AoRStore()
+{
+  JsonSerializerDeserializer* serializer_deserializer = new JsonSerializerDeserializer();
+  _connector = new Connector(store, serializer_deserializer);
+}
+
+AstaireAoRStore::~AstaireAoRStore()
+{
+  delete _connector; _connector = NULL;
```
sebrexmetaswitch: It surprised me at first that we don't `delete` the s`erializer_deserializer` here too. I can see that it's because we've passed ownership to `_connector`, but probably worth a comment in the `Connector`'s constructor or something explaining that it takes ownership of the serializer

src/astaire_aor_store.cpp
```diff
@@ -0,0 +1,293 @@
...
+}
+
+/// AstaireAoRStore methods
+
+/// Calls through into the connector get and set commands
+AoR* AstaireAoRStore::get_aor_data(const std::string& aor_id,
+                                   SAS::TrailId trail)
+{
+  return _connector->get_aor_data(aor_id, trail);
```
sebrexmetaswitch: Seriously, what is the value in having the `Connector`, if the methods are just passed through?

src/ut/handlers_test.cpp
```diff
@@ -55,6 +55,7 @@ const std::string HSS_NOT_REG_STATE = "<?xml version=\"1.0\" encoding=\"UTF-8\"?
                                         "<RegistrationState>NOT_REGISTERED</RegistrationState>"
                                       "</ClearwaterRegData>";
 
+
```
sebrexmetaswitch: extra line?

src/ut/scscf_test.cpp
```diff
@@ -364,7 +364,8 @@ class SCSCFTest : public SipTest
 
     _chronos_connection = new FakeChronosConnection();
     _local_data_store = new LocalStore();
-    _sdm = new SubscriberDataManager((Store*)_local_data_store, _chronos_connection, NULL, true);
+    _local_aor_store = new AstaireAoRStore(_local_data_store);
+    _sdm = new SubscriberDataManager((AoRStore*)_local_aor_store, _chronos_connection, NULL, true);
```
sebrexmetaswitch: I don't think you need to do that cast, you can just pass the `AstaireAorStore*`.

(this likely applies elsewhere too)

include/aor.h
```diff
@@ -0,0 +1,332 @@
...
+
+  /// CAS value for this AoR record.  Used when updating an existing record.
+  /// Zero for a new record that has not yet been written to a store.
+  uint64_t _cas;
+
+  // SIP URI for this AoR
+  std::string _uri;
+
+  /// Store code is allowed to manipulate bindings and subscriptions directly.
+  friend class AoRPair;
```
sebrexmetaswitch: Do we actually need this? (We didn't before, so I'm surprised we need it now)

include/astaire_aor_store.h
johadalin: Agree, we potentially should just cut out the middle man. I'll discuss with Ellie at some point, but for getting this in and working, don't need to do so yet.
That comment is just old and unchanged. I'll leave it for now, until we decide what the Connector is for

johadalin:
I've pulled in some changes Mat made to main.cpp, to pull out store and chronos_connection creation into a separate functions, here `create_astaire_stores` and `create_chronos_connection`. 
These will make it easier to put in place different store implementations in future.

Tested with both UTs and upgrading a deployment and running live tests. All pass (except current known failure of "SUBSCRIBE - reg-event with a GRUU"), so i think this is good to go. 
I'd like a review of the second attempt of cpp-common changes though, to double check. Then i'll merge in and fix anything that depends on that, and merge this.

PR #1904, opened 2017-08-22
Add route headers to sprout tests

KristaHyer:
I previously did some work reworking the sprout UTs - pull request #1890 

As part of this I accidentally cut out the code where the "Route" header was set in the scscf tests.
I've added this back in now.

I've also done minor tidying up around the codebase.

"make full_test" still passes.

### DIFF
--- a/src/ut/testingcommon.h
+++ b/src/ut/testingcommon.h
@@ -284,6 +284,16 @@ namespace TestingCommon
     std::string get_response();
   };
 
+  class SCSCFMessage : public Message
+  {
+  public:
+    SCSCFMessage()
+    {
+      Message::_route = "Route: <sip:sprout.homedomain;service=scscf>";
+    };
+    ~SCSCFMessage() {};
+  };
+
 }
 
 #endif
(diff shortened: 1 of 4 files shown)


src/ut/testingcommon.h
```diff
@@ -284,6 +284,17 @@ namespace TestingCommon
     std::string get_response();
   };
 
+  // Subclass which sets the correct Route header for the SCSCF tests.
+  class SCSCFMessage : public Message
```
sathiyan-sivathas: I think you should subclass this in the scscf_test file since that's where it's used. All the other tests don't need to see SCSCFMessage so it shouldn't go in the common file.

src/ut/testingcommon.h
```diff
@@ -284,6 +284,17 @@ namespace TestingCommon
...
+  {
+  public:
+    SCSCFMessage()
+    {
+      Message::_route = "Route: <sip:sprout.homedomain;service=scscf>";
```
sathiyan-sivathas: As mentioned in the previous review, it would be better to have a `setRoute()` method that you could call here rather than editing the member variable of the superclass directly. Can you remember to change this when you do that?

src/ut/testingcommon.h
KristaHyer: have moved

src/ut/testingcommon.h
KristaHyer: have added TODO

PR #1905, opened 2017-08-23
[Reviewer: Adam] Impistore refactor

sebrexmetaswitch:
This PR refactors the ImpiStore slightly:

* the old `ImpiStore` implementation is now an `AstaireImpiStore`, as its implementation is specific to Astaire
* the `AuthChallenge` objects are common to all possible store implementations, so remain in the `ImpiStore` rather than the `AstaireImpiStore`
* the `Impi` objects are now store-specific, so there's an `AstaireImpiStore::Impi`
* the `AuthChallenge` member variables are now private, so that Sprout must use the getters/setters to access them

### Testing
`make full_test` passes, and the live tests pass

### DIFF
--- a/include/astaire_impistore.h
+++ b/include/astaire_impistore.h
@@ -1,0 +1,101 @@
+/**
+ * @file astaire_impistore.h  Definition of class for storing IMPIs in a
+ *                            Memcached-like store
+ *
+ * Copyright (C) Metaswitch Networks 2017
+ * If license terms are provided to you in a COPYING file in the root directory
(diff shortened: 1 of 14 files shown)


sebrexmetaswitch:
One more comment - I've left the UTs using an AstaireImpiStore, rather than mocking out an ImpiStore, much as the other PR does with the AoRStore.

I think this is fine, as the tests are still testing all the same stuff

sebrexmetaswitch:
As discussed - we only care about whether an AuthChallenge has been updated, not which part of it was updated.

PR #1909, opened 2017-08-24
[Reviewer: Adam] Further impistore changes

sebrexmetaswitch:
A couple of further changes to the ImpiStore interface here.

1) A general `Impi` object no longer necessarily needs to be able to write itself as a JSON object, so that is now moved down from the `ImpiStore::Impi` to the `AstaireImpiStore::Impi`

2) I had left a `TABLE_IMPI` variable in the `ImpiStore`, which is actually an Astaire thing (It's still present in the `AstaireImpiStore`)

### DIFF
--- a/include/astaire_impistore.h
+++ b/include/astaire_impistore.h
@@ -43,7 +43,10 @@ public:
 
   private:
     /// Write to JSON writer.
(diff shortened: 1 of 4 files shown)


johadalin:
Looks reasonable :+1:

PR #1910, opened 2017-08-25
[Reviewer: Adam] Fix up handler UTs to use MockImpiStore

sebrexmetaswitch:
This is a follow-on from my PR #1905 
Just tweaks the UTs to use a MockImpiStore where appropriate

`make full_test` passes


johadalin:
Looks fine. 
Why were we deleting those impi objects before, and only in those three cases? 
Do the tests clean up adequately when they finish up? I assume so, because as you say, full_test passes. :man_shrugging:

sebrexmetaswitch:
We used to call `store->set_impi(impi, 0);`, but the caller continues to own the Impi that's set.
The handler would then (under the covers) be calling `get_impi()`, which returns a new Impi that the handler owns, and so the handler `delete`s.

So at the end, the test still owns the original Impi.

Now, however, we mock out the `get_impi()` call directly, and return the original Impi. This means that this is the Impi the handler owns, and so the handler `delete`s it, so we don't need to in the test.

PR #1911, opened 2017-08-25
[Reviewer: Adam] Set IMPU on AuthChallenge

sebrexmetaswitch:
Set the IMPU on an AuthChallenge.

`make full_test` still passes


johadalin:
Do we need UTs to just double check any function here? Otherwise sure, looks cool. Kick off the build chain :cocos_islands:

PR #1912, opened 2017-08-29
(Reviewer:Ellie) PPR sprout work - Send NOTIFYs on changed Associated URIs.

LucyBarcroft:
Sprout will be able to receive HTTP PUT requests from Homestead as a result of Push Profile requests. 

It will decode the XML, to find the Associated URIs for that IRS. It will get the AoR data from the store, update the current AoR with the Associated URIs and set it back to the store. In the case of changed Associated URIs it will send a NOTIFY to any subscribers of the registration set. It will send a return code back to Homestead based on whether it was successful in all this. 

Changed the AoR class to store the Associated URIs, in order such that it was able to determine whether they had changed, in order to know whether or not to send a NOTIFY. Thus, the AoR must be set with the associated URIs before updating the store. 

Unit Tested and make full_test passes
Live tested in the mainline case


include/sprout_xml_utils.h
```diff
@@ -0,0 +1,28 @@
...
+#include "httpconnection.h"
+#include "hssconnection.h"
+#include "rapidjson/error/en.h"
+#include "snmp_continuous_accumulator_table.h"
+#include "xml_utils.h"
+
+namespace SproutXmlUtils
+{
+bool decode_service_profile(const std::string public_user_identity,
+		            std::shared_ptr<rapidxml::xml_document<> > root,
```
eleanor-merry: Odd alignment - please fix throughout.

include/sprout_xml_utils.h
```diff
@@ -0,0 +1,28 @@
+#include <string>
+#include <vector>
+#include <memory>
```
eleanor-merry: I don't think you need these includes in the header. Can you put the ones you only need in the cpp file in the cpp file.

include/associated_uris.h
```diff
@@ -50,6 +59,26 @@ struct AssociatedURIs
   /// Add a mapping between a distinct IMPU and the wildcard it belongs to
   void add_wildcard_mapping(std::string wildcard, std::string distinct);
 
+  /// Returns the wildcard mappings.
+  std::map<std::string, std::string> get_wildcard_mappings();
```
eleanor-merry: Can this be in the same place as the other gets?

src/associated_uris.cpp
```diff
@@ -139,3 +142,91 @@ void AssociatedURIs::add_wildcard_mapping(std::string wildcard,
 {
   _distinct_to_wildcard.insert(std::make_pair(distinct, wildcard));
 }
+
+std::map<std::string, std::string> AssociatedURIs::get_wildcard_mappings()
+{
+  return _distinct_to_wildcard;
+}
+
+void AssociatedURIs::to_json(rapidjson::Writer<rapidjson::StringBuffer>& writer)
```
eleanor-merry: I'd like a comment that shows the expected format of the JSON. I think we can simplify this.

src/associated_uris.cpp
```diff
@@ -139,3 +142,91 @@ void AssociatedURIs::add_wildcard_mapping(std::string wildcard,
...
+  }
+}
+
+bool AssociatedURIs::is_equal_to(AssociatedURIs associated_uris_other)
+{
+  if (_associated_uris != associated_uris_other.get_all_uris() ||
+      get_barred_uris() != associated_uris_other.get_barred_uris() ||
+      _distinct_to_wildcard != associated_uris_other.get_wildcard_mappings())
+  {
+    TRC_DEBUG("Associated URIS is different in the 2 sets compared");
```
eleanor-merry: Not sure the TRC statements help here.

src/astaire_aor_store.cpp
```diff
@@ -223,6 +223,11 @@ AoR* AstaireAoRStore::JsonSerializerDeserializer::
       s->from_json(s_obj);
     }
 
+    JSON_ASSERT_CONTAINS(doc, JSON_ASSOCIATED_URIS);
```
eleanor-merry: I don't think you can assert this - during upgrade this won't be present.

src/handlers.cpp
```diff
@@ -831,3 +833,138 @@ void DeleteImpuTask::run()
...
 }
+
+// Will deal with requests sent from Homestead in Push Profile Requests.
+// Will send a HTTP return code to Homestead.
+void PushProfileTask::run()
+{
+  // HTTP method must be a PUT
+  if (_req.method() != htp_method_PUT)
+  {
+    TRC_WARNING("HTTP method isn't put");
```
eleanor-merry: This should be DEBUG level, and I'd rather it was something like 'Rejecting request as the HTTP method isn't PUT'.

src/handlers.cpp
```diff
@@ -831,3 +833,138 @@ void DeleteImpuTask::run()
...
+
+  if (rc != HTTP_OK)
+  {
+    TRC_WARNING("Request body is invalid, send %d", rc);
+    send_http_reply(rc);
+    delete this;
+    return;
+  }
+
+  rc = get_data(trail());
```
eleanor-merry: I'd combine the get/set into one function, and not have the aor_pair as a member variable. Also, can you come up with a more descriptive function name?

include/handlers.h
```diff
@@ -265,4 +265,43 @@ class DeleteImpuTask : public HttpStackUtils::Task
...
+  {};
+
+  void run();
+  HTTPCode parse_request(std::string body, SAS::TrailId trail);
+  HTTPCode get_data(SAS::TrailId trail);
+  HTTPCode set_data(SAS::TrailId trail);
+
+protected:
+  const Config* _cfg;
+  std::string _default_public_id;
```
eleanor-merry: Do these actually need to be member variables? I've suggested below ways to combine the functions which means you don't need to do this.

src/handlers.cpp
```diff
@@ -831,3 +833,138 @@ void DeleteImpuTask::run()
...
+  {
+    TRC_DEBUG("Successful, sending %d", rc);
+  }
+  send_http_reply(rc);
+  delete this;
+}
+
+HTTPCode PushProfileTask::parse_request(std::string body, SAS::TrailId trail)
+{
+  std::map<std::string, Ifcs> unused_ifcs_map;
```
eleanor-merry: Can you define these at the point you use them.

src/handlers.cpp
```diff
@@ -831,3 +833,138 @@ void DeleteImpuTask::run()
...
+  std::vector<std::string> aliases;
```
eleanor-merry: I think this wants to be unused_aliases as well.

Thinking about it some more, can you have SproutXmlUtils hide this - so it also has a decode_service_profile function that only takes an IMPU, the xml node, the associated URIs and a SAS trail, and then calls into the main function with the unused variables.

src/handlers.cpp
```diff
@@ -831,3 +833,138 @@ void DeleteImpuTask::run()
...
+
+  rapidxml::xml_document<>* root = new rapidxml::xml_document<>;
+  try
+  {
+    root->parse<0>(root->allocate_string(body.c_str()));
+  }
+  catch (rapidxml::parse_error& err)
+  {
+    // report to the user the failure and their locations in the document.
+    TRC_WARNING("Failed to parse XML:\n %s\n %s", body.c_str(), err.what());
```
eleanor-merry: DEBUG level.

src/sprout_xml_utils.cpp
```diff
@@ -0,0 +1,227 @@
+#include "sprout_xml_utils.h"
```
eleanor-merry: Copyright statement please

src/sprout_xml_utils.cpp
```diff
@@ -0,0 +1,227 @@
...
+namespace SproutXmlUtils
+{
+bool decode_service_profile(const std::string public_user_identity,
+			    std::shared_ptr<rapidxml::xml_document<> > root,
+    			    rapidxml::xml_node<>* node,
+			    std::map<std::string, Ifcs >& ifcs_map,
+                            AssociatedURIs& associated_uris,
+			    std::vector<std::string>& aliases,
+			    SIFCService* sifc_service,
+			    bool contain_ifc,
```
eleanor-merry: I think you can work this out from the root parameter; can you remove this option?

src/subscriber_data_manager.cpp
```diff
@@ -559,14 +558,14 @@ SubscriberDataManager::NotifySender::~NotifySender()
...
                                const std::string& aor_id,
-                               AssociatedURIs* associated_uris,
                                AoRPair* aor_pair,
                                int now,
                                SAS::TrailId trail)
 {
   std::vector<std::string> expired_binding_uris;
   ClassifiedBindings binding_info_to_notify;
   bool bindings_changed = false;
+  bool associated_uris_changed;
```
eleanor-merry: Default this to false.

src/subscriber_data_manager.cpp
```diff
@@ -648,16 +647,20 @@ void SubscriberDataManager::NotifySender::send_notifys(
...
   // Iterate over the subscriptions in the original AoR, and send NOTIFYs for
   // any subscriptions that aren't in the current AoR.
   send_notifys_for_expired_subscriptions(aor_id,
-                                         associated_uris,
                                          aor_pair,
                                          binding_info_to_notify,
                                          expired_binding_uris,
                                          now,
                                          trail);
-
```
eleanor-merry: You should keep this blank line.

src/subscriber_data_manager.cpp
```diff
@@ -648,16 +647,20 @@ void SubscriberDataManager::NotifySender::send_notifys(
     }
   }
 
+  AssociatedURIs associated_uris = aor_pair->get_current()->_associated_uris;
```
eleanor-merry: You don't need to pull these out into named variables - you can just directly compare them.

src/subscriber_data_manager.cpp
```diff
@@ -679,7 +682,7 @@ void SubscriberDataManager::NotifySender::send_notifys(
     // so don't try to check whether it's been refreshed.
     bool sub_refreshed = (!sub_created) && subscription->_refreshed;
 
-    if (bindings_changed || sub_created || sub_refreshed)
+    if (bindings_changed || associated_uris_changed || sub_created || sub_refreshed)
```
eleanor-merry: You should update the comment at the start of the for loop.

src/subscriber_data_manager.cpp
```diff
@@ -797,7 +806,7 @@ void SubscriberDataManager::NotifySender::send_notifys_for_expired_subscriptions
                                           &tdata_notify,
                                           s,
                                           aor_id,
-                                          associated_uris,
+                                          &associated_uris,
```
eleanor-merry: Just pass in the associated_uris directly.

src/ut/handlers_test.cpp
```diff
@@ -1132,3 +1132,164 @@ TEST_F(DeleteImpuTaskTest, BadMethod)
...
+  build_pushprofile_request(body, default_uri);
+
+  EXPECT_CALL(*_subscriber_data_manager, get_aor_data(default_uri, _)).WillOnce(Return(aor_pair));
+  EXPECT_CALL(*_subscriber_data_manager, set_aor_data(default_uri, aor_pair, _, _)).WillOnce(Return(Store::OK));
+
+  EXPECT_CALL(*_httpstack, send_reply(_, 200, _));
+  _task->run();
+}
+
+// The XML is not valid and therefore not able to be parsed. Sends HTTP_BAD_REQUEST.
```
eleanor-merry: Wrong test description. Also, make the XML valid in this test.

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -471,10 +612,13 @@ TEST_F(BasicSubscriberDataManagerTest, ExpiryTests)
                            "urn:uuid:00000000-0000-0000-0000-b4dd32817622:1",
                            "<sip:5102175698@192.91.191.29:59934;transport=tcp;ob>",
                            100));
-  rc = this->_store->set_aor_data(aor, &associated_uris, aor_data1, 0);
+  rc = this->_store->set_aor_data(aor, aor_data1, 0);
   EXPECT_TRUE(rc);
   delete aor_data1; aor_data1 = NULL;
 
+  EXPECT_TRUE(log.contains("Sending NOTIFY for subscription 1234: reason(s) bindings_changed subscription_created changed_associated_uris"));
+
```
eleanor-merry: Too many blank lines - there are quite a few examples of this in the UTs - can you go through and fix?

src/handlers.cpp
```diff
@@ -831,3 +833,138 @@ void DeleteImpuTask::run()
...
+  return HTTP_OK;
+}
+
+// Set the associated URIs to the current AoR and update the data to the store.
+HTTPCode PushProfileTask::set_data(SAS::TrailId trail)
+{
+  bool all_bindings_expired;
+  Store::Status set_rc;
+  _aor_pair->get_current()->_associated_uris = _associated_uris;
+  set_rc = _cfg->_sdm->set_aor_data(_default_public_id,
```
eleanor-merry: This isn't enough - you need to write to remote stores, and handle the case where all the bindings have expired. You should look at commonising the code at https://github.com/Metaswitch/sprout/blob/dev/src/handlers.cpp#L218

eleanor-merry:
Replaced by https://github.com/Metaswitch/sprout/pull/1918

PR #1913, opened 2017-08-30
[Reviewer: SS5, AJH] Made changes to sprout to use SIP Graylisting

Neel-Nanda:
Made changes to sprout to use SIP Graylisting, following Metaswitch/cpp-common#672

Not done unit tests yet


src/basicproxy.cpp
```diff
@@ -1667,6 +1670,7 @@ void BasicProxy::UACTsx::send_request()
   TRC_DEBUG("Sending request for %s",
             PJUtils::uri_to_string(PJSIP_URI_IN_REQ_URI, _tdata->msg->line.req.uri).c_str());
 
+
```
sathiyan-sivathas: Don't need this.

src/basicproxy.cpp
```diff
@@ -1676,16 +1680,25 @@ void BasicProxy::UACTsx::send_request()
...
-    PJUtils::set_dest_info(_tdata, _servers[_current_server]);
-  }
   else
   {
-    // We failed to get any valid destination servers, so fail the transaction.
-    status = PJ_ENOTFOUND;
+    // Stores the first element in servers_iter as _current_server and moves it
+    // along by 1. Next returns true if _servers_iter was non-empty, ie. a valid
+    // target was stored in _current_server.
+    _current_server_valid = _servers_iter->next(_current_server);
```
sathiyan-sivathas: `_current_server_valid` doesn't need to be a class variable since it is only used immediately after it is set. Actually you can get rid of if completely. Since `next()` returns a Boolean, you can just say `if (_servers_iter->next(_current_server))` below then set `_no_servers` in the if/else branches.

src/basicproxy.cpp
```diff
@@ -2040,10 +2064,10 @@ bool BasicProxy::UACTsx::retry_request()
...
   bool retrying = false;
 
-  // See if we have any more servers in the list returned by the resolver that
-  // we can try.
-  _current_server++;
-  if (_current_server < (int)_servers.size())
+  // See if we have any more target servers that we can try, and if so store the
+  // next one in _current_server.
+  _current_server_valid = _servers_iter->next(_current_server);
+  if (_current_server_valid)
```
sathiyan-sivathas: As above, you can move this into the if statement.

src/basicproxy.cpp
```diff
@@ -1676,16 +1680,25 @@ void BasicProxy::UACTsx::send_request()
...
-    // in the request.
-    PJUtils::set_dest_info(_tdata, _servers[_current_server]);
-  }
   else
   {
-    // We failed to get any valid destination servers, so fail the transaction.
-    status = PJ_ENOTFOUND;
+    // Stores the first element in servers_iter as _current_server and moves it
+    // along by 1. Next returns true if _servers_iter was non-empty, ie. a valid
+    // target was stored in _current_server.
```
sathiyan-sivathas: "first element of". Move what along by 1? Say this moves the iterator along by 1.

src/basicproxy.cpp
```diff
@@ -1526,8 +1526,9 @@ BasicProxy::UACTsx::UACTsx(BasicProxy* proxy,
   _index(index),
   _tsx(NULL),
   _tdata(NULL),
-  _servers(),
-  _current_server(0),
+  _servers_iter(NULL),
+  _current_server_valid(false),
+  _no_servers(true),
```
sathiyan-sivathas: Why does this default to true? I think you should default it to false then set it to true if no servers are found.

src/ut/bono_test.cpp
```diff
@@ -234,6 +234,9 @@ class StatefulProxyTestBase : public SipTest
                             bool emerg_reg_enabled = false)
   {
     SipTest::SetUpTestCase();
+    // Bono does not currently support SIP Graylisting, so this creates a SIP
```
sathiyan-sivathas: Newline before comment.

src/ut/sipresolver_test.cpp
```diff
@@ -704,3 +726,41 @@ TEST_F(SIPResolverTest, AllowedHostStateForIPAddr)
   EXPECT_EQ(1, targets.size());
   targets.pop_back();
 }
+
+// Simple test to verify that the resolve wrapper around resolve_iter for the
+// SIPResolver is working, and correctly returns a vector of the results in the
+// right order
```
sathiyan-sivathas: This test shouldn't worry about returning targets in the right order if one is blacklisted. That's all covered by the BaseResolver UTs. This test just wants to test the wrapper by checking both targets are returned in some order if they're both whitelisted. That's the simplest test that does what you need.

src/ut/basicproxy_test.cpp
```diff
@@ -3735,6 +3735,163 @@ TEST_F(BasicProxyTest, BlacklistOnTimeout)
 }
 
 
+// Checks that if there is a successful call to a graylisted target it moves to
+// the whitelist.
+TEST_F(BasicProxyTest, GraylistToWhitelistOnCallSuccess)
```
sathiyan-sivathas: I'm worried that this test can pass by chance even if the server remains on the graylist. Let's discuss if there are any better ways to test this.

src/basicproxy.cpp
Neel-Nanda: When, initially, _servers_iter is null and no servers have yet been resolved, I thought it would make most sense to say that there are no servers, since initially there aren't any.

src/ut/sipresolver_test.cpp
```diff
@@ -728,6 +729,44 @@ TEST_F(SIPResolverTest, AllowedHostStateForIPAddr)
...
+
+  EXPECT_EQ(2, targets.size());
+  EXPECT_THAT(result_str_1, MatchesRegex(whitelist_regex));
+  EXPECT_THAT(result_str_2, MatchesRegex(whitelist_regex));
+
+  // Verifies that the same target wasn't returned twice.
+  EXPECT_NE(result_str_1, result_str_2);
+}
+
+/*
```
sathiyan-sivathas: You should delete the commented out test.

include/sipresolver.h
```diff
@@ -18,9 +18,24 @@
 class SIPResolver : public BaseResolver
 {
 public:
+  SIPResolver(DnsCachedResolver* dns_client)
+    : SIPResolver(dns_client, DEFAULT_BLACKLIST_DURATION)
```
AlexHockey: We the sort of colon on the same line as the closing bracket.

include/sipresolver.h
```diff
@@ -18,9 +18,24 @@
...
+  }
+
+  SIPResolver(DnsCachedResolver* dns_client,
+              int blacklist_duration)
+    : SIPResolver(dns_client, blacklist_duration, blacklist_duration)
+  {
+    // Graylist duration is not configurable so it defaults to whatever the
+    // blacklist duration has been set to. This is desirable since if there is a
+    // need to increase graylist duration, eg low call load, this also means
+    // blacklist duration should increase.
```
AlexHockey: It's good that you're trying to comment on this sort of thing. However in this instance the comment is in the wrong place. The `SIPResolver` has three way to configure it, and it *does* allow the graylist time to be set independently of the blacklist time. It's just that sprout does not make use of that feature.

PR #1914, opened 2017-08-30
[Reviewer: Richard] Utility functions in aor.cpp to find modified bindings and subscriptions

johadalin:
Mostly as discussed:
Functions added to AoRPair in aor.cpp to compare original and current AoRs in the pair, and return a list of updated/removed bindings/subscriptions.

I have left it with both sets of functions returning a full `Bindings` or `Subscriptions` map, so that these can be more generic, and potentially more useful in future. Can't see a reason not to do this, but you might have thoughts.

Otherwise `make` and `make full_test` both pass, so think all is good

### DIFF
--- a/src/aor.cpp
+++ b/src/aor.cpp
@@ -556,7 +556,7 @@ AoR::Subscriptions AoRPair::get_updated_subscriptions()
       // The subscription is in both AoRs. Check if the expiry time has changed at all
       if (orig_aor_subscription_match->second->_expires != subscription->_expires)
       {
-        TRC_DEBUG("Subscription %s expiry has been changed refreshed", s_id.c_str());
+        TRC_DEBUG("Subscription %s expiry has been changed", s_id.c_str());
         updated_subscriptions.insert(std::make_pair(s_id, subscription));
       }
       else
@@ -565,6 +565,7 @@ AoR::Subscriptions AoRPair::get_updated_subscriptions()
       }
     }
   }
+
   return updated_subscriptions;
 }
 
@@ -580,11 +581,12 @@ AoR::Bindings AoRPair::get_removed_bindings()
         _current_aor->bindings().end())
     {
       // Binding is gone (which may mean deregistration or expiry)
-      TRC_DEBUG("The binding (%s) has been removed", orig_aor_binding.first.c_str());
+      TRC_DEBUG("Binding %s has been removed", orig_aor_binding.first.c_str());
       removed_bindings.insert(std::make_pair(orig_aor_binding.first,
                                              orig_aor_binding.second));
     }
   }
+
   return removed_bindings;
 }
 
@@ -601,7 +603,7 @@ AoR::Subscriptions AoRPair::get_removed_subscriptions()
         _current_aor->subscriptions().end())
     {
       // Subscription is gone
-      TRC_DEBUG("The subscription (%s) is no longer present",
+      TRC_DEBUG("Subscription %s is no longer present",
                     orig_aor_subscription.first.c_str());
       removed_subscriptions.insert(std::make_pair(orig_aor_subscription.first,
                                                   orig_aor_subscription.second));
--- a/src/aor.cpp
+++ b/src/aor.cpp
@@ -489,3 +489,123 @@ void AoR::copy_subscriptions_and_bindings(AoR* source_aor)
     *dst = *src;
   }
 }
+
+AoR::Bindings AoRPair::get_updated_bindings()
+{
+  AoR::Bindings updated_bindings;
+
+  // Iterate over the bindings in the current AoR. Figure out if the bindings
+  // have been created or updated.
+  for (std::pair<std::string, AoR::Binding*> current_aor_binding :
+         _current_aor->bindings())
+  {
+    std::string b_id = current_aor_binding.first;
+    AoR::Binding* binding = current_aor_binding.second;
+
+    // Find any binding match in the original AoR
+    AoR::Bindings::const_iterator orig_aor_binding_match =
+      _orig_aor->bindings().find(b_id);
+
+    // If the binding is only in the current AoR, it has been created
+    if (orig_aor_binding_match == _orig_aor->bindings().end())
+    {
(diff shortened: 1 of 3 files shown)


src/aor.cpp
```diff
@@ -489,3 +489,123 @@ void AoR::copy_subscriptions_and_bindings(AoR* source_aor)
...
+        _current_aor->subscriptions().end())
+    {
+      // Subscription is gone
+      TRC_DEBUG("The subscription (%s) is no longer present",
+                    orig_aor_subscription.first.c_str());
+      removed_subscriptions.insert(std::make_pair(orig_aor_subscription.first,
+                                                  orig_aor_subscription.second));
+    }
+  }
+  return removed_subscriptions;
```
richardwhiuk: Missing new line

src/aor.cpp
```diff
@@ -489,3 +489,123 @@ void AoR::copy_subscriptions_and_bindings(AoR* source_aor)
...
+        TRC_DEBUG("Subscription %s expiry has been changed refreshed", s_id.c_str());
+        updated_subscriptions.insert(std::make_pair(s_id, subscription));
+      }
+      else
+      {
+        TRC_DEBUG("Subscription %s is unchanged", s_id.c_str());
+      }
+    }
+  }
+  return updated_subscriptions;
```
richardwhiuk: Missing new line.

src/aor.cpp
```diff
@@ -489,3 +489,123 @@ void AoR::copy_subscriptions_and_bindings(AoR* source_aor)
...
+    if (_current_aor->bindings().find(orig_aor_binding.first) ==
+        _current_aor->bindings().end())
+    {
+      // Binding is gone (which may mean deregistration or expiry)
+      TRC_DEBUG("The binding (%s) has been removed", orig_aor_binding.first.c_str());
+      removed_bindings.insert(std::make_pair(orig_aor_binding.first,
+                                             orig_aor_binding.second));
+    }
+  }
+  return removed_bindings;
```
richardwhiuk: Missing new line.

src/aor.cpp
```diff
@@ -489,3 +489,123 @@ void AoR::copy_subscriptions_and_bindings(AoR* source_aor)
...
+    {
+      TRC_DEBUG("Subscription %s has been created", current_aor_subscription.first.c_str());
+      updated_subscriptions.insert(std::make_pair(s_id, subscription));
+    }
+    else
+    {
+      // The subscription is in both AoRs. Check if the expiry time has changed at all
+      if (orig_aor_subscription_match->second->_expires != subscription->_expires)
+      {
+        TRC_DEBUG("Subscription %s expiry has been changed refreshed", s_id.c_str());
```
richardwhiuk: This phrase doesn't make sense

src/aor.cpp
```diff
@@ -489,3 +489,123 @@ void AoR::copy_subscriptions_and_bindings(AoR* source_aor)
...
+
+  // Iterate over original bindings and record those not in current AoR
+  for (std::pair<std::string, AoR::Binding*> orig_aor_binding :
+         _orig_aor->bindings())
+  {
+    if (_current_aor->bindings().find(orig_aor_binding.first) ==
+        _current_aor->bindings().end())
+    {
+      // Binding is gone (which may mean deregistration or expiry)
+      TRC_DEBUG("The binding (%s) has been removed", orig_aor_binding.first.c_str());
```
richardwhiuk: This, and the function below use parentheses, unlike the above ones which dno't.

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -1062,3 +1062,90 @@ TEST_F(SubscriberDataManagerChronosRequestsTest, AoRTimerBadRequestNoIDTest)
 
   delete aor_data1; aor_data1 = NULL;
 }
+
+TEST_F(BasicSubscriberDataManagerTest, AoRComparisonTests)
```
richardwhiuk: This would be nicer if it was multiple tests which all tested one singular thing - e.g.

- AorPairComparisonCreatedSubscription
- AorPairComparisonUpdatedSubscription
- AorPairComparisonDeletedSubscription

and ditto for Bindings.

Generally UTs should:

- Perform some setup
- Make a call on the unit under test
- Check the results.

This makes them small, fast, and isolated.

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -1062,3 +1062,90 @@ TEST_F(SubscriberDataManagerChronosRequestsTest, AoRTimerBadRequestNoIDTest)
...
+  // Add the AoR record to the store.
+  std::string aor = "5102175698@cw-ngv.com";
+  AssociatedURIs associated_uris = {};
+  associated_uris.add_uri(aor, false);
+  rc = this->_store->set_aor_data(aor, &associated_uris, aor_data1, 0);
+  EXPECT_TRUE(rc);
+  delete aor_data1; aor_data1 = NULL;
+
+  // Get the AoR record from the store.
+  aor_data1 = this->_store->get_aor_data(std::string("5102175698@cw-ngv.com"), 0);
```
richardwhiuk: It'd be nicer if we didn't have to interact with the store to test the AorPair - e.g. by creating two Aors, and using the constructor to put it into the right state: https://github.com/Metaswitch/sprout/blob/8c84cfb014ceb519e0aec15dae32e8b472a65542/include/aor.h#L291-L294

src/ut/subscriber_data_manager_test.cpp
richardwhiuk: This should really be part of the SDM tests, but fine.

johadalin:
Should have resolved the issues above with this commit. Full coverage of the new functions still present, and `make full_test` is passing fine, as is `make`, so i'm going to merge this down and call it job done unless somehow Jenkins kicks me :P

johadalin:
Though probably should take some time soon to move the new tests out into an aor_test file, as well as creating an aor_store test suite split out from the sdm tests

PR #1918, opened 2017-09-01
(Reviewer:Ellie) final PPR sprout work - Send NOTIFYs on changed Associated URIs

LucyBarcroft:
Sprout will be able to receive HTTP PUT requests from Homestead as a result of Push Profile requests.

It will decode the XML, to find the Associated URIs for that IRS. It will get the AoR data from the store, update the current AoR with the Associated URIs and set it back to the store. In the case of changed Associated URIs it will send a NOTIFY to any subscribers of the registration set. It will send a return code back to Homestead based on whether it was successful in all this.

Changed the AoR class to store the Associated URIs, in order such that it was able to determine whether they had changed, in order to know whether or not to send a NOTIFY. Thus, the AoR must be set with the associated URIs before updating the store.


src/astaire_aor_store.cpp
```diff
@@ -184,15 +184,13 @@ AoR* AstaireAoRStore::JsonSerializerDeserializer::
               rapidjson::GetParseError_En(doc.GetParseError()));
     return NULL;
   }
-
```
eleanor-merry: Too aggressive on removing blank lines now!

src/handlers.cpp
```diff
@@ -831,3 +833,176 @@ void DeleteImpuTask::run()
...
+  {
+    TRC_WARNING("Request body is invalid, send %d", rc);
+    send_http_reply(rc);
+    delete this;
+    return;
+  }
+
+  rc = update_associated_uris(trail());
+
+  if (rc != HTTP_OK)
```
eleanor-merry: You do the same thing in each case here - do you need to split these out?

src/handlers.cpp
```diff
@@ -831,3 +833,176 @@ void DeleteImpuTask::run()
...
+  {
+    TRC_INFO("Failed to parse data as JSON: %s\nError: %s",
+             body.c_str(),
+             rapidjson::GetParseError_En(doc.GetParseError()));
+    return HTTP_BAD_REQUEST;
+  }
+
+  try
+  {
+    JSON_GET_STRING_MEMBER(doc, "user-data-xml", user_data_xml);
```
eleanor-merry: Can user-data-xml be defined as a constant?

src/handlers.cpp
```diff
@@ -831,3 +833,176 @@ void DeleteImpuTask::run()
...
+  }
+  catch (JsonFormatError err)
+  {
+    TRC_WARNING("User data not available in the JSON");
+    return HTTP_BAD_REQUEST;
+  }
+
+  const std::string prefix = "/registrations/";
+  std::string full_path = _req.full_path();
```
eleanor-merry: Do you actually need this anymore?

src/handlers.cpp
```diff
@@ -831,3 +833,176 @@ void DeleteImpuTask::run()
...
+  }
+  else
+  {
+    TRC_DEBUG("Successful, sending %d", rc);
+  }
+  send_http_reply(rc);
+  delete this;
+}
+
+HTTPCode PushProfileTask::parse_request(std::string body, SAS::TrailId trail)
```
eleanor-merry: parse_request isn't very descriptive - I think you want something more like get_associated_uris.

src/handlers.cpp
```diff
@@ -831,3 +833,176 @@ void DeleteImpuTask::run()
...
+    delete root;
+    root = NULL;
+    return HTTP_BAD_REQUEST;
+  }
+
+  rapidxml::xml_node<>* imss = root->first_node(RegDataXMLUtils::IMS_SUBSCRIPTION);
+
+  // Decode service profile from the XML. Create and populate an instance of the
+  // Associated URIs class
+  if (SproutXmlUtils::get_uris_from_service_profile(imss,
```
eleanor-merry: You could go for rc = SproutXmlUtils..., then delete root.. then return rc ? HTTP_OK : HTTP_BAD_REQUEST;

src/handlers.cpp
```diff
@@ -831,3 +833,176 @@ void DeleteImpuTask::run()
...
+    delete root; root = NULL;
+    return HTTP_BAD_REQUEST;
+  }
+}
+
+HTTPCode PushProfileTask::update_associated_uris(SAS::TrailId trail)
+{
+  AoRPair* aor_pair = NULL;
+
+  if(!sdm_access_common(&aor_pair,
```
eleanor-merry: space between if and ( please

src/handlers.cpp
```diff
@@ -831,3 +833,176 @@ void DeleteImpuTask::run()
...
+    TRC_DEBUG("Could not get AoR data from SDM");
+    return HTTP_SERVER_ERROR;
+  }
+
+  TRC_DEBUG("Obtained AoR data");
+
+  bool all_bindings_expired = false;
+  Store::Status set_rc;
+  aor_pair->get_current()->_associated_uris = _associated_uris;
+  set_rc = _cfg->_sdm->set_aor_data(_default_public_id,
```
eleanor-merry: You've not dealt with contention here yet.

src/handlers.cpp
```diff
@@ -831,3 +833,176 @@ void DeleteImpuTask::run()
...
+  }
+  catch (JsonFormatError err)
+  {
+    TRC_WARNING("User data not available in the JSON");
+    return HTTP_BAD_REQUEST;
+  }
+
+  const std::string prefix = "/registrations/";
+  std::string full_path = _req.full_path();
+  size_t end_of_impu = full_path.length();
```
eleanor-merry: Ah, you do for the all bindings expired case.

PR #1919, opened 2017-09-04
[Reviewer: Richard] Allow impistore to store expiry in ms or s

sebrexmetaswitch:
As discussed, we want the ImpiStore to be able to store the expiry time in ms or s.

This adds that, defaulting to the usual seconds.

`make full_test` passes

### DIFF
--- a/include/impistore.h
+++ b/include/impistore.h
@@ -312,7 +312,7 @@ public:
 
     /// Write to JSON writer (IMPI format).
     virtual void write_json(rapidjson::Writer<rapidjson::StringBuffer>* writer,
-                            bool expiry_im_ms = false) override;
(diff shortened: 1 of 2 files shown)


include/impistore.h
```diff
@@ -308,7 +311,8 @@ class ImpiStore
     virtual ~AKAAuthChallenge() {};
 
     /// Write to JSON writer (IMPI format).
-    virtual void write_json(rapidjson::Writer<rapidjson::StringBuffer>* writer);
+    virtual void write_json(rapidjson::Writer<rapidjson::StringBuffer>* writer,
+                            bool expiry_im_ms = false) override;
```
richardwhiuk: in_ms

PR #1922, opened 2017-09-05
[Reviewer RJW2] Ensure content type appears once in bodyparts

steelyvoid:
Hi Richard

Here's the Sprout PR companion to https://github.com/Metaswitch/pjsip-upstream/pull/68. Note that in addition to extending the UTs to check that bodyparts are as expected, I've added a verbose log of the 3rd party REGISTER (much as we have for sproutet-sproutlet traffic) so that you can see what it is in UTs (when you don't have the benefit of SAS).

Cheers

Steve

### DIFF
--- a/src/registration_utils.cpp
+++ b/src/registration_utils.cpp
@@ -563,7 +563,7 @@ static void send_register_to_as(SubscriberDataManager* sdm,
   // Set the SAS trail on the request.
   set_trail(tdata, trail);
 
-  // Verbose trace
+  if (Log::enabled(Log::VERBOSE_LEVEL))
   {
     char buf[PJSIP_MAX_PKT_LEN];
     pj_ssize_t size;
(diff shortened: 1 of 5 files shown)


src/registration_utils.cpp
```diff
@@ -563,6 +563,33 @@ static void send_register_to_as(SubscriberDataManager* sdm,
   // Set the SAS trail on the request.
   set_trail(tdata, trail);
 
+  // Verbose trace
```
richardwhiuk: This is quite expensive (doing a full print of the SIP message from struct to buffer, which we may end up not logging), so can we put `if (Log::enabled(Log::VERBOSE_LEVEL))`. You can probably get rid of the comment then as well.

src/ut/siptest.hpp
```diff
@@ -275,15 +275,19 @@ class PjMsg
 class MsgMatcher
 {
 public:
-  MsgMatcher(string expected_body="") :
-    _expected_body(expected_body)
+  MsgMatcher(string expected_body="",
+             string body_regex="") :
+    _expected_body(expected_body),
+    _body_regex(body_regex)
```
richardwhiuk: Could have just added another constructor here. Doesn't really matter though.

PR #1924, opened 2017-09-05
Add support for urn URI

KristaHyer:
We want to support a urn URI in the Request header, as this will be passed to us in a MESSAGE from Perimeta if a user makes an emergency call.

I've changed the code so that we support this URI, and written several UTs to make sure we support it.

"make full_test" passes.


src/uri_classifier.cpp
```diff
@@ -203,6 +203,14 @@ URIClass URIClassifier::classify_uri(const pjsip_uri* uri, bool prefer_sip, bool
     }
   }
 
+  // Finally, check if this is a urn URI.
+  // We support the URI "urn:service:sos", which is sent from Perimeta when a
+  // user makes an emergency call.
+  if (PJSIP_URI_SCHEME_IS_URN(uri))
```
eleanor-merry: Is this safe? I wonder if it's better to leave this as UNKNOWN?

src/uri_classifier.cpp
```diff
@@ -203,6 +203,14 @@ URIClass URIClassifier::classify_uri(const pjsip_uri* uri, bool prefer_sip, bool
...
+  {
```
eleanor-merry: If you keep this, can you put it in the same if/else-if as for SIP and Tel URIs?

src/ut/icscfsproutlet_test.cpp
```diff
@@ -3614,3 +3614,71 @@ TEST_F(ICSCFSproutletTest, RouteOutOfDialogAck)
   poll();
   delete tp;
 }
+
+// Test the case where the Request URI is "urn:service:sos".
+// This will be received from Perimeta when a subscriber has made an emergency
```
eleanor-merry: As before, we shouldn't have this type of comment in the code.

src/ut/icscfsproutlet_test.cpp
```diff
@@ -3614,3 +3614,71 @@ TEST_F(ICSCFSproutletTest, RouteOutOfDialogAck)
...
+  // Check that the MESSAGE is forwarded as expected.
+  ASSERT_EQ(1, txdata_count());
+  tdata = current_txdata();
+
+  // Check the method and content of the MESSAGE is correct.
+  ReqMatcher r1("MESSAGE");
+  r1.matches(tdata->msg);
+  expect_target("TCP", "10.10.10.1", 5058, tdata);
+  EXPECT_EQ("urn:service:sos", r1.uri());
+  EXPECT_THAT(get_headers(current_txdata()->msg, "To"),
```
eleanor-merry: The To header should probably be the subscriber.

src/ut/icscfsproutlet_test.cpp
```diff
@@ -3614,3 +3614,71 @@ TEST_F(ICSCFSproutletTest, RouteOutOfDialogAck)
...
+              testing::MatchesRegex("To: <urn:service:sos>"));
+  EXPECT_THAT(get_headers(current_txdata()->msg, "Route"),
```
eleanor-merry: I don't think you need to check the Route or the P-Asserted-Identity, but I would check the Req URI

src/ut/scscf_test.cpp
```diff
@@ -9290,3 +9290,76 @@ TEST_F(SCSCFTest, MMFPreAndPostAs)
...
+// emergency call. This message should be passed on to the terminating TAS for
+// that user (through configured iFCs) to alert the TAS not to apply any call
+// blocking, etc. on calls to that subscriber for a period of time (in case the
+// emergency call is dropped and the emergency services need to call back).
+// This URI should be accepted, and the MESSAGE should be forwarded onto the
+// server specified in the iFC. This server should terminate the MESSAGE, and
+// return a 200 OK, which should then be forwarded back to the source.
+// (NOTE: Since the TAS should terminate the MESSAGE, if there are multiple
+// terminating TASs for a user, only one will be notified.)
+TEST_F(SCSCFTest, SCSCFHandlesUrnUri)
```
eleanor-merry: Same comments as for the I-CSCF test

src/ut/testingcommon.cpp
```diff
@@ -293,6 +293,21 @@ std::string Message::get_request()
   std::string requri = target;
   std::string route = _route.empty() ? "" : _route + "\r\n";
 
+  // Set the value of the To header.
```
eleanor-merry: Rather than using _urn_uri, I'd rather you just set the To header and the Req URI separately.

src/bono.cpp
```diff
@@ -675,9 +675,10 @@ static int proxy_verify_request(pjsip_rx_data *rdata)
   // This would have been checked by transport layer.
 
   // 2. URI scheme.
-  // We support "sip:" and "tel:" URI schemes in this simple proxy.
+  // We support "sip:", "tel:" amd "urn:" URI schemes in this simple proxy.
```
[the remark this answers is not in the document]
KristaHyer: done

src/ifc.cpp
```diff
@@ -280,6 +280,15 @@ bool Ifc::spt_matches(const SessionCase& session_case,  //< The session case
       // of 3GPP TS 29.228.
       test_string = PJUtils::pj_str_to_string(&req_uri->number);
     }
+    else if (PJSIP_URI_SCHEME_IS_URN(msg->line.req.uri))
+    {
+      pjsip_other_uri* req_uri = (pjsip_other_uri*)pjsip_uri_get_uri(msg->line.req.uri);
+
+      // There is nothing in TS Spec 29.228 about what to match against in the
```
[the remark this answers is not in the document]
KristaHyer: done

src/ut/fakehssconnection.cpp
```diff
@@ -187,10 +187,6 @@ long FakeHSSConnection::get_xml_object(const std::string& path,
     catch (rapidxml::parse_error& err)
     {
       // report to the user the failure and their locations in the document.
-      printf("Failed to parse Homestead response:\n %s\n %s\n %s\n",
```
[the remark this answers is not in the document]
KristaHyer: done

src/ut/icscfsproutlet_test.cpp
KristaHyer: done

src/ut/fakehssconnection.cpp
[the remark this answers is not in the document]
eleanor-merry: What did you do?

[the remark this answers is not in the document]
KristaHyer: oops I always forget this notifies you. This just means "I'm done with this comment" so for this I didn't make any changes

src/ut/icscfsproutlet_test.cpp
KristaHyer: Making no change here, as the Perimeta spec has "urn:service:sos" in the To header.
https://metaswitch.sharepoint.com/sites/perimeta/_layouts/15/WopiFrame.aspx?sourcedoc=%7Bcbb3dd16-d48a-490a-8dc1-2857a2755827%7D&action=default

src/ut/icscfsproutlet_test.cpp
KristaHyer: The request uri is checked here, on the line which says:
EXPECT_EQ("urn:service:sos", r1.uri()).

I thought it was safest to check the other headers:
-the P-Asserted-Id is where the AS will get the info about the caller from, so I wanted to make sure it was right.
-the iFCs will only match if the "orig" stays in the Route header, so I wanted to make sure it did.

Are these headers just so mainline that we shouldn't worry about them?

src/uri_classifier.cpp
KristaHyer: Turns out make full_test passes when this is set to UNKNOWN, so for safety I will leave it as UNKNOWN

PR #1925, opened 2017-09-07
[Reviewer: Seb] Allow Tel URIs in reg-event SUBSCRIBEs

richardwhiuk:
Seb,

We've seen that some UEs put a Tel URI in the Request URI when subscribing for reg-event.

This is fine, because they've all registered with that as an IMPU.

I've made a small tweak to the Sprout code to allow this to work, and then dealt with a bunch of fall out in the tests.


sebrexmetaswitch:
Looks reasonable. (I'm assuming that `make full_test` passes)

richardwhiuk:
Yep it does.

PR #1926, opened 2017-09-11
Log Call-IDs from In-Reply-To headers as markers to SAS

metaswitch-colin-haywood:
Idea is to correlate related SIP MESSAGE flows together so they show in SAS as a single trail group (search result). (Fix is limited in scope to SIP MESSAGEs.)

Alex - feel free to assign to someone else, just picked you because of your timely help on the RTR/PPR issue. :-)

### DIFF
--- a/include/pjutils.h
+++ b/include/pjutils.h
@@ -226,7 +226,9 @@ void create_random_token(size_t length, std::string& token);
 
 std::string get_header_value(pjsip_hdr*);
 
-void mark_sas_call_branch_ids(const SAS::TrailId trail, pjsip_cid_hdr* cid_hdr, pjsip_msg* msg);
+void mark_sas_call_branch_ids(const SAS::TrailId trail,
+                              pjsip_msg* msg,
+                              const std::vector<std::string>& cids = std::vector<std::string>());
 
 bool is_emergency_registration(pjsip_contact_hdr* contact_hdr);
 
--- a/src/common_sip_processing.cpp
+++ b/src/common_sip_processing.cpp
@@ -207,9 +207,55 @@ static void sas_log_rx_msg(pjsip_rx_data* rdata)
   {
     PJUtils::report_sas_to_from_markers(trail, rdata->msg_info.msg);
 
+    std::vector<std::string> call_ids;
+
     pjsip_cid_hdr* cid = (pjsip_cid_hdr*)rdata->msg_info.cid;
 
-    PJUtils::mark_sas_call_branch_ids(trail, cid, rdata->msg_info.msg);
+    if (cid != NULL)
+    {
+      call_ids.push_back(PJUtils::pj_str_to_string(&cid->id));
+    }
+
+    // If this is a SIP MESSAGE then also pull out any In-Reply-To
+    // headers in order to correlate this trail with the trails
+    // for the calls identified in those headers.
+    pjsip_method* method = &rdata->msg_info.msg->line.req.method;
+
+    if ((method->id == PJSIP_OTHER_METHOD) &&
+        (pj_strcmp2(&method->name, "MESSAGE") == 0))
+    {
+      TRC_DEBUG("MESSAGE method - pull out In-Reply-To headers");
+      pjsip_in_reply_to_hdr* in_reply_to;
+
+      for (in_reply_to = (pjsip_in_reply_to_hdr*)pjsip_msg_find_hdr_by_name(rdata->msg_info.msg,
+                                                                          &STR_IN_REPLY_TO,
+                                                                          NULL);
+           in_reply_to != NULL;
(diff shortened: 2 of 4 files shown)


include/pjutils.h
```diff
@@ -226,7 +226,7 @@ void create_random_token(size_t length, std::string& token);
 
 std::string get_header_value(pjsip_hdr*);
 
-void mark_sas_call_branch_ids(const SAS::TrailId trail, pjsip_cid_hdr* cid_hdr, pjsip_msg* msg);
+void mark_sas_call_branch_ids(const SAS::TrailId trail, pjsip_msg* msg, const std::vector<std::string>& cids = std::vector<std::string>());
```
eleanor-merry: Can you put the parameters on separate lines? (yes, I know that the rest of the file doesn't follow the coding standards, but it's worth fixing it up given you're changing the line anyway).

src/pjutils.cpp
```diff
@@ -1693,11 +1693,14 @@ void PJUtils::mark_sas_call_branch_ids(const SAS::TrailId trail, pjsip_cid_hdr*
                                  (pjsip_method_cmp(&msg->line.req.method, pjsip_get_subscribe_method()) == 0) ||
                                  (pjsip_method_cmp(&msg->line.req.method, pjsip_get_notify_method()) == 0)));
 
-  if (cid_hdr != NULL)
+  for (std::vector<std::string>::const_iterator it = cids.begin();
```
eleanor-merry: Can this be `for (std::string cid : cids)` instead?

src/pjutils.cpp
```diff
@@ -1671,8 +1671,8 @@ std::string PJUtils::get_header_value(pjsip_hdr* header)
   return std::string(buf2, len);
 }
 
-/// Add SAS markers for the specified call ID and branch IDs on the message (call ID may be omitted, but not message).
-void PJUtils::mark_sas_call_branch_ids(const SAS::TrailId trail, pjsip_cid_hdr* cid_hdr, pjsip_msg* msg)
+/// Add SAS markers for the specified call IDs and branch IDs on the message (msg must not be NULL).
```
eleanor-merry: Can you keep the lines to 80 characters?

src/common_sip_processing.cpp
```diff
@@ -207,9 +207,50 @@ static void sas_log_rx_msg(pjsip_rx_data* rdata)
...
+    // If this is a SIP MESSAGE then also pull out any In-Reply-To
+    // headers in order to correlate this trail with the trails
+    // for the calls identified in those headers.
+    pjsip_method* method = &rdata->msg_info.msg->line.req.method;
+
+    if ((method->id == PJSIP_OTHER_METHOD) &&
+        (pj_strcmp2(&method->name, "MESSAGE") == 0))
+    {
+      TRC_DEBUG("MESSAGE method - pull out In-Reply-To headers");
+      pjsip_in_reply_to_hdr* inreplyto;
```
eleanor-merry: I don't really mind, but I'd rather this was in_reply_to rather than inreplyto

src/common_sip_processing.cpp
```diff
@@ -207,9 +207,50 @@ static void sas_log_rx_msg(pjsip_rx_data* rdata)
...
+                                                                          &STR_IN_REPLY_TO,
+                                                                          NULL);
+           inreplyto != NULL;
+           inreplyto = (pjsip_in_reply_to_hdr*)pjsip_msg_find_hdr_by_name(rdata->msg_info.msg,
+                                                                          &STR_IN_REPLY_TO,
+                                                                          inreplyto->next))
+      {
+        TRC_DEBUG("Found In-Reply-To header %.*s", inreplyto->hvalue.slen, inreplyto->hvalue.ptr);
+
+        // Split the header value by commas. Each resulting value is a Call-ID.
```
eleanor-merry: Can there be whitespace here that will throw the correlation off?

src/common_sip_processing.cpp
metaswitch-colin-haywood: Fixed.

src/pjutils.cpp
metaswitch-colin-haywood: Sure. (Wasn't sure if we supported C++11)

src/pjutils.cpp
metaswitch-colin-haywood: This isn't the 1980s, is it? :) Fixed anyway.

include/pjutils.h
metaswitch-colin-haywood: Done.

src/common_sip_processing.cpp
metaswitch-colin-haywood: Shouldn't be, according to the spec. But you can't trust third-party boxes so I added some trimming.

metaswitch-colin-haywood:
In regard to testing, I wrote this live test https://github.com/Metaswitch/clearwater-live-test-private/pull/96 and checked that it produced a single SAS trace rather than two.

PR #1928, opened 2017-09-13
[Reviewer: Richard] Increase Sprout's timeout to Homestead to 750ms

sebrexmetaswitch:
This change is to accompany https://github.com/Metaswitch/homestead/pull/475

We expect that initial registers will take over 500ms in a GR deployment, so we need to increase the timeout Sprout specifies on its requests to Homestead. It defaulted to 500ms before, but this increases it to 750ms.

(All the other options in the constructor are the defaults from https://github.com/Metaswitch/cpp-common/blob/master/include/httpconnection.h#L31)

`make full_test` and the live tests pass.

### DIFF
--- a/sprout-base.root/etc/init.d/sprout
+++ b/sprout-base.root/etc/init.d/sprout
@@ -105,7 +105,7 @@ get_settings()
         # Calculate the correct homestead latency:
         #  - if we have sprout_homestead_timeout_ms set, use that
         #  - else:
-        #    - if we have diameter_timeout_ms set, use 550 + that
+        #    - if we have diameter_timeout_ms set, use 550 + double that
         #    - else use 750
         if [ -z "$sprout_homestead_timeout_ms" ];
         then
@@ -113,7 +113,7 @@ get_settings()
           then
             sprout_homestead_timeout_ms=750
           else
-            sprout_homestead_timeout_ms=$((diameter_timeout_ms + 550))
+            sprout_homestead_timeout_ms=$((diameter_timeout_ms * 2 + 550))
           fi
         fi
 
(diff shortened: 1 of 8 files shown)


rkd-msw:
@pdavies - this might be interesting from a 4-site GR point of view (i.e. 100ms intra-site latency may break again then).

richardwhiuk:
@rkd I think for 4 site GR we'll need to exploit some more parallelism in both Sprout and Homestead to avoid unacceptable latency.

sebrexmetaswitch:
I've now made the timeout configurable via a new config option: `sprout_homestead_timeout_ms`. If not set, it will default to `diameter_timeout_ms + 550`. If that's not set, it will default to 750

I've live tested that changing these config options work (i.e. sprout picks them up correctly and passes them to the HssConnection)

sprout-base.root/etc/init.d/sprout
```diff
@@ -102,6 +102,21 @@ get_settings()
...
+        #  - else:
+        #    - if we have diameter_timeout_ms set, use 550 + that
+        #    - else use 750
+        if [ -z "$sprout_homestead_timeout_ms" ];
+        then
+          if [ -z "$diameter_timeout_ms" ];
+          then
+            sprout_homestead_timeout_ms=750
+          else
+            sprout_homestead_timeout_ms=$((diameter_timeout_ms + 550))
```
richardwhiuk: As discussed, we think this should be twice the diameter timeout.

src/main.cpp
```diff
@@ -442,6 +444,7 @@ static void usage(void)
        "                            then the iFC is skipped over.\n"
        "     --http-acr-logging     Whether to include the bodies of ACR HTTP requests when they are logged \n"
        "                            to SAS\n"
+       "     --homestead-timeout    The timeout to use on HTTP requests to Homestead\n"
```
richardwhiuk: Should we say 'in ms' here?

PR #1932, opened 2017-09-15
Move Mock constructor / destructors out of headers

MatMeredith:


src/ut/mock_sproutlet.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+MockSproutlet::~MockSproutlet() {}
+
+MockSproutletTsx::MockSproutletTsx() :
+    SproutletTsx(NULL)
+{}
+
+MockSproutletTsx::~MockSproutletTsx() {}
+
+
```
richardwhiuk: Be nice to not have bogus blank lines at the bottom of the file.

PR #1933, opened 2017-09-15
[Reviewer: Adam] Scscf uri on as chain

sebrexmetaswitch:
This PR is a fix for https://github.com/Metaswitch/clearwater-issues/issues/2418

There's lots of detail in the issue, but in summary:

A change was made recently to allow Sprout to calculate the S-CSCF URI based on the route header that routed the request to Sprout, as that's usually what we want to do. However, when Sprout routes a request via an AS, the route header that routes the request back to Sprout will have Sprout's IP address in it, so the ScscfSproutlet was setting the S-CSCF URI to its IP address. It then sends this to Homestead, which may send it to the HSS on an SAR, which is bad.

This fix is to store the S-CSCF URI on the AsChain when we first create it. Then, when we get the request back from the AS, we correlate that transaction to the AsChain and retrieve the S-CSCF URI from the AsChain. This ensures that the S-CSCF uri that we send to Homestead is the one that we decided on when Sprout received the initial request.

For ASs which retarget the request, we leave the S-CSCF URI as it was originally.

#### Testing:
* I've added a check to an existing UT that failed before I added this code, and now passes.
* `make full_test` passes
* I've live tested and verified that a deployment which hit the issue before now doesn't

### DIFF
--- a/src/ut/scscf_test.cpp
+++ b/src/ut/scscf_test.cpp
@@ -6914,6 +6914,15 @@ TEST_F(SCSCFTest, OriginatingTerminatingAS)
 
   EXPECT_EQ(1, ((SNMP::FakeEventAccumulatorTable*)_scscf_sproutlet->_audio_session_setup_time_tbl)->_count);
   EXPECT_EQ(0, ((SNMP::FakeEventAccumulatorTable*)_scscf_sproutlet->_video_session_setup_time_tbl)->_count);
+
+  // Make sure that we haven't sent a request to homestead with 127.0.0.1 as the domain of the S-CSCF URI
+  bool found_wrong_uri = false;
+  for (FakeHSSConnection::UrlBody body : _hss_connection->_calls)
+  {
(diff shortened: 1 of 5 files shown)


src/ut/scscf_test.cpp
```diff
@@ -6914,6 +6914,15 @@ TEST_F(SCSCFTest, OriginatingTerminatingAS)
 
   EXPECT_EQ(1, ((SNMP::FakeEventAccumulatorTable*)_scscf_sproutlet->_audio_session_setup_time_tbl)->_count);
   EXPECT_EQ(0, ((SNMP::FakeEventAccumulatorTable*)_scscf_sproutlet->_video_session_setup_time_tbl)->_count);
+
+  // Make sure that we haven't sent a request to homestead with 127.0.0.1 as the domain of the S-CSCF URI
```
johadalin: I think it's worth adding a more to the comment here to cover why we're checking this, and what the error case is that would lead to us sending 127.0.0.1 across to Homestead

PR #1935, opened 2017-09-18
[Reviewer: Seb] Av spam

johadalin:
These changes shift where we set the timer for the auth_challenge in the authentication sproutlet, so that we can store off the timer ID, and later delete it once authentication has successfully completed. 
This should stop us getting a lot of timer pop spam, which is then unable to find a related auth_challenge.

Tested live, and it appears to work as expected. Just working on some UTs now, but make full_test still passes with coverage already anyway.


src/authenticationsproutlet.cpp
```diff
@@ -791,14 +791,43 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
...
     std::string nonce = auth_challenge->get_nonce();
 
+    // Create a timer to track the authentication challenge expiry.
+    if ((!impu_for_hss.empty()) && (_authentication->_chronos))
+    {
+      TRC_DEBUG("Set chronos timer for AUTHENTICATION_TIMEOUT SAR");
+
+      // We need to set a Chronos timer so that an AUTHENTICATION_TIMEOUT SAR
+      // is sent to the HSS when the challenge expires. We do this here so
+      // that the timer_id can be stored alongside the auth_challenge.
```
sebrexmetaswitch: May be worth explicitly adding that we don't have the id until we've set the timer, as Chronos generates it for us, which is why we need to do this first

src/authenticationsproutlet.cpp
```diff
@@ -841,8 +848,28 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
       TRC_DEBUG("Failed to store nonce in memcached");
       rsp->line.status.code = PJSIP_SC_INTERNAL_SERVER_ERROR;
       rsp->line.status.reason = *pjsip_get_status_text(PJSIP_SC_INTERNAL_SERVER_ERROR);
+
+      // Also attempt to delete the chronos timer we stored for this challenge.
+      // This stops the a timer pop not finding an AV, and triggering a cycle of
+      // timer pops in every site attempting to find an AV that never existed.
+      // If we failed to create the timer, so have a blank ID, the
+      // chronos_connection will catch this, and not send a delete out.
```
sebrexmetaswitch: I think we should be checking for the id being blank here, if it's expected. If the chronos connection decides not to send the delete, we'll get an ERROR log, which I don't think we want (the error was earlier, not here)

src/authenticationsproutlet.cpp
```diff
@@ -1018,6 +1045,20 @@ void AuthenticationSproutletTsx::on_rx_initial_request(pjsip_msg* req)
         // authentications before it happens.
         auth_challenge->set_nonce_count(nonce_count + 1);
 
+        // The challenge has been authenticated against successfully, so we can
+        // remove the Chronos timer set at creation to trigger expiry
+        if (_authentication->_chronos)
+        {
+          HTTPCode status;
+          status = _authentication->_chronos->send_delete(auth_challenge->get_timer_id(),
```
sebrexmetaswitch: We definitely want to be checking for the timer id being empty here before sending the delete. This is actually a mainline case, if you're authenticating non-REGISTER requests, we will have set the timer id to be blank, but we don't want to log an error

johadalin:
make full_test passes fine, and tested live before last changes, so should all be good

src/ut/authentication_test.cpp
```diff
@@ -2778,6 +2774,151 @@ TEST_F(AuthenticationTest, DigestAuthFailureWithSetError)
...
+TEST_F(AuthenticationTest, AuthSuccessTimerCreationFail)
+{
+  pjsip_tx_data* tdata;
+  // Set up the HSS response for the AV query using a default private user identity.
+  _hss_connection->set_result("/impi/6505550001%40homedomain/av?impu=sip%3A6505550001%40homedomain&server-name=sip%3Ascscf.sprout.homedomain%3A5058%3Btransport%3DTCP",
+                              "{\"digest\":{\"realm\":\"homedomain\",\"qop\":\"auth\",\"ha1\":\"12345678123456781234567812345678\"}}");
+
+  // Set up the timer creation to return an error, simulating timer creation failure.
+  EXPECT_CALL(*_chronos_connection, send_post(_,_,"/authentication-timeout",_,_,_))
+              .WillOnce(DoAll(SetArgReferee<0>("TIMER"),
```
sebrexmetaswitch: I don't think we should be setting the `post_identity` here, as that's not what the real chronos connection does (https://github.com/Metaswitch/cpp-common/blob/master/src/chronosconnection.cpp#L110)

src/ut/authentication_test.cpp
```diff
@@ -2778,6 +2774,151 @@ TEST_F(AuthenticationTest, DigestAuthFailureWithSetError)
...
+// Basic Authentication flow, but checking that we create and delete the timer correctly.
+TEST_F(AuthenticationTest, AuthSuccessTimerDelete)
+{
+  pjsip_tx_data* tdata;
+  // Set up the HSS response for the AV query using a default private user identity.
+  _hss_connection->set_result("/impi/6505550001%40homedomain/av?impu=sip%3A6505550001%40homedomain&server-name=sip%3Ascscf.sprout.homedomain%3A5058%3Btransport%3DTCP",
+                              "{\"digest\":{\"realm\":\"homedomain\",\"qop\":\"auth\",\"ha1\":\"12345678123456781234567812345678\"}}");
+
+  // Check we're sending a timer out. This sets the timer id argument to "TIMER_ID"
+  EXPECT_CALL(*_chronos_connection, send_post(_,_,"/authentication-timeout",_,_,_));
```
sebrexmetaswitch: Does it set the timer ID to "TIMER_ID"? (Hint: no - It doesn't because you've not used `SetArgReferee`.)

I'm pretty sure this means that the `accept_all_requests()` call above is actually being invoked here, so you're not testing your chronos expectations correctly

johadalin:
OK. So this caused more annoyance than i wanted, but should now be in a good state. 
We've extended the templating of the test class, so that we can pass in a class to either set up a FakeChronosConnection or a MockChronosConnection. 

The tests now properly check all the right  conditions. I've had to force verification at the end of each test, as the chronos connection is only killed at the end of all of them, so it said all passed, and then raised errors at the end.
I've tested running these tests against master, and they all correctly fail. They all then pass, and full_test passes too, on this branch.

johadalin:
One last review would probably help :)

src/ut/authentication_test.cpp
```diff
@@ -2828,21 +2890,24 @@ TEST_F(AuthenticationTest, AuthSuccessTimerDelete)
...
   pjsip_tx_data* tdata;
   // Set up the HSS response for the AV query using a default private user identity.
   _hss_connection->set_result("/impi/6505550001%40homedomain/av?impu=sip%3A6505550001%40homedomain&server-name=sip%3Ascscf.sprout.homedomain%3A5058%3Btransport%3DTCP",
                               "{\"digest\":{\"realm\":\"homedomain\",\"qop\":\"auth\",\"ha1\":\"12345678123456781234567812345678\"}}");
 
   // Set up the timer creation to return an error, simulating timer creation failure.
-  EXPECT_CALL(*_chronos_connection, send_post(_,_,"/authentication-timeout",_,_,_))
-              .WillOnce(DoAll(SetArgReferee<0>("TIMER"),
+  EXPECT_CALL(*MockChronosConnectionHelper::get_chronos_connection(), send_post(_,_,"/authentication-timeout",_,_,_))
+              .WillOnce(DoAll(SetArgReferee<0>(""),
```
sebrexmetaswitch: Minor things, but we technically shouldn't be doing `SetArgReferee<0>("")` here, we should just return the error without setting the first argument

PR #1940, opened 2017-09-19
Add a new SAS event to clearly state that an invalid iFC is being skipped

KristaHyer:
A member of the Dev team was confused by the lack of a SAS log, which I have added in now.

They were using a OpenIMSHSS during testing, and this HSS incorrectly set the "ProfilePartIndicator" in one of their iFCs to 2. (The valid options are 0 and 1.)

In our iFC parsing code, we then wouldn't accept this iFC.
We hit an error in parse_integer, since the value of 2 was outside the allowed range from 0-1.
Here is where we parse the node: https://github.com/Metaswitch/sprout/blob/dev/src/ifc.cpp#L451
We throw an error here: https://github.com/Metaswitch/cpp-common/blob/5c4ab31d64b7af1a7e5be6808c9e64ba4de882bd/src/xml_utils.cpp#L87
We catch the error here: https://github.com/Metaswitch/sprout/blob/dev/src/ifc.cpp#L558

We didn't log a SAS event when we caught the error, stating in a comment that one has already been logged.
However, for this specific error this wasn't the case.

The logs usually look like this (for no match, then a match):
`Testing whether an iFC matches`
`The iFC for sip:tas-srv-4.vm7.richlab.datcon.co.uk;transport=TCP;oc-mode=scc does not match: the service triggers evaluated to false`
`Testing whether an iFC matches`
`The iFC for sip:tas-srv-4.vm7.richlab.datcon.co.uk;transport=TCP;oc-mode=mmtel;oc-charge-mode=offline matches`

Instead they looked like this (for the invalid iFC, then no match):
`Testing whether an iFC matches`
`Testing whether an iFC matches`
`The iFC for sip:tas-srv-a-1.vm7.richlab.datcon.co.uk;transport=tcp;oc-mode=mmtel;oc-charge-mode=offline does not match the message's Profile Part Indicator`
We state that we are testing for a match, then don't say if there was/wasn't a match, and move on.
This was confusing.

I've corrected this by adding a SAS event to state that the invalid iFC is being skipped, and updating the comment.

Make full_test passes.


KristaHyer:
I discussed the option of adding a SAS event into the parse_integer code as well, to give the details of why the iFC was invalid.
However, when making this change, it started to feel like overkill.
I think that for this specific error a SAS event stating that the iFC was invalid should be enough.
(Especially since an error log is logged in this case, so we do note down the error, just not in a SAS event.)

src/ifc.cpp
```diff
@@ -557,10 +557,13 @@ bool Ifc::filter_matches(const SessionCase& session_case,
...
   catch (xml_error err)
   {
-    // Ignore individual criteria which can't be parsed. SAS logging
-    // should already have happened by this point.
+    // Generic SAS event is logged to make it clear that this iFC is being
+    // skipped since it is invalid. In most cases, a specific SAS event
+    // detailing the exact error will already have been logged as well.
     std::string err_str = "iFC evaluation error: " + std::string(err.what());
     TRC_ERROR(err_str.c_str());
+    SAS::Event event(trail, SASEvent::INVALID_IFC_IGNORED, 0);
```
eleanor-merry: Can we put the error in the SAS event as well?

KristaHyer: This is such a good idea!

PR #1944, opened 2017-09-27
Notify on deregister

Ying-metaswitch:
Code and UT complete. Live testing not done yet.


src/subscriber_data_manager.cpp
```diff
@@ -561,28 +562,30 @@ void SubscriberDataManager::NotifySender::send_notifys(
                                const std::string& aor_id,
                                AoRPair* aor_pair,
                                int now,
+                               const bool admin_dereg,
                                SAS::TrailId trail)
 {
-  std::vector<std::string> expired_binding_uris;
+  std::vector<std::string> deleted_binding_uris;
```
eleanor-merry: I'm not sure I like the word deleted - I would go for missing, and bring up the comment from line 771 to explain it.

src/subscriber_data_manager.cpp
```diff
@@ -763,9 +768,12 @@ void SubscriberDataManager::NotifySender::send_notifys_for_expired_subscriptions
     NotifyUtils::RegistrationState::ACTIVE :
     NotifyUtils::RegistrationState::TERMINATED;
 
-  // expired_binding_uris lists bindings which have expired - we no longer have a valid connection to
-  // these endpoints, so shouldn't send a NOTIFY to them (even to say that their subscription is
-  // terminated).
+  // Deleted_binding_uris lists bindings which no longer exists in AoR.
```
eleanor-merry: No capital in deleted (as it's describing a variable that doesn't have a capital).
Also, *exist in the AoR.

src/subscriber_data_manager.cpp
```diff
@@ -763,9 +768,12 @@ void SubscriberDataManager::NotifySender::send_notifys_for_expired_subscriptions
     NotifyUtils::RegistrationState::ACTIVE :
     NotifyUtils::RegistrationState::TERMINATED;
 
-  // expired_binding_uris lists bindings which have expired - we no longer have a valid connection to
-  // these endpoints, so shouldn't send a NOTIFY to them (even to say that their subscription is
-  // terminated).
+  // Deleted_binding_uris lists bindings which no longer exists in AoR. 
+  // They may be deleted by administrative deregistration from Sprout/HSS, and
```
eleanor-merry: *have been. Also, I don't think you need the 'from Sprout/HSS' either. Also, removed rather than deleted?

src/subscriber_data_manager.cpp
```diff
@@ -763,9 +768,12 @@ void SubscriberDataManager::NotifySender::send_notifys_for_expired_subscriptions
     NotifyUtils::RegistrationState::ACTIVE :
     NotifyUtils::RegistrationState::TERMINATED;
 
-  // expired_binding_uris lists bindings which have expired - we no longer have a valid connection to
-  // these endpoints, so shouldn't send a NOTIFY to them (even to say that their subscription is
-  // terminated).
+  // Deleted_binding_uris lists bindings which no longer exists in AoR. 
+  // They may be deleted by administrative deregistration from Sprout/HSS, and 
+  // corresponding endpoints need to be NOTIFYed of their termination.
+  // Or they may be deleted by endpoint de-register or Chronos expiry, and we
```
eleanor-merry: They may have been removed because the endpoint expired or deregistered, and we..

src/ut/chronoshandlers_test.cpp
```diff
@@ -82,15 +82,16 @@ TEST_F(ChronosAoRTimeoutTasksTest, MainlineTest)
            .WillOnce(DoAll(SetArgReferee<3>(AssociatedURIs(associated_uris)), //IMPUs in IRS
                            Return(HTTP_OK)));
       EXPECT_CALL(*store, get_aor_data(aor_id, _)).WillOnce(Return(aor));
-      EXPECT_CALL(*store, set_aor_data(aor_id, aor, _, _)).WillOnce(Return(Store::OK));
+      bool admin_dereg = false;
+      EXPECT_CALL(*store, set_aor_data(aor_id, aor, _, _, admin_dereg)).WillOnce(Return(Store::OK));
```
eleanor-merry: Just put false directly into the expect call (or the new ENUM value). Same is true for the other tests.

src/ut/mock_subscriber_data_manager.h
```diff
@@ -23,10 +23,11 @@ class MockSubscriberDataManager : public SubscriberDataManager
 
   MOCK_METHOD2(get_aor_data, AoRPair*(const std::string& aor_id,
                                       SAS::TrailId trail));
-  MOCK_METHOD4(set_aor_data, Store::Status(const std::string& aor_id,
+  MOCK_METHOD5(set_aor_data, Store::Status(const std::string& aor_id,
                                            AoRPair* data,
                                            SAS::TrailId trail,
-                                           bool& all_bindings_expired));
+                                           bool& all_bindings_expired,
+                                           bool admin_dereg));
```
eleanor-merry: This should be const surely?

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -54,7 +55,7 @@ class BasicSubscriberDataManagerTest : public SipTest
     _chronos_connection = new FakeChronosConnection();
     _datastore = new LocalStore();
     _aor_store = new AstaireAoRStore(_datastore);
-    _analytics_logger = new MockAnalyticsLogger();
+    _analytics_logger = new NiceMock<MockAnalyticsLogger>();
```
eleanor-merry: Why did you need this change?

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -445,6 +446,141 @@ TEST_F(BasicSubscriberDataManagerTest, AssociatedURIsTests)
   delete aor_data1; aor_data1 = NULL;
 }
 
+// Use log to check that Admin Deregistration will send a final NOTIFY, while
```
eleanor-merry: *Test that if a binding is removed due to administrative deregistration then a NOTIFY is sent, and if it's removed because it's expired then a NOTIFY isn't sent.

(You want the more detailed comments within the UT).

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -445,6 +446,141 @@ TEST_F(BasicSubscriberDataManagerTest, AssociatedURIsTests)
...
+  CapturingTestLogger log;
+  AoRPair* aor_data1;
+  AoR::Binding* b0;
+  AoR::Binding* b1;
+  AoR::Binding* b2;
+  AoR::Subscription* s0;
+  AoR::Subscription* s1;
+  AssociatedURIs associated_uris = {};
+  bool rc;
+  int now;
```
eleanor-merry: You don't need to define these upfront surely - define them when you use them (e.g. int now = time(NULL)).

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -445,6 +446,141 @@ TEST_F(BasicSubscriberDataManagerTest, AssociatedURIsTests)
...
+
+  // Get an initial empty AoR record.
+  now = time(NULL);
+  aor_data1 = this->_store->get_aor_data(std::string("5102175691@cw-ngv.com"), 0);
+  ASSERT_TRUE(aor_data1 != NULL);
+  aor_data1->get_current()->_timer_id = "AoRtimer";
+  aor_data1->get_current()->_associated_uris = associated_uris;
+  EXPECT_EQ(0u, aor_data1->get_current()->bindings().size());
+
+  // Add a binding and corresponding subscription. URI has to be exact match as the code assumes so.
```
eleanor-merry: *The URI is the same for the binding and the subscription as they're made from the same endpoint.

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -445,6 +446,141 @@ TEST_F(BasicSubscriberDataManagerTest, AssociatedURIsTests)
...
+  b1->_private_id = "5102175691@cw-ngv.com";
+  b1->_emergency_registration = false;
+
+  s1 = aor_data1->get_current()->get_subscription("1234");
+  s1->_req_uri = std::string("<sip:5102175691@192.91.191.29:59934;transport=tcp;ob>");
+  s1->_from_uri = std::string("<sip:5102175691@cw-ngv.com>");
+  s1->_from_tag = std::string("4321");
+  s1->_to_uri = std::string("<sip:5102175691@cw-ngv.com>");
+  s1->_to_tag = std::string("1234");
+  s1->_cid = std::string("xyzabc@192.91.191.29");
```
eleanor-merry: Why is b1/s1 before b0/s0?

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -445,6 +446,141 @@ TEST_F(BasicSubscriberDataManagerTest, AssociatedURIsTests)
...
+  s0->_req_uri = std::string("sip:5678@5678");
+  s0->_from_uri = std::string("<sip:5102175691@cw-ngv.com>");
+  s0->_from_tag = std::string("4321");
+  s0->_to_uri = std::string("<sip:5102175691@cw-ngv.com>");
+  s0->_to_tag = std::string("5678");
+  s0->_cid = std::string("xyzabc@192.91.191.29");
+  s0->_route_uris.push_back(std::string("<sip:abcdefgh@bono-1.cw-ngv.com;lr>"));
+  s0->_expires = now + 300;
+
+  // Add another binding so that when the other twos are removed, the AoR is not
```
eleanor-merry: *two

src/ut/subscriber_data_manager_test.cpp
```diff
@@ -445,6 +446,141 @@ TEST_F(BasicSubscriberDataManagerTest, AssociatedURIsTests)
...
+  b2->_priority = 0;
+  b2->_path_uris.push_back(std::string("sip:abcdefgh@bono-1.cw-ngv.com;lr"));
+  b2->_path_headers.push_back(std::string("\"Bob\" <sip:leftbinding@bono-1.cw-ngv.com;lr>;tag=6ht7"));
+  b2->_params["+sip.instance"] = "\"<urn:uuid:111111-0000-0000-0000-b4dd32817622>\"";
+  b2->_params["reg-id"] = "1";
+  b2->_params["+sip.ice"] = "";
+  b2->_private_id = "111111@cw-ngv.com";
+  b2->_emergency_registration = false;
+
+  // Write AoR record back to store.
```
eleanor-merry: At this point you want to mention that you're looking at the logs to detect NOTIFYs being sent (and how that's quite fragile).

src/ut/subscriber_data_manager_test.cpp
Ying-metaswitch: To suppress the console from printing "uninteresting function call: xxxxx" as I thought it messes up the error message

src/handlers.cpp
```diff
@@ -574,9 +580,11 @@ AoRPair* DeregistrationTask::deregister_bindings(
 
     aor_pair->get_current()->_associated_uris = associated_uris;
     set_rc = current_sdm->set_aor_data(aor_id,
+                                       SubscriberDataManager::EventTrigger::USER,
```
eleanor-merry: Should be ADMIN

src/subscriber_data_manager.cpp
```diff
@@ -576,13 +578,14 @@ void SubscriberDataManager::NotifySender::send_notifys(
...
+    // Compare the original and current lists to see whether this binding has
+    // been deleted.
     // Emergency bindings are excluded from notifications.
     if ((!binding->_emergency_registration) &&
         (aor_pair->get_current()->bindings().find(b_id) == aor_pair->get_current()->bindings().end()))
     {
-      TRC_DEBUG("Binding %s has expired", b_id.c_str());
-      expired_binding_uris.push_back(binding->_uri);
+      TRC_DEBUG("Binding %s has been expired", b_id.c_str());
+      missing_binding_uris.push_back(binding->_uri);
```
eleanor-merry: You've not changed the ContactEvent yet - this needs to be investigated before this can be checked in.

src/subscriber_data_manager.cpp
```diff
@@ -559,11 +560,12 @@ SubscriberDataManager::NotifySender::~NotifySender()
 
 void SubscriberDataManager::NotifySender::send_notifys(
                                const std::string& aor_id,
+                               const SubscriberDataManager::EventTrigger& event_trigger,
                                AoRPair* aor_pair,
                                int now,
                                SAS::TrailId trail)
 {
-  std::vector<std::string> expired_binding_uris;
+  std::vector<std::string> missing_binding_uris;
```
eleanor-merry: You should move the relevant comment up here (see my previous feedback)

src/subscriber_data_manager.cpp
```diff
@@ -763,9 +768,12 @@ void SubscriberDataManager::NotifySender::send_notifys_for_expired_subscriptions
     NotifyUtils::RegistrationState::ACTIVE :
     NotifyUtils::RegistrationState::TERMINATED;
 
-  // expired_binding_uris lists bindings which have expired - we no longer have a valid connection to
-  // these endpoints, so shouldn't send a NOTIFY to them (even to say that their subscription is
-  // terminated).
+  // deleted_binding_uris lists bindings which no longer exist in AoR.
```
eleanor-merry: Wrong name

src/ut/chronoshandlers_test.cpp
```diff
@@ -82,15 +82,15 @@ TEST_F(ChronosAoRTimeoutTasksTest, MainlineTest)
            .WillOnce(DoAll(SetArgReferee<3>(AssociatedURIs(associated_uris)), //IMPUs in IRS
                            Return(HTTP_OK)));
       EXPECT_CALL(*store, get_aor_data(aor_id, _)).WillOnce(Return(aor));
-      EXPECT_CALL(*store, set_aor_data(aor_id, aor, _, _)).WillOnce(Return(Store::OK));
+      EXPECT_CALL(*store, set_aor_data(aor_id, _, aor, _, _)).WillOnce(Return(Store::OK));
```
eleanor-merry: You should have at least one check that the right event trigger is being used.

src/ut/subscriber_data_manager_test.cpp
eleanor-merry: Actually - can you not put this change in this review - it's not really part of the work.

src/ut/subscriber_data_manager_test.cpp
eleanor-merry: ^ See above comment

src/ut/subscription_test.cpp
```diff
@@ -126,7 +126,7 @@ class SubscriptionTest : public SipTest
     std::string aor = "sip:6505550231@homedomain";
     AssociatedURIs associated_uris = {};
     associated_uris.add_uri(aor, false);
-    _sdm->set_aor_data(aor, aor_pair, 0);
+    _sdm->set_aor_data(aor, SubscriberDataManager::EventTrigger::TIMEOUT, aor_pair, 0);
```
eleanor-merry: These should be USER

include/subscriber_data_manager.h
```diff
@@ -128,8 +129,9 @@ class SubscriberDataManager
                                    const std::string& aor_id,
                                    AoRPair* aor_pair,
                                    ClassifiedBindings binding_info_to_notify,
-                                   std::vector<std::string> expired_binding_uris,
+                                   std::vector<std::string> deleted_binding_uris,
                                    int now,
+                                   const bool admin_dereg,
```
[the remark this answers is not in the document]
Ying-metaswitch: It will be SDM::EventTrigger {USER, ADMIN, TIMEOUT}

1.	DeregistrationTask & Registrar & SubcriptionSproutlet – USER
2.	DeleteImpuTask & PushProfileTask & Registration_Utils – ADMIN
3.	AoRTimeout & SCSCFSproutlet – TIMEOUT
In UTs:
4.	SCSCFTest -- USER

Ying-metaswitch:
Add in new NotifyUtils::ContactEvent for unregistered (user deregistration) and deactivated (admin deregistration). See https://tools.ietf.org/html/rfc3680 for detail. 

`make full_test` passed.

PR #1945, opened 2017-09-27
Sprout infrastructure changes to support MMF at network function boundary

huws:
These are the changes to allow MMF to be invoked at network function boundaries.

Some notes:
 - I ended up having to pass the depth and network function parameters through the PendingRequests queue.  This really isn't what I wanted (nor what I initially implemented), but the ACK-processing behaviour of the SproutletProxy meant that I couldn't reliably use pointers to upstream sproutlets.  On receipt of an ACK request, the SproutletWrapper deletes itself before the downstream SproutletWrapper is notified of the request.  I believe that this is necessary, as we won't get a later kick to tidy up this transaction.
 - I had to move the definition of `Sproutlet` up in its file, so that I could use it within the `SproutletTsx::get_network_function()` method.  I could have defined the method in a .cpp file, but there isn't currently one for `sproutlet.h`.  Alternatively, I could have added the method definition outside the class definition, but still in `sproutlet.h`, but that felt confusing, too.
 - Git doesn't seem to have tracked my rename of `forwardingsproutlet.*` to `compositesproutlet.*` - it shows as a delete and and add.
 - `make full_test` passes, but I haven't run the live tests yet.


huws:
I've just seen Alex's note saying that he may not have time for the code-review on Thursday.  Re-assigning to you, Krista.  Give me a call if you're missing any context.

include/sproutletproxy.h
```diff
@@ -220,6 +225,8 @@ class SproutletProxy : public BasicProxy, SproutletHelper
       pjsip_tx_data* req;
       std::pair<SproutletWrapper*, int> upstream;
       int allowed_host_state;
+      int sproutlet_depth;
+      std::string upstream_nf;
```
KristaHyer: I want to rename all "nf" to "network_func", since it's unclear what nf means.
However, I don't feel very strongly, so you can make this choice.

include/sproutletproxy.h
```diff
@@ -354,6 +368,18 @@ class SproutletWrapper : public SproutletTsxHelper
   // Immutable reference to the transport used by the original request.
   pjsip_transport* _original_transport;
 
+  // The name of the Network function of this Sproutlet.
```
KristaHyer: Sorry picky comment.
Your capitalization of Network Function is different here 😛 
Should we also give the default here (or is the other place enough?)

include/sproutletproxy.h
```diff
@@ -354,6 +368,18 @@ class SproutletWrapper : public SproutletTsxHelper
...
+  std::string _this_nf;
+
+  // The name of the Network Function of the upstream Sproutlet (if any).
+  // This is used to detect transitions between Network Functions, so that we
+  // can perform SIP-entity-level operations like sending 100 Trying responses
+  // and decrementing the Max-Forwards counter.
```
KristaHyer: I love this comment! Good explanation.

include/sproutletproxy.h
```diff
@@ -354,6 +368,18 @@ class SproutletWrapper : public SproutletTsxHelper
...
+  std::string _upstream_nf;
+
+  // The depth of this wrapper in the transaction tree.  Used to detect loops.
```
KristaHyer: another good comment!
(assuming people know what depth means, I still don't know if it's just me who didn't know what it was before we started this work)

src/Makefile
```diff
@@ -67,6 +67,7 @@ SPROUT_COMMON_SOURCES := logger.cpp \
                          chronoshandlers.cpp \
                          contact_filtering.cpp \
                          sproutletproxy.cpp \
+                         compositesproutlet.cpp \
```
KristaHyer: Why is this now here, when forwardingsproutlet.cpp wasn't?
Is it because you removed it from "sprout_scscf.so_SOURCES"?
Why this change?

src/icscfsproutlet.cpp
```diff
@@ -38,15 +38,19 @@ ICSCFSproutlet::ICSCFSproutlet(const std::string& icscf_name,
...
                                HSSConnection* hss,
                                ACRFactory* acr_factory,
                                SCSCFSelector* scscf_selector,
                                EnumService* enum_service,
                                SNMP::SuccessFailCountByRequestTypeTable* incoming_sip_transactions_tbl,
                                SNMP::SuccessFailCountByRequestTypeTable* outgoing_sip_transactions_tbl,
                                bool override_npdi) :
-  Sproutlet(icscf_name, port, uri, "", {}, incoming_sip_transactions_tbl, outgoing_sip_transactions_tbl),
+  Sproutlet(icscf_name, port, uri, "", {}, incoming_sip_transactions_tbl,
+            outgoing_sip_transactions_tbl, network_function),
```
KristaHyer: Picky again
This is just a bit painful to look at - can the variables all be on their own line?

src/scscfsproutlet.cpp
```diff
@@ -54,14 +56,16 @@ SCSCFSproutlet::SCSCFSproutlet(const std::string& name,
                                int session_terminated_timeout_ms,
                                AsCommunicationTracker* sess_term_as_tracker,
                                AsCommunicationTracker* sess_cont_as_tracker) :
-  Sproutlet(name, port, uri, "", {}, incoming_sip_transactions_tbl, outgoing_sip_transactions_tbl),
+  Sproutlet(name, port, uri, "", {}, incoming_sip_transactions_tbl,
+            outgoing_sip_transactions_tbl, network_function),
```
KristaHyer: Same comment as above - I'd like it better split out on separate lines

src/sproutletproxy.cpp
```diff
@@ -872,11 +878,25 @@ void SproutletProxy::UASTsx::schedule_requests()
...
+    if ((mf_hdr != NULL) && (mf_hdr->ivalue <= 0))
+    {
+      // Max-Forwards has decayed to zero - we've detected a loop.
+      TRC_DEBUG("Max-Forwards too low");
+      loop_detected = true;
+    }
+    else if (req.sproutlet_depth <= 0)
+    {
+      // Maximum recursion depth for Sproutlets reached - it's a loop.
+      TRC_ERROR("Exceeded maximum Sproutlet tree depth - coding error");
```
KristaHyer: Can we say this in a customer facing log?
I don't doubt it's true, but can we phrase it so that they don't instantly see it's our mistake?

src/sproutletproxy.cpp
```diff
@@ -872,11 +878,25 @@ void SproutletProxy::UASTsx::schedule_requests()
...
+      loop_detected = true;
+    }
+
+    if (loop_detected)
     {
-      // Max-Forwards has decayed to zero, so either reject the request or
-      // discard it if it's an ACK.
+      // We've detected a loop, so either reject the request or discard it if
+      // it's an ACK.
```
KristaHyer: Overall this whole section is clear, and nice to read.

src/sproutletproxy.cpp
```diff
@@ -2074,16 +2111,24 @@ void SproutletWrapper::aggregate_response(pjsip_tx_data* rsp)
...
+      deregister_tdata(rsp);
+      pjsip_tx_data_dec_ref(rsp);
+    }
+    else
+    {
+      // This Sproutlet does not automatically send 100 Trying responses.  Pass
+      // this one through now.
+      TRC_DEBUG("Forward 100 Trying response");
+      tx_response(rsp);
+    }
```
KristaHyer: Also very clear!

src/ut/icscfsproutlet_test.cpp
```diff
@@ -214,7 +216,7 @@ using TestingCommon::Message;
 
 TEST_F(ICSCFSproutletTest, RouteRegisterHSSServerName)
 {
-  // Tests routing of REGISTER requests when the HSS responses with a server
+  // Tests routing of REGISTER requests when the HSS responds with a server
```
KristaHyer: Good spot!

src/ut/sproutletproxy_test.cpp
```diff
@@ -41,15 +42,20 @@ class FakeSproutlet : public Sproutlet
                 const std::string& service_host,
                 std::string alias = "",
                 SNMP::FakeSuccessFailCountByRequestTypeTable* fake_inc_tbl = NULL,
-                SNMP::FakeSuccessFailCountByRequestTypeTable* fake_out_tbl = NULL) :
-    Sproutlet(service_name, port, uri, service_host, { alias }, fake_inc_tbl, fake_out_tbl)
+                SNMP::FakeSuccessFailCountByRequestTypeTable* fake_out_tbl = NULL,
+                const std::string& network_function="",
+                const std::string& next_hop="") :
+    Sproutlet(service_name, port, uri, service_host, { alias }, fake_inc_tbl, fake_out_tbl, network_function),
+    _next_hop(next_hop)
```
KristaHyer: Alignment issue?

src/ut/sproutletproxy_test.cpp
```diff
@@ -2096,9 +2115,95 @@ TEST_F(SproutletProxyTest, SproutletChain)
...
+  // All done!
+  ASSERT_EQ(0, txdata_count());
+
+  delete tp;
+}
+
+TEST_F(SproutletProxyTest, LoopDetectionMaxForwards)
 {
   // Test loop detection of requests passing through a chain of sproutlets.
+  // We use two sproutlets that return different network function namse,
```
KristaHyer: typo

src/ut/sproutletproxy_test.cpp
```diff
@@ -2096,9 +2115,95 @@ TEST_F(SproutletProxyTest, SproutletChain)
...
+  EXPECT_EQ("sip:bob@proxy1.awaydomain:5060;transport=TCP", str_uri(tdata->msg->line.req.uri));
+  EXPECT_EQ("", get_headers(tdata->msg, "Route"));
+  EXPECT_EQ("", get_headers(tdata->msg, "Record-Route"));
+  free_txdata();
+
+  // All done!
+  ASSERT_EQ(0, txdata_count());
+
+  delete tp;
+}
```
KristaHyer: I love this whole test - it's very clear

huws:
FYI - I've now run the Live Tests, and fixed a typo that caused a loop when invoking the I-CSCF! (a loop detected by the new max_sproutlet_depth code - phew)

On the UTs for the 100 Trying stuff:
 - This behaviour is hidden away in the interactions between Sproutlets.  The existing tests check that the first Sproutlet returns a 100 Trying, but are not party to whether 100 Tryings are sent between chained Sproutlets.
 - The composite sproutlet test that I've added exercises the 100 Trying pass-through behaviour (which is how this change maintains 100% code-coverage for `sproutletproxy.cpp`).
 - It might be good to have a UT covering the explicit case of the `scscf-proxy` sending an INVITE to an unresponsive application server.  I'll see if I can easily add such a test.

KristaHyer:
Sounds good to me!

include/compositesproutlet.h
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef COMPOSITESPROUTLET_H__
+#define COMPOSITESPROUTLET_H__
+
+#include "sproutlet.h"
+
+///
+/// A Sproutlet TSX that is part of a more complicated Network Function.  When
+/// it sends outbound standalone or dialog-creating requests, they are routed
+/// to the next Sproutlet in the Network Function.
```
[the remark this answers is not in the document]
huws: In-dialog messages are routed according to standard SIP routing - i.e. the Record-Route headers are used.  Comment updated.

include/sproutletproxy.h
huws: Done

src/Makefile
huws: Previously only Sproutlets within the S-CSCF extended ForwardingSproutletTsx. Now we want to reuse it between Sproutlets in multiple network functions, so we have to move it up to be built as part of Sprout.

src/sproutletproxy.cpp
huws: You're right - I'll tone it down.

src/ut/sproutletproxy_test.cpp
huws: No - I think that's right.  We've got two initializers, one is a call to the superclass's constructor, the second of which initializes a member variable.  They should, and do, start at the same column, indented from the constructor name.

include/cfgoptions.h
```diff
@@ -119,6 +119,7 @@ struct options
   int                                  session_continued_timeout_ms;
   int                                  session_terminated_timeout_ms;
   std::set<std::string>                stateless_proxies;
+  int                                  max_sproutlet_depth;
```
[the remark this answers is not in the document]
huws: Recall that we are not counting the number sproutlets invoked, we're counting how deep we have gone in the tree of sproutlets.  So it would be possible to invoke more than 50 (max depth) sproutlets in total while handling a single inbound request, provided the tree of sproutlets that we create never grows beyond 50 sproutlets on a single branch.

I think I'm happy with the current naming, but shout if you think there's a better way to say it.

src/Makefile
KristaHyer: Oh ok got it :)

include/cfgoptions.h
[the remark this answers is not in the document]
KristaHyer: After your explanation, I'm happy to leave this :)

PR #1948, opened 2017-09-29
[Reviewer: Adam] Use target-specific exclusions

sebrexmetaswitch:
Use the new feature of the makefile in https://github.com/Metaswitch/clearwater-build-infra/pull/70 to exlcude all the Bono tests unless the variable `SLOW` is set. This allows us to speed up the valgrind check for most use-cases. I'm expecting that most developers will want to run without SLOW set, but that our automated builds will run with it set.

So, to run valgrind over everything, you now run

`SLOW=T make valgrind_check`

like we have in etcd.

An example of the speed-up I saw (on the repo server, so not one where we have to drastic slowdown that we've seen in some cases):

```
time make valgrind_check
...
real    4m2.077s
user    3m53.094s
sys     0m1.847s
```
vs
```
time SLOW=T make valgrind_check
...
real    5m50.543s
user    5m41.626s
sys     0m1.866s
```

That's a 2-minute saving on a 5-minute build, which is pretty decent.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -317,6 +317,9 @@ VPATH = ../modules/cpp-common/src ../modules/cpp-common/test_utils ../modules/ap
 # Use valgrind suppression file for UT
 sprout_test_VALGRIND_ARGS := --suppressions=ut/sprout_test.supp
 
+# Exclude the Bono tests from valgrind unless SLOW is set
+sprout_test_VALGRIND_EXCL = $(if ${SLOW},,Stateful*Proxy*Test.*)
+
 include ../build-infra/cpp.mk
 
 # Special extra objects for sprout_test


johadalin:
Looks sensible. I'm slightly worried that we're just going to lose track of this, and end up not running the tests ever. I assume our build servers are going to keep running with the SLOW flag, but is it better to have this as an opt-in flag, rather than opt-out? 
Seems like a tough trade off of 'people might just forget to set the flag and get annoyed at test times' and  'people might change the code, and then not end up testing it properly'.
Equally, i don't think we're going to end up changing this code much, so it's probably fine.

Assuming that it's all tested and works, given the output you've included, so if you're happy with my concern here, shipit :ship:

sebrexmetaswitch:
Just for lols, decided to check what difference this made on my local VM. Timings:

```
time make valgrind_check
...
real    3m27.660s
user    3m17.408s
sys     0m2.084s
```
```
time SLOW=T make valgrind_check
...
real    38m39.461s
user    38m26.560s
sys     0m5.676s
```
That's a pretty sweet 30+ minute saving

sebrexmetaswitch:
As for your comments: I think the right thing to do here is default to skipping them. I'll send people a PSA announcing that I've done this, and if anyone complains I can look at defaulting it the other way.

sebrexmetaswitch:
(but we do the same for etcd, and only run the slow tests if you explicitly say you want to)

PR #1949, opened 2017-10-04
[Reviewer: Tom] Reject on queue

eleanor-merry:
This PR ages out old requests if they've been on the queue for too long; rejecting them in the same way that we reject on overload.

It's not been tested, and I've not thought about what a sensible default would be.

### DIFF
--- a/include/cfgoptions.h
+++ b/include/cfgoptions.h
@@ -147,6 +147,7 @@ struct options
   bool                                 reject_if_no_matching_ifcs;
   std::string                          dummy_app_server;
   bool                                 http_acr_logging;
+  int                                  request_on_queue_timeout;
 };
 
 // Objects that must be shared with dynamically linked sproutlets must be
--- a/sprout-base.root/etc/init.d/sprout
+++ b/sprout-base.root/etc/init.d/sprout
@@ -157,6 +157,7 @@ get_daemon_args()
         [ -z "$chronos_hostname" ] || chronos_hostname_arg="--chronos-hostname=$chronos_hostname"
         [ -z "$sprout_chronos_callback_uri" ] || sprout_chronos_callback_uri_arg="--sprout-chronos-callback-uri=$sprout_chronos_callback_uri"
         [ -z "$dummy_app_server" ] || dummy_app_server_arg="--dummy-app-server=$dummy_app_server"
+        [ -z "$request_on_queue_timeout" ] || request_on_queue_timeout_arg="--request-on-queue-timeout=$request_on_queue_timeout"
 
         DAEMON_ARGS="
                      --domain=$home_domain
@@ -200,6 +201,7 @@ get_daemon_args()
                      $override_npdi_arg
                      $exception_max_ttl_arg
                      $force_3pr_body_arg
+                     $request_on_queue_timeout_arg
                      --http-address=$local_ip
                      --http-port=9888
                      --analytics=$log_directory
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -139,6 +139,7 @@ enum OptionTypes
   OPT_REJECT_IF_NO_MATCHING_IFCS,
   OPT_DUMMY_APP_SERVER,
   OPT_HTTP_ACR_LOGGING,
+  OPT_REQUEST_ON_QUEUE_TIMEOUT
 };
 
 
@@ -229,6 +230,7 @@ const static struct pj_getopt_option long_opt[] =
   { "reject-if-no-matching-ifcs",   no_argument,       0, OPT_REJECT_IF_NO_MATCHING_IFCS},
   { "dummy-app-server",             required_argument, 0, OPT_DUMMY_APP_SERVER},
   { "http-acr-logging",             no_argument,       0, OPT_HTTP_ACR_LOGGING},
+  { "request-on-queue-timeout",     required_argument, 0, OPT_REQUEST_ON_QUEUE_TIMEOUT},
   { NULL,                           0,                 0, 0}
 };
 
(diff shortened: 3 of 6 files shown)


include/cfgoptions.h
```diff
@@ -147,6 +147,7 @@ struct options
   bool                                 reject_if_no_matching_ifcs;
   std::string                          dummy_app_server;
   bool                                 http_acr_logging;
+  int                                  request_on_queue_timeout;
```
tomwatson1024: It's consistent with the other options to omit the unit (_ms) here, but it probably would make it clearer. Not sure whether it's better to change it or not.

src/main.cpp
```diff
@@ -1588,6 +1598,7 @@ int main(int argc, char* argv[])
   opt.reject_if_no_matching_ifcs = false;
   opt.dummy_app_server = "";
   opt.http_acr_logging = false;
+  opt.request_on_queue_timeout = 4000; // TODO decide on actual default
```
tomwatson1024: We agreed in the whiteboard that this needs some consideration once we've got a way to test overload - I'm just commenting so we remember!

src/thread_dispatcher.cpp
```diff
@@ -90,6 +90,7 @@ static SNMP::EventAccumulatorByScopeTable* latency_table = NULL;
 static LoadMonitor* load_monitor = NULL;
 static SNMP::EventAccumulatorByScopeTable* queue_size_table = NULL;
 static ExceptionHandler* exception_handler = NULL;
+static unsigned long request_on_queue_timeout_us = 1;
```
tomwatson1024: Why does this default to 1 over any other value? (more curiosity than anything else)

sprout-base.root/etc/init.d/sprout
```diff
@@ -157,6 +157,7 @@ get_daemon_args()
         [ -z "$chronos_hostname" ] || chronos_hostname_arg="--chronos-hostname=$chronos_hostname"
         [ -z "$sprout_chronos_callback_uri" ] || sprout_chronos_callback_uri_arg="--sprout-chronos-callback-uri=$sprout_chronos_callback_uri"
         [ -z "$dummy_app_server" ] || dummy_app_server_arg="--dummy-app-server=$dummy_app_server"
+        [ -z "$request_on_queue_timeout" ] || request_on_queue_timeout_arg="--request-on-queue-timeout=$request_on_queue_timeout"
```
AlexHockey: It's easier to add the new options to the block around line 286.

src/main.cpp
```diff
@@ -1309,6 +1311,14 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
       }
       break;
 
+    case OPT_REQUEST_ON_QUEUE_TIMEOUT:
+      {
+        VALIDATE_INT_PARAM(options->request_on_queue_timeout,
```
AlexHockey: This can't be zero or negative. Is there a better validation macro to use?

src/thread_dispatcher.cpp
AlexHockey: Does it need a default. Isn't the timeout a mandatory parameter to `init_thread_dispatcher`?

src/thread_dispatcher.cpp
```diff
@@ -144,67 +145,112 @@ static int worker_thread(void* p)
...
-          {
-            TRC_ERROR("Exception Call-Id: %.*s (maybe)",
-                      ((pjsip_cid_hdr*)rdata->msg_info.cid)->id.slen,
-                      ((pjsip_cid_hdr*)rdata->msg_info.cid)->id.ptr);
-          }
-          if (rdata->msg_info.cseq != NULL)
+          TRC_ERROR("Failed to get timestamp: %s", strerror(errno));
+        }
+
+        if ((latency_us > (request_on_queue_timeout_us)) &&
```
AlexHockey: Odd bracketing of `request_on_queue_timeout_us`

src/thread_dispatcher.cpp
```diff
@@ -144,67 +145,112 @@ static int worker_thread(void* p)
...
+            SAS::Marker end_marker(trail, MARKER_ID_END, 2u);
+            SAS::report_marker(end_marker);
 
-          if ((rdata->msg_info.msg->type == PJSIP_REQUEST_MSG) &&
-             (rdata->msg_info.msg->line.req.method.id != PJSIP_ACK_METHOD))
-          {
-            TRC_DEBUG("Returning 500 response following exception");
             pjsip_retry_after_hdr* retry_after =
-                           pjsip_retry_after_hdr_create(rdata->tp_info.pool, 600);
+                           pjsip_retry_after_hdr_create(rdata->tp_info.pool, 0);
```
AlexHockey: Why is 0 the right value to use for the Retry-After header? Regardless, this is probably worth a comment.

src/thread_dispatcher.cpp
```diff
@@ -144,67 +145,112 @@ static int worker_thread(void* p)
...
+                                       NULL);
+            }
+
+            if (num_worker_threads == 1)
+            {
+              // There's only one worker thread, so we can't sensibly proceed.
+              exit(1);
+            }
+          }
+          CW_END
```
AlexHockey: I've not the reviewed the catch block as it looked like it was just moved. Shout if this is wrong.

src/thread_dispatcher.cpp
```diff
@@ -144,67 +145,112 @@ static int worker_thread(void* p)
...
-          load_monitor->request_complete(latency_us);
-        }
-        else
-        {
-          TRC_ERROR("Failed to get done timestamp: %s", strerror(errno));
+          if (me->stop_watch.read(latency_us))
+          {
+            TRC_DEBUG("Request latency = %ldus", latency_us);
+            latency_table->accumulate(latency_us);
+            load_monitor->request_complete(latency_us);
```
AlexHockey: I think this means that the latency of requests that are rejected for having been on the queue for too long is included in this stat. I honestly don't know if that's right or not. Thoughts?

tomwatson1024:
I'm going to go ahead and merge this into overload-integration, but any markups can be made on that branch.

PR #1950, opened 2017-10-04
[Reviewer: Adam] Allow parallel make

sebrexmetaswitch:
There are some changes here designed to allow the initial sprout build to be run in parallel, with the hope it'd speed it up.

The changes are:

1) Force all submodules' sub-makes to run in serial (with `-j 1`) as we can't guarantee that the submodule builds are parallelizable. I'm not 100% certain this is needed, but figured this was safer than allowing them to run in parallel.

2) Make a previously implicit dependency in the sprout makefile explicit instead, so that it can be safely run in parallel mode

3) Change the cassandra makefile so that we don't run 2 identical steps in parallel if we're running a parallel build

4) Change the cassandra makefile so that we don't include auto-generated files if they don't exist. To do this, I moved the auto-generation of the dependency files to the point at which we produce the `.o` files. We then include them with `-include` which will only include them if they don't already exist

The gains here weren't quite as good as I was hoping, but still seem worthwhile. In serial, I got:

```
ubuntu@ip-10-0-249-176:~/repos/sprout$ time make
...
real    20m1.121s
user    17m53.300s
sys     2m14.506s

```

And running in parallel (2 jobs simultaneously on my dual-core machine):
```
ubuntu@ip-10-0-249-176:~/repos/sprout$ time make -j2
...
real    16m53.964s
user    29m52.922s
sys     3m2.491s
```

I did see that my CPU usage went up from 50% to roughly 100% in the parallel build, so clearly we're now making better use of the CPU available.

I've not tried to default it to running in parallel, as I suspect that's too disruptive (and making too many assumptions about people's build environments)


src/Makefile
```diff
@@ -340,6 +340,7 @@ include ../modules/cpp-common/makefiles/alarm-utils.mk
 	$< -j "../sprout-base.root/usr/share/clearwater/infrastructure/alarms/sprout_alarms.json" -n "sprout"
 	mv sprout_alarmdefinition.h $@
 ${sprout_OBJECT_DIR}/main.o : ../usr/include/sprout_alarmdefinition.h
+${sprout_scscf.so_OBJECT_DIR}/scscfplugin.o : ../usr/include/sprout_alarmdefinition.h
```
johadalin: Do we need to add similar for the other plugins?

mk/cassandra.mk
```diff
@@ -24,9 +25,12 @@ ${CASSANDRA_BUILD_DIR}/interface/thrift:
 
 cassandra: ${LIB_PATH}
 
-${C_PATHS}: ${CASSANDRA_DIR}/interface/cassandra.thrift ${CASSANDRA_THRIFT_DIR} ${INSTALL_DIR}/bin/thrift
+${CASSANDRA_CPP_SENTINEL} : ${CASSANDRA_DIR}/interface/cassandra.thrift ${CASSANDRA_THRIFT_DIR} ${INSTALL_DIR}/bin/thrift
 	${INSTALL_DIR}/bin/thrift --gen cpp -o ${CASSANDRA_THRIFT_DIR} ${CASSANDRA_DIR}/interface/cassandra.thrift
 	cp ${CASSANDRA_GEN_DIR}/*.h ${INSTALL_DIR}/include/
+	touch $@
```
johadalin: Will this be cleaned up  when necessary for new builds if the files need to change? If i understand it right this is just a flag so we don't re-do this step?

src/Makefile
sebrexmetaswitch: No, just the scscfplugin depends on it because of https://github.com/Metaswitch/sprout/blob/8ac85f801a960ba18e586d26939c7ab7320f4be1/src/scscfplugin.cpp#L21

mk/cassandra.mk
sebrexmetaswitch: Yeah, this file is just a flag so we don't do this every time. If the files on the right are newer, this target will re-run, and touch the file at the end (so that next time, the target file is newer than the sources, and it won't run again until they change)

sebrexmetaswitch: As a fun aside, I just tested touching one of the dependencies specified in the `.d` file and it doesn't trigger a rebuild, but then I realised it never has. So I'm going to merge this PR and raise an issue for the broken dependency

PR #1956, opened 2017-10-09
[Reviewer: EM] SIP Priority Queueing

tomwatson1024:
Modifies Sprout to allow SIP messages to be prioritized by its worker threads based on the message content or type, and to avoid certain types of messages being rejected during overload.

Prioritizes and avoids rejecting SIP options polls.

Refactors thread_dispatcher and common_sip_options to be consistent with this new approach.

Adds some basic thread_dispatcher UTs.


tomwatson1024:
The changes pass the existing unit and live tests.

PR #1958, opened 2017-10-10
[Reviewer: Adam] Add sprout_{http/worker}_threads to override num_*

richardwhiuk:
Adam,

Hopefully an uncontroversial change to add a couple of configuration options for the thread counts to Sprout.

I've tested setting `sprout_worker_threads` and `sprout_http_threads` on a Chef deployment and checking that they get picked up.

### DIFF
--- a/sprout-base.root/etc/init.d/sprout
+++ b/sprout-base.root/etc/init.d/sprout
@@ -102,6 +102,16 @@ get_settings()
         authentication=Y
         . /etc/clearwater/config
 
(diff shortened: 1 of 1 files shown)


johadalin:
Looks reasonable :shipit: 
Then we should look at the best way to do some sort of profiling work around this

johadalin:
We should potentially also add these to the config reference docs, if we decide to keep  them implemented

PR #1961, opened 2017-10-11
Ignore downstream latency and spot blocking IO on transport thread. 

AlexHockey:
This builds on the code in https://github.com/Metaswitch/cpp-common/pull/695 and makes sprout:

- Stop the stopwatch when doing blocking IO.
- Produce a log when the PJSIP transport thread does any blocking IO. 

No UTs yet - this PR is to get early feedback on the approach before I go any further.


AlexHockey:
@eleanor-merry I've made this PR point at overload-integration rather than master. I'm assigning to you to merge at a suitable time.

PR #1962, opened 2017-10-12
[Reviewer: Mat] Run AoR Timeout Task on worker pool

richardwhiuk:
Mat,

Currently Chronos processes work for a AoRTimeout immediately after it calls `send_http_reply(HTTP_OK);` to reply to the HTTP request, in the same function call.

However, the HTTP server stack, libevhtp, doesn't process the reply until the thread returns to it's control. This means that the reply isn't sent until the function returns, which in the case of an AoRTimeout when we need to potentially send a request to the HSS and perform work in a different GR site may be hundreds of ms later.

This means that we can see double pops of timers, as Chronos doesn't get a response in time, and thus the timer pops multiple times.

This fixes it by moving the work to the Sprout worker thread pool, using our common PJUtils code to do so.

I've tested that the UTs pass, and the live tests pass, including "SUBSCRIBE - Registration timeout" which should test exactly this behaviour.

### DIFF
--- a/include/chronoshandlers.h
+++ b/include/chronoshandlers.h
@@ -14,6 +14,8 @@
 
 #include "handlers.h"
 
+class ChronosAoRTimeoutTaskHandler;
+
 class ChronosAoRTimeoutTask : public AoRTimeoutTask
 {
 public:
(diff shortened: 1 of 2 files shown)


MatMeredith:
:+1:

PR #1963, opened 2017-10-13
Raise SAS log when Sprout decides not to do an ENUM lookup

metaswitch-colin-haywood:
Hi Alex,

Some thoughts on this fix:
- is `uri_class == LOCAL_PHONE_NUMBER` the only case we want to log for? (It maps to the "enforce_global_only_lookups=Y" settings option via `URIClassifier::enforce_global`, which was the main cause of complaint in the original issue https://github.com/Metaswitch/clearwater-issues/issues/1467)
- any thoughts on that option being implemented in the wrong place? That is to say, the "enforce_global_only_lookups=Y" option controls ENUM lookup behaviour, but instead is used to control URI classification behaviour (which then has knock-on effects on ENUM lookups). So as it stands a change to `URIClassifier` could introduce an unwanted side effect of raising/not raising this log. But it seems like a big task to refactor the code now :/ [see also a comment from SL on the new SAS log change in the linked PR]
- do I want to log in both the `translate_request_uri` and `update_request_uri_np_data` functions or only the former?

I've live-tested this with a call between two local subscribers (and it passes make full_test).

Thanks,

Colin


AlexHockey:
> - is `uri_class == LOCAL_PHONE_NUMBER` the only case we want to log for? (It maps to the "enforce_global_only_lookups=Y" settings option via `URIClassifier::enforce_global`, which was the main cause of complaint in the original issue https://github.com/Metaswitch/clearwater-issues/issues/1467)

I think support are in a better position to make a decision on this than I am. Do you think users will be confused if other types of URI do not receive ENUM lookups? 

> - any thoughts on that option being implemented in the wrong place? That is to say, the "enforce_global_only_lookups=Y" option controls ENUM lookup behaviour, but instead is used to control URI classification behaviour (which then has knock-on effects on ENUM lookups). So as it stands a change to `URIClassifier` could introduce an unwanted side effect of raising/not raising this log. But it seems like a big task to refactor the code now :/ [see also a comment from SL on the new SAS log change in the linked PR]

I don't particularly like the way that the config option changes the classification, but this is mainly due to the name of the config option. If it were called "treat_all_numbers_as_global" I don't think it would bother me. So I personally don't think this is worth updating (and if we were to make a change it should be to change the name of the option).

> - do I want to log in both the `translate_request_uri` and `update_request_uri_np_data` functions or only the former?

Again, I think that support are better placed to answer this than me.

PR #1965, opened 2017-10-13
Optimize initial reg

MatMeredith:
Optimized initial Register processing in Sprout taking advantage of the new <PreviousRegistrationState> indication on reg-data PUT responses: if this is an initial registration then assume that there won't be any bindings in the AoR store so attempt to ADD the data rather than getting it and updating it.  Prior to this change, before writing the data to the local store, we would first try to GET it from the local store and then (since there isn't any data in the local store) also check each of the remote stores in sequence (which can be expensive in a GR deployment with significant inter-site latency).

See also: https://github.com/Metaswitch/cpp-common/pull/697 and https://github.com/Metaswitch/homestead/pull/485.

At the same time, this PR also:
- Gets rid of duplicated code between handlers.cpp and registrarsproutlet.cpp for reading binding data from local / remote stores
- Gets rid of the binding from the Authentication Sproutlet into the Registrar Sproutlet for determining expiry times.   This is required as a result of moving `expiry_for_binding` into RegistrationUtils.   That change was going to be required at one stage but actually ended up not being needed.   I've left the change in though because it seems like a good function to have in RegistrationUtils, and passing max_expires to the Authentication Sproutlet on creation is consistent with the way that the Subscription Sproutlet works.   @AlexHockey Let me know if there is a reason we didn't go this way for the Authentication Sproutlet?).
- Fixes a bug in write_to_store where we previously claimed to return the longest expiry time, but actually just returned the expiry time for the last contact in the request.   The only implication is that we use this time to determine whether the request is a REGISTER or de-REGISTER for stats, and I think that how we count a REGISTER that registers some contacts and de-registers others is undefined...
- Fixes the Registrar UTs which claimed to register a binding and then test retrieving it, but actually the second request was just being turned about by PJSIP because we were failing to update the branch parameter.

ToDo: 
- add a UT which tests the case where the ADD fails and verifies that we fall back to get and update instead.
- live test


MatMeredith:
Missing UT now added.

AlexHockey:
>  @AlexHockey Let me know if there is a reason we didn't go this way for the Authentication Sproutlet?

No particular reason. The new code works equally well :-).

src/registrarsproutlet.cpp
```diff
@@ -870,230 +874,217 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
+AoRPair* RegistrarSproutletTsx::write_to_store(
+                   SubscriberDataManager* primary_sdm,         ///<store to write to
+                   std::string aor,                            ///<address of record to write to
+                   AssociatedURIs* associated_uris,
+                                                               ///<Associated IMPUs in Implicit Registration Set
+                   pjsip_msg* req,                             ///<received request to read headers from
+                   int now,                                    ///<time now
+                   int& max_expiry,                            ///<[out] longest expiry time
+                   bool is_initial_registration,               ///<Does the caller believe that this is an initial registration?
+                   bool& out_no_bindings_found,                ///<[out] true if no bindings were found.
```
johadalin: On my third pass now, and just seen why this was messing my head up. This is called above with the bools: (prev_regstate == RegDataXMLUtils::STATE_NOT_REGISTERED), and is_initial_registration. 
These map to is_initial_registration, and out_no_bindings_found. 

I REALLY don't like that.
Can we not have two different variables swapping the same name between them across the function call please

src/registrarsproutlet.cpp
```diff
@@ -870,230 +874,217 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
+
+      // If this fails because there actually is already data for the AoR then
+      // we'll get a DATA_CONTENTION error and retry.  Next time we want to do
+      // the usual GET and update.
+      is_initial_registration = false;
+    }
+    else
+    {
+      if (!RegistrationUtils::get_bindings(&aor_pair,
+                        aor,
```
johadalin: alignment

src/registrarsproutlet.cpp
```diff
@@ -870,230 +874,217 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
+                        primary_sdm,
+                        backup_sdms,
+                        backup_aor,
+                        trail()))
+      {
+        // Error reading bindings.
```
johadalin: Is it? Isn't this a store error getting an AoR?

src/registration_utils.cpp
```diff
@@ -765,3 +765,112 @@ bool RegistrationUtils::remove_bindings(SubscriberDataManager* sdm,
...
+/// the following places and stop as soon as it finds a record for the
+/// specified AoR that has > 0 bindings.
+/// -- the primary_sdm
+/// -- the provided backup_aor_pair (if not null)
+/// -- the specified backup_sdms (if any have been provided).
+/// The function returns true unless it is unable to connect to the
+/// primary_sdm.  If it returns True then it will have created an AoRPair (even
+/// if it is empty because no bindings were found) and ownership of this
+/// AoRPair is transferred to the caller on exit (aor_pair).
+bool RegistrationUtils::get_bindings(AoRPair** aor_pair,
```
johadalin: Also, this returns true even if it fails to find bindings. If it finds a local AoR that has no bindings, but fails to contact any remotes, this returns true. That's surely wrong

src/registrarsproutlet.cpp
```diff
@@ -870,230 +874,217 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
+      aor_pair = new AoRPair(aor);
+
+      // If this fails because there actually is already data for the AoR then
+      // we'll get a DATA_CONTENTION error and retry.  Next time we want to do
+      // the usual GET and update.
+      is_initial_registration = false;
+    }
+    else
+    {
+      if (!RegistrationUtils::get_bindings(&aor_pair,
```
johadalin: I don't really like that this get is used as a check. It isn't clear here without going and reading the function what this is really doing.

src/registrarsproutlet.cpp
```diff
@@ -870,230 +874,217 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
+                        backup_sdms,
+                        backup_aor,
+                        trail()))
+      {
+        // Error reading bindings.
+        TRC_ERROR("Failed to get AoR binding for %s", aor.c_str());
+        break;
       }
-      contact = (pjsip_contact_hdr*)pjsip_msg_find_hdr(req, PJSIP_H_CONTACT, contact->next);
+      out_no_bindings_found = out_no_bindings_found && aor_pair->get_current()->bindings().empty();
```
johadalin: I think this is flawed too.
If 
on first loop:

    we fall into the else above, 
    read a local AoR with no bindings, 
    read a remote with bindings, and copy them over
    set out_no_bindings_found to false
    fail the set below from data contention

on second loop.

    else branch above
    read local, no bindings
    fail remote reads
    out_no_bindings_found is then false && true, which is false, which in this case is now not true

we return bindings found, but we found no bindings.

src/registrarsproutlet.cpp
```diff
@@ -201,7 +201,7 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
   {
     num_contacts++;
     pjsip_expires_hdr* expires = (pjsip_expires_hdr*)pjsip_msg_find_hdr(req, PJSIP_H_EXPIRES, NULL);
-    expiry = _registrar->expiry_for_binding(contact_hdr, expires);
+    expiry = RegistrationUtils::expiry_for_binding(contact_hdr, expires, _registrar->_max_expires);
```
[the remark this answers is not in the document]
MatMeredith: Could do I guess...  I thought globals were generally "bad" and we tried to avoid them.  Do you think there is any other precedent in Sprout for storing config in global variables?   I couldn't see any and if not, adding it just for this seems like it's going to be confusing through inconsistency...

src/registrarsproutlet.cpp
```diff
@@ -445,12 +446,14 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
   bool all_bindings_expired;
+  int max_expiry;
   AoRPair* aor_pair = write_to_store(_registrar->_sdm,
                                      aor,
                                      &associated_uris,
                                      req,
                                      now,
-                                     expiry,
+                                     max_expiry,
+                                     (prev_regstate == RegDataXMLUtils::STATE_NOT_REGISTERED),
```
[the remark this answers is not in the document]
MatMeredith: Yes -- this is confusing.   I'll fix.   Ditto the other comments related to this.

src/registrarsproutlet.cpp
```diff
@@ -870,230 +874,217 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
-  // The registration service uses optimistic locking to avoid concurrent
-  // updates to the same AoR conflicting.  This means we have to loop
-  // reading, updating and writing the AoR until the write is successful.
-  AoRPair* aor_pair = NULL;
-  bool backup_aor_alloced = false;
-  bool is_initial_registration = true;
-  bool all_bindings_expired = false;
-  Store::Status set_rc;
+  // @@@ ToDo: Optimization -- we don't need to repeat this in the loop in the
+  // case of an initial registration -- it's going to be the same outcome
```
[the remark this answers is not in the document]
MatMeredith: Sorry -- that's just an out of date comment.   It just needs deleting.

src/registrarsproutlet.cpp
```diff
@@ -870,230 +874,217 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
 
-        if (binding_id == "")
+      if ((cid != binding->_cid) ||
+          (cseq > binding->_cseq))
+      {
+        // Either this is a new binding, has come from a restarted device, or
+        // is an update to an existing binding.
+        binding->_uri = contact_uri;
+
+        // TODO Examine Via header to see if we're the first hop
```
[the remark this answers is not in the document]
MatMeredith: Nor do I.   Feel free to take it up with Sathiyan -- this isn't code I've touched...

src/registrarsproutlet.cpp
MatMeredith: Fine.   I think this is consistent with what we do in lots of other places, but I agree it could be clearer.

src/registrarsproutlet.cpp
MatMeredith: So this is the way the logic has always been.   If bindings are found (at any point) then we return no_bindings_found=false.   This flag is supposed to reflect the state of the stores _before_ we processed this request.  In your example above, there _were_ some bindings in one of the stores for this AoR, so this doesn't seem like unreasonable behaviour?

src/registration_utils.cpp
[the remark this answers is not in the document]
MatMeredith: Yeah -- that's fair.   I'm not convinced that we are consistent about this elsewhere (I suspect AoRs were originally just for bindings and so the terms are used a bit interchangeably), but that's no reason not to get it right here.

MatMeredith: I think the comment accurately reflects the behaviour here?
- If we're asked to look up the data for an AoR and we can't connect to the local store we consider that an error and we return false.
- If we can contact the primary store then we consider that to be sufficient for us to return success and a view as to what the data is for this AoR.
- Backup stores are seen as being of secondary importance.   If we don't find any data in the primary store and we can contact the backup stores then we will, but we don't consider it to be a failure if we can't contact them.
This was the existing behaviour and I haven't changed it.   

Is this the right behaviour?   I think that local SDM failure essentially indicates that all Astaire nodes in  the site are unavailable?   I vaguely recall that there was an architectural concerns whiteboard about this some time ago where we discussed whether it was better for the SPN / DGN nodes in this scenario to start talking cross-site or fail requests and so force GR failover.   Do you recall that, or what the outcome was?

MatMeredith: I think the existing comment on the function was OK, but I've attempted to further clarify...

src/ut/registrar_test.cpp
```diff
@@ -3415,6 +3443,69 @@ TEST_F(RegistrarTestMockStore, SubscriberDataManagerGetsFail)
...
+  std::string expiry_time = std::to_string(time(NULL) + 300);
+
+  std::string initial_data = "{\"bindings\":{\"<urn:uuid:00000000-0000-0000-0000-777777777777>:1\":{\"uri\":\"sip:f5cc3de4334589d89c661a7acf228ed7@10.114.61.214:5061;transport=tcp;ob\",\"cid\":\"0gQAAC8WAAACBAAALxYAAAL8P3UbW8l4mT8YBkKGRKc5SOHaJ1gMRqs1042ohntC@10.114.61.213\",\"cseq\":10000,\"expires\":" + expiry_time + ",\"priority\":0,\"params\":{\"+sip.ice\":\"\",\"+sip.instance\":\"\\\"<urn:uuid:00000000-0000-0000-0000-777777777777>\\\"\",\"reg-id\":\"1\"},\"path_headers\":[\"<sip:GgAAAAAAAACYyAW4z38AABcUwStNKgAAa3WOL+1v72nFJg==@ec2-107-22-156-220.compute-1.amazonaws.com:5060;lr;ob>\"],\"paths\":[\"sip:GgAAAAAAAACYyAW4z38AABcUwStNKgAAa3WOL+1v72nFJg==@ec2-107-22-156-220.compute-1.amazonaws.com:5060;lr;ob\"],\"private_id\":\"Alice\",\"emergency_reg\":false}},\"subscriptions\":{},\"associated-uris\":{\"uris\":[{\"uri\":\"sip:6505550231@homedomain\",\"barring\":false}],\"wildcard-mapping\":{}},\"notify_cseq\":2,\"timer_id\":\"post_identity\",\"scscf-uri\":\"sip:scscf.sprout.homedomain:5058;transport=TCP\"}";
```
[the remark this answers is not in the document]
MatMeredith: I've added a comment.  And I've split it up.   It still looks ugly but I'm struggling to see how to make it look a _lot_ better without more effort than is justified though.

src/registrarsproutlet.cpp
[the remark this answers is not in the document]
johadalin: Sure. Not really sure what the accepted way of handling this is. Just seemed a bit odd having it done twice, but not really fussed either way, and if it's not already done here, no need to change

src/registrarsproutlet.cpp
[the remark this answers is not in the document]
johadalin: Right. Fair enough.
@sathiyan-sivathas any view on how important it is that we do this TODO, and is it tracked anywhere other than inline in the code?

johadalin:
These changes make it a bunch clearer what's going on. Much happier with it all, and assuming it all still passes tests etc. i'd say good to merge :shipit:

src/registrarsproutlet.cpp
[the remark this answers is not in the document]
sathiyan-sivathas: I don't think this is my TODO. I added some when I did the Sproutlet refactor but I believe I addressed them all. This TODO is from the initial Clearwater code so you can probably leave it.

https://github.com/Metaswitch/sprout/blame/3f68c22b13046127abc5cc00ee44705be4bffc35/src/registrar.cpp#L369-L370

PR #1966, opened 2017-10-16
[Reviewer AJH] Treat originating SIP URIs for global phone numbers as Tel URIs

steelyvoid:
Hi Alex

Can you review this one for me?  The motivation is that we need to handle a scenario wherein AS'es forward originating INVITEs to the platform with a P-Asserted_ID of "sip:<DN>@<arbitrary domain>;user=phone", expecting us to treat the URI as though it was a Tel URI.  I discussed with MIRW, and we agreed that this should be the default behaviour, but that there should be a configuration option that allows it to be turned off.

Tested live using scripts to send in INVITEs with one of these user=phone URIs.  Confirmed that the lookup was successful and the INVITE reached the target as expected.

Note that I disabled the SAS log for invocations from bono and the ifchandler class, because I wasn't convinced that I could necessarily call on a valid trail ID in these cases (I think the ifchandler instance is test-only - can't see that function called from non-UT code).


src/pjutils.cpp
```diff
@@ -296,6 +298,31 @@ pjsip_uri* PJUtils::orig_served_user(const pjsip_msg* msg)
...
+      (uri != NULL) && 
+      PJSIP_URI_SCHEME_IS_SIP(uri))
+  {
+    // Determine whether this SIP URI can be translated into a Tel URI
+    URIClass uri_class = URIClassifier::classify_uri(uri, false);
+    pjsip_sip_uri* sip_uri = (pjsip_sip_uri*)uri;
+
+    if (uri_class == GLOBAL_PHONE_NUMBER)
+    {
+      TRC_DEBUG("Change request URI from SIP URI to tel URI");
```
sathiyan-sivathas: I think you copied this log from the code in the I-CSCF that does this translation for terminating services. Not a functional problem but it would be pretty confusing to see a log saying the request URI was being changed during originating processing! Could you change it?

steelyvoid:
Thanks Sathiyan - very much appreciated!

steelyvoid:
Not necessary for 11.1 (implementing using MMF there instead)

PR #1967, opened 2017-10-18
Fix graylisting ACK bug.

AlexHockey:
This PR fixes a bug in graylisting where a host would get blacklisted whenever it gets sent an ACK. 

The problem was that the `Target` class white/blacklists the host automatically if it hasn't been told the state of the host by the time the `Target` gets destroyed or unset (e.g. because we're moving onto another target). This caused a problem for ACKs since the only way we know a host is definitely healthy is receiving a response from it, and ACKs don't have responses. 

I've fixed this by having the `UACTsx` tell the `Target` whether the host should be blacklisted by default. This flag is only set for requests that have a PJSIP transaction (meaning they require a response), and the host is not a stateless proxy. 

Tested in UT. ST will follow when I've merged this PR to the integration branch.

### DIFF
--- a/src/basicproxy.cpp
+++ b/src/basicproxy.cpp
@@ -2274,7 +2274,31 @@ bool BasicProxy::UACTsx::get_next_server()
     AddrInfo addr;
     if (_servers_iter->next(addr))
     {
-      _current_server.set(addr, _stateless_proxy);
+      // Work out whether to blacklist this address by default (i.e. if the
+      // UACTsx gives up on it without explicitly determining its health. We
+      // blacklist by default if:
+      //
+      // -  The request has a transaction associated with it, meaning that the
+      //    request is not an ACK request. ACKs do not have a response so there
+      //    is no clear indication that it has succeeded.
+      //
+      // -  The address is a stateful proxy. For stateful proxies, the most
+      //    likely thing is that the host has failed in some weird way that the
+      //    UACTsx doesn't cope with, so the best thing to do is blacklist the
+      //    host.  State*less* proxies are only blacklisted on transport
+      //    failures. These are easy to spot, so it is unlikely the UACTsx
+      //    didn't spot this, and the best thing to do is mark the host as
+      //    successful.
+      bool blacklist_by_default = (_tsx != nullptr) && !_stateless_proxy;
+      _current_server.set(addr, blacklist_by_default);
+
+      if (Log::enabled(Log::DEBUG_LEVEL))
+      {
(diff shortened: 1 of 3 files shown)


src/ut/basicproxy_test.cpp
```diff
@@ -4102,6 +4102,96 @@ TEST_F(BasicProxyTest, FailedProbeDoesNotUngraylist)
...
+  RespMatcher(100).matches(tdata->msg);
+  tp->expect_target(tdata);
+  free_txdata();
+
+  ASSERT_EQ(1, txdata_count());
+  tdata = current_txdata();
+  ReqMatcher("INVITE").matches(tdata->msg);
+
+  // Check that it was sent to the first server, which should not have been
+  // graylisted by being sent an ACK.
```
sathiyan-sivathas: blacklisted?

src/ut/basicproxy_test.cpp
```diff
@@ -4102,6 +4102,96 @@ TEST_F(BasicProxyTest, FailedProbeDoesNotUngraylist)
...
+  // Check the response is forwarded back to the source.
+  ASSERT_EQ(1, txdata_count());
+  tdata = current_txdata();
+  tp->expect_target(tdata);
+  RespMatcher(200).matches(tdata->msg);
+  free_txdata();
+
+  delete tp;
+}
+
```
sathiyan-sivathas: Extra newline

src/basicproxy.cpp
```diff
@@ -2274,7 +2274,31 @@ bool BasicProxy::UACTsx::get_next_server()
...
+      //    failures. These are easy to spot, so it is unlikely the UACTsx
+      //    didn't spot this, and the best thing to do is mark the host as
+      //    successful.
+      bool blacklist_by_default = (_tsx != nullptr) && !_stateless_proxy;
+      _current_server.set(addr, blacklist_by_default);
+
+      if (Log::enabled(Log::DEBUG_LEVEL))
+      {
+        std::string host_str = addr.to_string();
+        TRC_DEBUG("Selected host %s, %s blacklist by default",
```
sathiyan-sivathas: I don't think this sentence makes sense. Can you say "be blacklisted" instead?

src/ut/basicproxy_test.cpp
AlexHockey: I think that's common (as a way of creating more of a break between tests). There are lots of double blank lines already in the file.

PR #1969, opened 2017-10-18
Don't query Homestead after invoking terminating AS in the mainline case

MatMeredith:
I figured that this was so easy to fix in the mainline case that we just should.

I was a bit dubious about just returned registered=false in the case that we fail to query Homestead, but worth noting that:
- the cases where we now come in to this function and haven't already queried homestead is where we had previously called is_retarget()
- is_retarget doesn't check the return code from querying homestead.   If it fails then we will just carry on 
and is_user_registered would have returned the default value of is_registered which is false.
Still doesn't seem great -- but it's no worse than it was I think.

And I dislike the previous assumption that is_registered will only ever be called after we happen to have called one of the functions that queries Homestead.   Someone will end up calling is_registered in a case where we don't happen to have already queried homestead and they will just silently get the wrong result.

### DIFF
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -1026,17 +1026,24 @@ bool SCSCFSproutletTsx::is_retarget(std::string new_served_user)
 {
   std::string old_served_user = _as_chain_link.served_user();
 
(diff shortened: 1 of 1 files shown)


MatMeredith:
Closing at least for now -- this doesn't actually solve the problem...

PR #1970, opened 2017-10-18
Move IP address allowed host state checking to the base resolver

AlexHockey:
Companion PR to https://github.com/Metaswitch/cpp-common/pull/699. This copes with the host state logic being moved into the base resolver, and also updates the UTs.

### DIFF
--- a/src/ut/sipresolver_test.cpp
+++ b/src/ut/sipresolver_test.cpp
@@ -675,75 +675,12 @@ TEST_F(SIPResolverTest, NoMatchingNAPTR)
             RT(_sipresolver, "sprout.cw-ngv.com").resolve_iter());
 }
 
-// Test the behaviour of SIPResolver's IP address allowed host state
-// verification
+// Check that the SIPResolver is checking the host state before accepting an IP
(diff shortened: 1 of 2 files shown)


src/ut/sipresolver_test.cpp
```diff
@@ -81,15 +82,36 @@ class SIPResolverTest : public ::testing::Test
                                            service, regex, replacement);
   }
 
-  void add_ip_to_blacklist(const std::string address,
-                           const int port=5060)
+  AddrInfo ip_port_to_addrinfo(const std::string address,
```
sathiyan-sivathas: You've got two very similar functions to turn a IP address into an AddrInfo (here and in the cpp-common tests). Is it worth putting a common function in Utils?

It's a bit cheeky me asking this since you gave me the same feedback in Metaswitch/astaire#87 and I ignored it! But if we have two implementations if feels worth it now.

AlexHockey: I'm going to merge this PR (as I'm under a bit of time pressure), but I will consider this.

PR #1971, opened 2017-10-19
[Reviewer: Ellie] Hss connection refactor

Ying-metaswitch:
Store two structs as member fields of HSS: hss_query_param, hss_query_return.
So that get_registration_data, update_registration_state and any other related functions will only take them as argument, rather than a variable number (up to 13) of arguments. The previous implementation also makes it confusing to differentiate between parameters used for query and returned values.  

A few unit tests commented out as havn't got time to rework it. But rest of UT pass fine.


src/handlers.cpp
```diff
@@ -198,30 +198,35 @@ static void update_hss_on_aor_expiry(std::string aor_id,
...
-                                                  associated_uris,
+  HSSConnection::hss_query_param_t hss_query_param(aor_id);
+  HSSConnection::hss_query_return_t hss_query_return;
+
+  HTTPCode http_code = hss->get_registration_data(hss_query_param,
+                                                  hss_query_return,
                                                   trail);
 
-  unbarred_irs_impus = associated_uris.get_unbarred_uris();
+  ifc_map = hss_query_return.service_profiles;
```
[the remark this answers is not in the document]
Ying-metaswitch: This method get_reg_data currently doesn't take irs_info as input, and the ifc_map is used only once. Should I still add irs_info and replace the usage?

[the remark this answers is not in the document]
Ying-metaswitch: Well I think so

Ying-metaswitch:
Sorry Ellie I messed up the git history for the review. I did a commit amend by accident and then tried `git reset --soft HEAD@{1}`, it didn't help...

Ying-metaswitch:
All UT passed, and `make full_test` passed. Passing back for review, I will resolve the conflict before merging.

eleanor-merry:
You're going to sort out the merge conflicts before review.

Ying-metaswitch:
Conflict resolved and `make full_test` passed.

include/hssconnection.h
```diff
@@ -32,6 +32,48 @@
...
+    std::string _prev_regstate;
+    std::map<std::string, Ifcs> _service_profiles;
+    AssociatedURIs _associated_uris;
+    std::vector<std::string> _aliases;
+    std::deque<std::string> _ccfs;
+    std::deque<std::string> _ecfs;
+
+    irs_info() : 
+      _regstate(""),
+      _prev_regstate(""),
```
eleanor-merry: Why is this a member of irs_info?

src/ut/handlers_test.h
```diff
@@ -36,11 +39,11 @@ class TestWithMockSdms : public SipTest
 
   virtual void SetUp()
   {
-    store = new MockSubscriberDataManager();
-    remote_store1 = new MockSubscriberDataManager();
-    remote_store2 = new MockSubscriberDataManager();
-    mock_hss = new MockHSSConnection();
-    stack = new MockHttpStack();
+    store = new NiceMock<MockSubscriberDataManager>();
```
eleanor-merry: This is quite a change - I don't think you want this. Please can you revert the moves to NiceMock throughout.

include/hssconnection.h
Ying-metaswitch: I think of previous registration state of an IRS as intrinsic information of this IRS. It's retrieved from homestead xml and used by RegistrationSproutlet to determine if the registration for this IRS is an initial registration. It's used in both querying and updating reg data, and therefore stored in irs_info rather than irs_query.

eleanor-merry: Ah OK.

PR #1972, opened 2017-10-19
[Reviewer: RJW2] Blacklisted scscfs

metaswitch-ak10:
Hi Richard,

Would you mind reviewing this?
This is the enhancement that allows subscriber migration to clearwater v11 by using S-CSCF blacklist. The blacklist is loaded from the `shared_config` file that lives in `/etc/clearwater/`. Adding an entry of the form `blacklisted_scscfs='<scscf1>,<scscf2>'` to this config file will add `scssf1` and `scscf2` to the blacklist.

I have live-tested the gitlab version of the build by uploading the code onto the l3-cc3-spn1 node and running the following test.
1. add l3-cc4-sprout1 to blacklisted S-CSCFs
1. point Perimeta to l3-cc4 and make sure a SIP client registers on the l3-cc4-sprout1 S-CSCF
2. point Perimeta to l3-cc3 and send another REGISTER request
Result: The HSS returned a blaclisted S-CSCF (l3-cc4-sprout1) so the HSS is queried for capabilities and the SIP client subsequently registers on the l3-cc3-spn1 S-CSCF. Moreover, no messages get routed through the l3-cc4-sprout S-CSCF.

For the current github version in question, I have written a custom Unit Test to carry out the above test. After running `make full_test` from my local repo, all UTs passed. I have also checked that all my changes to the code are covered by the UTs by running `make coverage_raw`.

[RJW2: Edit formatting]


src/main.cpp
```diff
@@ -1011,6 +1014,14 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
       }
       break;
 
+    case OPT_BLACKLISTED_SCSCFS:
+      TRC_INFO("Hello");
+      options->blacklisted_scscfs.clear();
+      Utils::split_string(std::string(pj_optarg), ',', options->blacklisted_scscfs, 0, false);
+      TRC_INFO("%d blacklisted S-CSCFs passed on the command line",
+                options->blacklisted_scscfs.size());
```
richardwhiuk: Given we never expect a huge number (and the model will break if they do), I'd suggest:

```
      TRC_INFO("%d blacklisted S-CSCF URIs passed on the command line: %s",
                options->blacklisted_scscfs.size(), pj_optarg);
```

This way we'll log out the S-CSCF URIs blacklisted.

src/main.cpp
```diff
@@ -1021,7 +1032,7 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
     case OPT_OVERRIDE_NPDI:
       options->override_npdi = true;
       TRC_INFO("Number portability lookups will be done on URIs containing the 'npdi' indicator");
-      break;
+      break;
```
richardwhiuk: Please don't add trailing whitespace!

src/main.cpp
```diff
@@ -1718,6 +1729,7 @@ int main(int argc, char* argv[])
   opt.dummy_app_server = "";
   opt.http_acr_logging = false;
   opt.homestead_timeout = 750;
+  opt.blacklisted_scscfs.push_back("");
```
richardwhiuk: Why do we need to do this? Surely any empty vector/set is fine?

src/ut/icscfsproutlet_test.cpp
```diff
@@ -215,6 +215,112 @@ class ICSCFSproutletTest : public ICSCFSproutletTestBase
...
+  
+  // Check the response is forwarded back to the source.
+  ASSERT_EQ(1, txdata_count());
+  tdata = current_txdata();
+  expect_target("TCP", "1.2.3.4", 49152, tdata);
+  RespMatcher r2(200);
+  r2.matches(tdata->msg);
+
+  free_txdata();
+  _hss_connection->delete_result("/impi/6505551000%40homedomain/registration-status?impu=sip%3A6505551000%40homedomain&visited-network=homedomain&auth-type=REG");
```
richardwhiuk: I don't think this UT needs to contain anything between Line 231 and 270. The UT should just be able to test the later flow by set the returned S-CSCF on the HSS Connection, as you are doing on Line 279.

src/ut/icscfsproutlet_test.cpp
```diff
@@ -215,6 +215,112 @@ class ICSCFSproutletTest : public ICSCFSproutletTestBase
...
+  
+  // Add the returned S-CSCF together with others to the blacklist.
+  _icscf_sproutlet->set_blacklisted_scscfs({"sip:scscf2.homedomain:5058;transport=TCP",
```
richardwhiuk: Hmm, you could extend from a different Test harness (e.g. ICSCFBlacklistedTest instead of ICSCFSproutletTest) and then pass it into the constructor. That feels nicer, but it's a fair amount of code. I think the alternative of just directly sending the member variable is better - i.e.

```
_icscf_sproutlet->_blacklisted_scscfs = {"sip:scscf2.homedomain:5058;transport=TCP",
                                         "sip:scscf3.homedomain:5058;transport=TCP",
                                         "sip:scscf4.homedomain:5058;transport=TCP",
                                         "sip:scscf5.homedomain:5058;transport=TCP"};
```

src/ut/icscfsproutlet_test.cpp
```diff
@@ -215,6 +215,112 @@ class ICSCFSproutletTest : public ICSCFSproutletTestBase
...
+  tdata = current_txdata();
+  expect_target("TCP", "1.2.3.4", 49152, tdata);
+  RespMatcher r4(200);
+  r4.matches(tdata->msg);
+
+  free_txdata();
+  _hss_connection->delete_result("/impi/6505551000%40homedomain/registration-status?impu=sip%3A6505551000%40homedomain&visited-network=homedomain&auth-type=REG");
+  _hss_connection->delete_result("/impi/6505551000%40homedomain/registration-status?impu=sip%3A6505551000%40homedomain&visited-network=homedomain&auth-type=CAPAB");
+
+  _icscf_sproutlet->set_blacklisted_scscfs({});
```
richardwhiuk: Don't really need to bother doing this - the harness should be recreated for the next test.

include/icscfrouter.h
```diff
@@ -27,7 +27,8 @@ class ICSCFRouter
               SCSCFSelector* scscf_selector,
               SAS::TrailId trail,
               ACR* acr,
-              int port);
+              int port,
+              std::vector<std::string> blacklisted_scscfs={});
```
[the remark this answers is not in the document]
metaswitch-ak10: Should I also change the vector _attempted_scscfs into a set? I believe it is used in a similar way.

src/icscfrouter.cpp
```diff
@@ -92,9 +108,14 @@ int ICSCFRouter::get_scscf(pj_pool_t* pool,
     else if (_queried_caps)
     {
       // We queried capabilities from the HSS, so select a suitable S-CSCF.
+      // We pass both _blacklisted_scscfs and _attempted_scscfs to be rejected 
+      // since these are not suitable S-CSCFs.
+      std::vector<std::string> _reject_scscfs;
+      _reject_scscfs.insert(_reject_scscfs.end(), _attempted_scscfs.begin(), _attempted_scscfs.end());
+      _reject_scscfs.insert(_reject_scscfs.end(), _blacklisted_scscfs.begin(), _blacklisted_scscfs.end());
```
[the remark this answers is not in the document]
metaswitch-ak10: However, only the first blacklisted S-CSCF encountered is added to _attempted_scscfs. I pass _blacklisted_scscfs in case there are multiple blacklisted S-CSCFs so that they do not get selected.

metaswitch-ak10:
Hi Richard,
I have made the changes you suggested. I also had a question about the _attempted_scscfs vector. I believe this is used in the same way as my _blacklisted_scscfs vector, which, as you suggested, I have changed to a set. Should I therefore change _attempted_scscfs into a set too?

richardwhiuk:
Sorry, I was away last week, and have only just got round to this again.

Assuming you've retested these changes in the same scenarios given above, I think they look fine.

Regarding _attempted_scscfs, it'd be nice if that was a set as well, but it's not a pressing concern.

metaswitch-ak10:
@richardwhiuk Thank you for your review and your help.

PR #1974, opened 2017-10-20
[BGCF-597] Add Via headers for internal network function hops

huws:
This PR adds Via headers for internal hops between network functions.  This allows these network functions to determine the party that sent the received message.

Also contains a tidy-up fix, made in passing, which ensures that routing headers (Route, Record-Route and Via) are added immediately above any other headers of the same type.  This avoids messing up SIP messages, by mixing headers together.  While this has no function effect (SIP cares about the relative order of headers of a given type, but not the relative order between different header types), it makes diagnosing routing problems in SIP flows easier for humans, as they don't have to scan the entire message when inspecting headers of a given type.

The new code has 100% coverage.  I'm running the Live Tests against this now.


huws:
Chris - please can you do the internal code review?

PR #1976, opened 2017-10-25
[Reviewer: Adam] Send 480 if unregistered user tries to SUBSCRIBE

sebrexmetaswitch:
This is a fix for https://github.com/Metaswitch/clearwater-issues/issues/2305

Rel-10 of TS 24.229 doesn't give any guidance on how the S-CSCF should respond to an unregistered user sending a SUBSCRIBE message, but rel-13 is more explicit and says:

> When an incoming SUBSCRIBE request addressed to S-CSCF arrives containing the Event header field with the reg event package, the S-CSCF shall:
>    0) if the Request-URI of the SUBSCRIBE request contains a URI for which currently no binding exists, then send a 480 (Temporarily Unavailable) response indicating that the subscription was not successful and skip the remainder of the steps;

Previously, we sent a 504 for some reason, and had a UT to check that we did. I checked though and the UT was just to get more coverage, it wasn't because we thought we should send a 504 specifically.

I've added a SAS log too, in the other PR

I've checked `make full_test` still passes and run the live tests over a deployment with this fix.

I've not explicitly live-tested that we now send the 480, but I'm happy that the UT does that sufficiently


johadalin:
Yup, looks sensible. I like the comment reference to the right bit of the specs. The SAS log looks good to go too :+1:

PR #1983, opened 2017-10-27
[BGCF-598] Allow host state restrictions to be passed downstream

huws:
This fixes a bug where it is not possible for one sproutlet to place restrictions on required host state for requests (e.g. only attempt to send to whitelisted addresses), if there is another sproutlet downstream of it.

We now pass the required host state restrictions downstream, and pass any internal error state back upstream, between sproutlets.

I discussed this approach briefly with MIRW, and the changes are fairly minor.  Please advise whether you think I should also get AJH to look at this.


include/sproutletproxy.h
```diff
@@ -434,6 +438,12 @@ class SproutletWrapper : public SproutletTsxHelper
   /// until all these timers have popped or been cancelled.
   std::set<TimerID> _pending_timers;
 
+  // The allowed host state for outbound requests from the sproutlet wrapped by
+  // this wrapper.  If there are no addresses of the appropriate state (e.g.
+  // whitelisted), then a 503 response will be internally generated, and the
+  // error state will be set to indicate that there were no matching addresses.
+  int _allowed_host_state;
+
```
jack-atack: Fairly minor, but the precedent seems to be to use 3 `///` in Sprout header files

PR #1984, opened 2017-10-27
[Reviewer: Krista] Add max rate

eleanor-merry:
This plumbs through a config option for the max token rate to the load monitor.
I've also made the various load monitor options specific to the process type.
Tested in UTs only,

I've got the same change in all the repos - I won't send these to you for review.


eleanor-merry:
The UTs for the load monitor function are in cpp-common-test. For this testing all I've done is make full_test in sprout (which is really just checking that it compiles).

PR #1987, opened 2017-11-01
Send AS timeout error upstream without delay

KristaHyer:
This pull request should fix the issue where:
- The timer for waiting for a response from an AS times out, so the S-CSCF sproutlet tries to send a 408 timeout response upstream.
- The sproutlet proxy holds on to this response because the sproutlet has an outstanding downstream fork, which might produce a better response.
- The sproutlet proxy doesn't signal to the UACTsx that it should stop retrying.
- Eventually the UACTsx finishes its retries (potentially MUCH later) and sends its own response upstream. The sproutlet proxy has now heard back from all forks so sends the best response upstream.

I have edited the code so that the 408 will now be sent upstream without waiting.

I've fully UTed the default handing session terminated, and make full_test passes. However:
- The behavior of the SCSCF doesn't match what might be expected - this will need to be looked into and possibly corrected.
- Default handling of session continued hasn't been fully tested.

As we've spoken about, I'm not fully happy with the design of this fix.
There is because:
- There is a class variable _pending_responses. If this does not equal 0, then the SproutletWrapper will never be deleted.
- I reduce _pending_responses when we give up on a fork, since we are no longer waiting for a response from this fork, and then set a flag on this fork.
- If _pending_responses is ever reduced again, without checking the flag (or if the flag has been edited elsewhere!), _pending_responses may end up equal to -1 (in which case there will be a memory leak as the SproutletWrapper will never be deleted), or incorrectly end up set to 0 (in which case the SproutletWrapper will be incorrectly deleted, which will _cause a crash?_ when the UACTsx later attempts to contact it.
- I believe I've edited all the current places where _pending_responses is reduced, but others could be added in the future.

This feels risky to me.

On a separate note, I haven't added any logs because I didn't really see a place for them, but it's something to look out for.


src/ut/scscf_test.cpp
```diff
@@ -3617,21 +3612,376 @@ TEST_F(SCSCFTest, DefaultHandlingTerminateTimeout)
...
+  ReqMatcher("ACK").matches(out);
+  tpAS1.expect_target(current_txdata(), true);
+  free_txdata();
+
+  // Check error is not forwarded on, as timeout error has already been sent.
+  ASSERT_EQ(0, txdata_count());
 }
 
 
-// Disabled because terminated default handling is broken at the moment.
```
KristaHyer: I was just fixing up another test here (and below).
I think this comment was mistakenly added over confusion over the test name, which is why I removed it.
(I can't find where it was added.)

src/ut/scscf_test.cpp
```diff
@@ -3495,7 +3494,7 @@ TEST_F(SCSCFTest, DefaultHandlingTerminate)
                                    "call",
                                    RegDataXMLUtils::STATE_REGISTERED,
                                    subscription.return_sub());
-  EXPECT_CALL(*_sess_term_comm_tracker, on_failure(_, HasSubstr("408")));
+//  EXPECT_CALL(*_sess_term_comm_tracker, on_failure(_, HasSubstr("408")));
```
KristaHyer: I'm confused over these EXPECT_CALLS, and want to ask you about them at some point.

KristaHyer:
This fix is working, and `make full_test` passes.
I checked `make full_test` passed with the workaround to avoid using expect_target(), then removed the workaround (so that we wouldn't forget to remove it in the future).

src/ut/scscf_test.cpp
AlexHockey: Why have you commented this out?

src/ut/scscf_test.cpp
```diff
@@ -3566,17 +3565,444 @@ TEST_F(SCSCFTest, DefaultHandlingTerminate)
...
+    .addIdentity("sip:6505551000@homedomain")
+    .addIfc(1, {"<Method>INVITE</Method>"}, "sip:1.2.3.4:56789;transport=tcp", 0, 1);
+  SubscriptionBuilder subscription = SubscriptionBuilder()
+    .addServiceProfile(service_profile);
+  _hss_connection->set_impu_result("sip:6505551000@homedomain",
+                                   "call",
+                                   "UNREGISTERED",
+                                   subscription.return_sub());
+
+//  EXPECT_CALL(*_sess_term_comm_tracker, on_failure(_, HasSubstr("timeout")));
```
AlexHockey: Ditto.

src/sproutletproxy.cpp
```diff
@@ -1737,6 +1736,21 @@ void SproutletWrapper::cancel_pending_forks(int st_code, std::string reason)
...
 }
 
+void SproutletWrapper::mark_pending_forks_as_timed_out()
+{
+  for (size_t ii = 0; ii < _forks.size(); ++ii)
+  {
+    if ((_forks[ii].state.tsx_state != PJSIP_TSX_STATE_NULL) &&
+        (_forks[ii].state.tsx_state != PJSIP_TSX_STATE_TERMINATED))
+    {
+      if (_forks[ii].req->msg->line.req.method.id == PJSIP_INVITE_METHOD)
```
AlexHockey: Why are we only doing this on an INVITE? Did you do any tests that used other types of request (e.g. MESSAGE). I think you should, as there may be a bug here.

src/ut/scscf_test.cpp
```diff
@@ -3566,17 +3565,444 @@ TEST_F(SCSCFTest, DefaultHandlingTerminate)
...
+
+  // INVITE passed on to AS.
+  out = current_txdata()->msg;
+  ReqMatcher r1("INVITE");
+  ASSERT_NO_FATAL_FAILURE(r1.matches(out));
+  free_txdata();
+
+  // Advance time without receiving a response. The application server is
+  // bypassed.
+  cwtest_advance_time_ms(6001);
```
AlexHockey: Isn't 6 seconds way longer than we need to wait?

src/ut/scscf_test.cpp
```diff
@@ -3566,17 +3565,444 @@ TEST_F(SCSCFTest, DefaultHandlingTerminate)
...
+    .addIdentity("sip:6505551000@homedomain")
+    .addIfc(1, {"<Method>INVITE</Method>"}, "sip:1.2.3.4:56789;transport=tcp", 0, 1);
+  SubscriptionBuilder subscription = SubscriptionBuilder()
+    .addServiceProfile(service_profile);
+  _hss_connection->set_impu_result("sip:6505551000@homedomain",
+                                   "call",
+                                   "UNREGISTERED",
+                                   subscription.return_sub());
+
+//  EXPECT_CALL(*_sess_term_comm_tracker, on_failure(_, HasSubstr("timeout")));
```
AlexHockey: Ditto. I'm going to stop commenting on these.

src/ut/scscf_test.cpp
```diff
@@ -3566,17 +3565,444 @@ TEST_F(SCSCFTest, DefaultHandlingTerminate)
...
+  // INVITE passed on to AS.
+  // Save off the INVITE, as it is needed later on in the test.
+  out = current_txdata()->msg;
+  ReqMatcher r1("INVITE");
+  ASSERT_NO_FATAL_FAILURE(r1.matches(out));
+  pjsip_tx_data* inv_for_as = pop_txdata();
+
+  // Advance time without receiving a response. The application server is
+  // bypassed.
+  cwtest_advance_time_ms(6001);
```
AlexHockey: Ditto.

src/ut/scscf_test.cpp
```diff
@@ -3566,17 +3565,444 @@ TEST_F(SCSCFTest, DefaultHandlingTerminate)
...
+  // Respond to the AS with a CANCEL, as the timeout error has already been sent
+  // to the caller.
+  ASSERT_EQ(1, txdata_count());
+  out = current_txdata()->msg;
+  ReqMatcher("CANCEL").matches(out);
+  tpAS1.expect_target(current_txdata(), true);
+  free_txdata();
+
+  // Check no other messages are pending.
+  ASSERT_EQ(0, txdata_count());
```
AlexHockey: I think it's worth finishing off the transaction here i.e. having the AS send back a 200 to the CANCEL and a 487 to the INVITE, and check that sprout ACKs the 487.

src/ut/scscf_test.cpp
```diff
@@ -3566,17 +3565,444 @@ TEST_F(SCSCFTest, DefaultHandlingTerminate)
...
+
+  // Advance some more time.
+  cwtest_advance_time_ms(6000);
+  poll();
+
+  // Now the AS finally responds with a 403 (a 4xx error code chosen at random).
+  inject_msg(respond_to_txdata(inv_for_as, 403), &tpAS1);
+
+  // Send an ACK for to the AS for the error.
```
AlexHockey: That sentence doesn't make sense.

src/ut/scscf_test.cpp
```diff
@@ -3566,17 +3565,444 @@ TEST_F(SCSCFTest, DefaultHandlingTerminate)
...
+  ASSERT_EQ(0, txdata_count());
+
+  // Advance some more time.
+  cwtest_advance_time_ms(6000);
+  poll();
+
+  // Now the AS finally responds with a 501 (a 5xx error code chosen at random).
+  inject_msg(respond_to_txdata(inv_for_as, 501), &tpAS1);
+
+  // Send an ACK for to the AS for the error.
```
AlexHockey: Ditto

src/ut/scscf_test.cpp
```diff
@@ -4026,7 +4476,8 @@ TEST_F(SCSCFTest, DefaultHandlingContinueImmediateError)
               testing::MatchesRegex("Route: <sip:1\\.2\\.3\\.4:56789;transport=UDP;lr>\r\nRoute: <sip:odi_[+/A-Za-z0-9]+@127.0.0.1:5058;transport=UDP;lr;service=scscf>"));
 
   // ---------- AS1 immediately rejects the request with a 500 response.  This
-  // gets returned to the caller because the 183 indicated the AS is live.
+  // doesn't get returned to the caller, because no 183 has arrived indicating
```
AlexHockey: Any response will indicate the AS i alive, it doesn't have to be a 183. Can you update the comment please?

include/sproutlet.h
```diff
@@ -301,6 +301,9 @@ class SproutletTsxHelper
   ///
   virtual void cancel_pending_forks(int st_code = 0, std::string reason = "") = 0;
 
+  /// Marks all pending forked requests as timed out.
+  virtual void mark_pending_forks_as_timed_out() = 0;
```
AlexHockey: I don't like the fact we're talking about timed out forks (both here, and in the flag on the fork structure). The sproutlet proxy doesn't know why the sproutlet has given up on the forks, only that it no longer cares about the responses. Please could you call this `abandon_pending_forks` or something similar that is not timeout specific?

include/sproutletproxy.h
```diff
@@ -438,6 +440,8 @@ class SproutletWrapper : public SproutletTsxHelper
     bool pending_cancel;
     int cancel_st_code;
     std::string cancel_reason;
+    bool pending_response;
+    bool timed_out;
```
AlexHockey: Similarly, I would prefer abandoned rather than timed_out here.

src/ut/scscf_test.cpp
KristaHyer: I thought it had to be a 183 from this comment:
https://github.com/Metaswitch/sprout/blob/dev/src/ut/scscf_test.cpp#L4112

AlexHockey: Ah OK. So a 100 will turn off the liveness timer, but will still allow a 500 error to trigger default handling.

include/sproutlet.h
KristaHyer: done

include/sproutletproxy.h
KristaHyer: done

src/ut/scscf_test.cpp
KristaHyer: done

src/ut/scscf_test.cpp
KristaHyer: you're right it doesn't 😛 - fixed

src/ut/scscf_test.cpp
KristaHyer: fixed

src/ut/scscf_test.cpp
KristaHyer: I've tried to make this comment clearer.

src/sproutletproxy.cpp
KristaHyer: Good spot! I've removed this check, and added tests for MESSAGEs.

KristaHyer:
I've made the changes you suggested, and confirmed `make full_test` still passes.
Passing this back for re-review.

AlexHockey:
Looks good to me :-)

PR #1988, opened 2017-11-02
[Reviewer Sathiyan] Treat orig user=phone uri as tel

steelyvoid:
Hi Sathiyan

Are you OK to rereview the "Treat 'user=phone' URIs as though they are Tel URIs" fix you reviewed for me a couple of weeks back?  At the time, we were just going to deliver it to the customer in 11.0.2 and get them to use MMF to achieve the same end in 11.1, but the customer pushed back (they were worried about the implied perf impact of MMF) and AJH reconsidered the previous decision and agreed that it should go into PC/dev for inclusion in 11.1.

The code is as it was when you reviewed it, with the following exceptions

-  Alex didn't like calling the flag "...._conv" as that implied we were converting the value in the request (not just interpreting it as a tel URI for originating purposes), so I changed it (and various comments that talked about "conversion") to "...._coerce", which was his favoured term for what we're doing.
-  Alex and I agreed that we shouldn't consider sip:<DN>:<domain> URIs (i.e. **without** ";user=phone") as a Tel URI even if enforce_user_phone is false, because that is too likely to break existing customer usage (quite a lot of customers are likely to have both enforce_user_phone off and subscribers with <DN>@<domain> IMPUs and we don't want them all to suddenly fail to REGISTER because we're looking for tel:<DN>).  Therefore, I left the prefer_sip flag as the default (true) at line https://github.com/Metaswitch/sprout/blob/8aff86a20eef2d1e08b9877b198f9b8a340f9a22/src/pjutils.cpp#L306.


sathiyan-sivathas:
Hi Steve,

I should have time to review this tomorrow. Is that OK?

steelyvoid:
That would be great – thanks!

steelyvoid:
Sorry Sathiyan - did you mean, have a test with user=phone **present** in the PAI and enforce_global_only_lookups true?  In this case, the test user will be treated as a local phone number and the coercion won't happen.

If user=phone isn't present in the PAI, the coercion won't happen whether or not enforce_global_only_lookups is set.

sathiyan-sivathas:
We discussed what I meant here. The confusion was because I wrote down the wrong config option! I meant enforce_user_phone not enforce_global_only_lookups.

steelyvoid:
Hi Sathiyan.  I've managed to bork my fix branch during an attempt to back merge dev, so I'm going to nuke it and start again.   I'll close this PR and point the new one at it (which I'll merge, as I've done your last comment above) so we don't lose your comments

steelyvoid:
Hi Sathiyan.  I've managed to bork my fix branch during an attempt to back merge dev, so I'm going to nuke it and start again.   I'll close this PR and point the new one at it (which I'll merge, as I've done your last comment above) so we don't lose your comments

PR #1989, opened 2017-11-03
[Reviewer: Adam] 250 worker threads per core

sebrexmetaswitch:
Default to 250 sprout worker threads per core, because the threads are blocking.

I've live tested that this defaults correctly, and we've already live tested that this number of threads fixes the latency issues.


johadalin:
I wanted to make an emoji pun here, but there's no seal emoji. so you get this :whale: 

Looks good :+1:

PR #1991, opened 2017-11-03
[Reviewer: Sathiyan] Add SAS logs, fix crash

eleanor-merry:
This PR adds three SAS logs that give more context when a NOTIFY is sent, fixes up the SAS logging for admin requests, and fixes a crash when getting the subscriptions from the remote store fails.

I've not added a UT for the crash yet - this is because I'm doing a wider fix up of the UTs, and I wanted to get a fix for the crash into dev sooner as it's blocking some testing.


src/subscriptionsproutlet.cpp
```diff
@@ -598,24 +590,35 @@ Store::Status SubscriptionSproutletTsx::update_subscription_in_stores(
...
+      if (remote_aor_pair == NULL)
       {
-        TRC_DEBUG("Hit data contention attempting to write AoR %s to remote store", aor.c_str());
-        remote_aor_pair = read_and_cache_from_store(sdm, aor, _cached_aors);
-        if (remote_aor_pair == NULL)
-        {
-          // LCOV_EXCL_START We test behaviour on store error elsewhere,
-          // and UT-ing this case is more effort than it's worth
+        // LCOV_EXCL_START
+        TRC_DEBUG("Hit an error reading AoR %s from the remote store, unable to update subscription %s",
```
sathiyan-sivathas: Just to check - you're going to remove this exclusion when you add UTs?

eleanor-merry: Yup.

PR #1996, opened 2017-11-07
[Reviewer: CJR] Fix Tel URI bug

huws:
Fix for the crash we've seen when a request without a SIP routing URI is routed between `CompositeSproutletTsx`s within a network function.  Specifically:
 - There is no Route header on the request
 - The Request URI is a Tel URI
 - The network function has been invoked based on a port-match

In such circumstances, we were passing in a Tel URI as though it were a SIP URI, and crashing when it turned out not to be.  Now we protect against this, and use the root URI, if no others are available.


src/ut/sproutletproxy_test.cpp
```diff
@@ -592,6 +592,28 @@ class FakeSproutletTsxNextHop : public CompositeSproutletTsx
...
+{
+public:
+  FakeSproutletURIForwarder(Sproutlet* sproutlet) :
+   CompositeSproutletTsx(sproutlet, static_cast<FakeSproutlet<FakeSproutletURIForwarder>*>(sproutlet)->_next_hop)
+  {
+  }
+
+  void on_rx_initial_request(pjsip_msg* req)
+  {
+    // TODO HJS CONT
```
ChrisRees: should this still be here?
Appreciate it's test code, but we shouldn't have TODO flags.

huws: Definitely shouldn't be there - thanks for spotting!

PR #1997, opened 2017-11-07
[Reviewer: AJH] Fork error SAS log

metaswitch-ak10:
Hi Alex, could you please review this fix?

This fix is for [Issue 514494](http://sfr/prodAsp/scripts/MSG/Issue.asp?issue_id=514494&index=2). I have added a SAS log in the case of a fork error that reports the response code being send (Sprint complained in the original issue about the 503 response code appearing out of blue in SAS logs such as "Received downstream 503 response, generating and returning 500 response upstream" - see issue diags). 

Testing: 
I have managed to reproduce the original problem by calling a number at a made-up domain corresponding to nonexistent IP address. When the BGCF tries to forward the request to that IP address the following SAS logs are made:
- "Fork 0 failed with error TRANSPORT_ERROR - sending response code 503" in case of TCP being used
- "Fork 0 failed with error TIMEOUT - sending response code 408" in case of UDP being used


src/sproutletproxy.cpp
```diff
@@ -1975,6 +1975,13 @@ void SproutletWrapper::rx_fork_error(ForkErrorState fork_error, int fork_id)
                                                   NULL,
                                                   &rsp);
 
+    // SAS log the error and response
+    SAS::Event event(trail(), SASEvent::RX_FORK_ERROR, 0);
+    event.add_static_param(fork_id);
+    event.add_static_param(fork_error);
```
AlexHockey: Could you add a comment to the definition of `ForkErrorState` to warn people that if they change it they also need to update the enum in the resource bundle (though remember that the resource bundle is private, so don't include a link to it in the code for example).

metaswitch-ak10: OK, no problem

metaswitch-ak10:
Deleting this branch due to a problem with git. Created replacement branch called "bgcf_sas_log" and cherry-picked all commits.

PR #2001, opened 2017-11-08
[Reviewer: Richard] memory leaks in basicproxy

sebrexmetaswitch:
`pjsip_tsx_send_msg` doesn't decrease the reference count on the `tdata`if it fails to send it. We therefore need to check the return code of `pjsip_tsx_send_msg` and decrease the reference count if it fails.

There are some old comments which suggest we deliberately don't do this because we expect a failure to send a message to trigger a call to `on_tsx_state` with a transport error. However, our `on_tsx_state` function doesn't then decrease the reference count, so we were leaking the `tdata`.

A couple of thoughts on this:

1) I don't think that `on_tsx_state` *can* safely decrease the reference count, because it can also get called if we successfully send the data but then get a TCP reset on that connection. In that case, `pjsip_tsx_send_msg` will have already decreased the reference count for us, and so we must not try to do it in `on_tsx_state`.

2) Furthermore, if we do fail to send the data, the call to `on_tsx_state` happens on the sending thread, *before* `pjsip_tsx_send_msg` returns, so we can be safe in the knowledge that our `tdata` hasn't been `dec_ref`ed already when we enter `on_tsx_state` in this scenario

So I'm pretty sure that decreasing the reference count after failing to send is the correct fix.

I've managed to test the changes at line 1277 and 1758 in `basicproxy.cpp` and the one in `sproutletproxy.cpp` live, by generating load that's routing to a non-existent BGCF and then killing the load generator, and can confirm that they don't cause crashes.

The others I haven't managed to hit, but am fairly confident that decreasing the ref count is correct.


sebrexmetaswitch:
As discussed, I'm pretty sure there are more of these.

I added some extra logs to PJUtils, and think that we're leaking some of those too, but I don't think it's worth tracking under this PR as I'm more convinced that these ones are safe

src/sproutletproxy.cpp
```diff
@@ -1090,7 +1090,13 @@ void SproutletProxy::UASTsx::tx_response(SproutletWrapper* downstream,
       int st_code = rsp->msg->line.status.code;
       set_trail(rsp, trail());
       on_tx_response(rsp);
-      pjsip_tsx_send_msg(_tsx, rsp);
+      pj_status_t status = pjsip_tsx_send_msg(_tsx, rsp);
+
+      if (status != PJ_SUCCESS)
+      {
+        // pjsip_tsx_send_msg only decreases the ref count on success
```
richardwhiuk: Be nice to do a:

```cpp
        TRC_INFO("Failed to send UAS TSX message:",
                 PJUtils::pj_status_to_string(status).c_str());
```

here

PR #2007, opened 2017-11-09
Add checks for rph.json config file

KristaHyer:
In this pull request I have added a schema for the new rph.json config file.
I have also added the reload script to be run when rph.json is updated.
Finally, I have added a validation script, which compares against the schema, and does additional validation.

- This script is in python, but is not very pythonic. That is because it was based off other validation files,
 such as [this one](https://github.com/Metaswitch/clearwater-etcd/blob/dev/clearwater-config-manager.root/usr/share/clearwater/clearwater-config-manager/scripts/validate_json.py), which are not pythonic either

There is significant overlap between the new file I have added, and the existing [validate_json.py](https://github.com/Metaswitch/clearwater-etcd/blob/dev/clearwater-config-manager.root/usr/share/clearwater/clearwater-config-manager/scripts/validate_json.py) file. Because of their non-pythonic nature, I couldn't import anything from the original file to reduce the duplication. If you think that this duplication is an issue, it should be noted down as technical debt. Just let me know.

Further details on the new config are in pull request https://github.com/Metaswitch/clearwater-etcd/pull/535.


sprout-base.root/usr/share/clearwater/clearwater-config-manager/scripts/config_validation/rph_validation.py
```diff
@@ -0,0 +1,183 @@
...
+            else:
+                parsed_config.update({header:priority})
+
+# Check the priorites are set in a way which is valid.
+# A higher priority header cannot be given a lower priority than a lower priority
+# header (ie. wps.0 must be higher proirity that wps.1).
+for header_list in HEADERS_LISTS:
+    headers_to_check = HEADERS_LISTS[header_list]
+    place_in_list = 0
+    for header in headers_to_check:
```
sathiyan-sivathas: I think you have more recursion than you need from this point down. I think what this code does is:
 1) Loops through the header values in a namespace.
 2) If a value is present, loop through the rest of the values to check they have higher priorities.

Once you've done 2) once, you can break out of 1) otherwise you're doing the same work again. Alternatively, in 2) you could just check that the next value has higher priority. Then 1) will handle the rest of the list.

PR #2011, opened 2017-11-13
Sas for iFC match

Ying-metaswitch:
Together with https://github.com/Metaswitch/clearwater-sas-resources/pull/245, this PR improves SAS trace for iFC match. This is based on request from https://github.com/Metaswitch/clearwater-issues/issues/2511

1. Adds `IFC_UNUSUAL` to log unusual but still valid iFCs. Currently this is used when RequestURI starts with `sip:/tel:` rather than containing only the hostport. The SPT won't be skipped for matching.
2. Renames the function `ifc_invalid()` to `ifc_error()` to handle raise SAS event for both invalid and unusual iFCs. This function no longer calls to `xml_error()` which merely raises a different SAS event for the same error. 
3. Adds details to iFC match/not match SAS event about how each SPT and group is matched.

NB: I would like to remove `the service triggers evaluated to false` from below as corresponding phrase isn't present when iFC matches, and sounds a bit obscure.
Originally I wrote an UT for unusual iFC but it doesn't test much, so I didn't keep it.

# Example output 1
The iFC for sip:match.yh.cw-ngv.com matches
> For each group, AND all SPT condition match results.
  SPT in group 0 is not matched.
  SPT in group 0 is matched.
  SPT in group 1 is matched.
Overall result group 0 is not matched.
Overall result group 1 is matched.
>
> OR result from all group to determine overall iFC match.

# Example output 2
This iFC for sip:unusual.yh.cw-ngv.com is unusual.
> Request URI service point trigger should include hostport only

The iFC for sip:unusual.yh.cw-ngv.com does not match: the service triggers evaluated to false
> For each group, AND all SPT condition match results.
  SPT in group 0 is not matched.
Overall result group 0 is not matched.
>
> OR result from all group to determine overall iFC match.


src/ifc.cpp
```diff
@@ -305,14 +304,22 @@ bool Ifc::spt_matches(const SessionCase& session_case,  //< The session case
       test_string = hostport;
     }
 
-    req_uri_regex = boost::regex(XMLUtils::get_text_or_cdata(node),
+    std::string req_uri = XMLUtils::get_text_or_cdata(node);
+    if ((req_uri.compare(0, 4, "sip:") == 0) ||
+        (req_uri.compare(0, 4, "tel:") == 0))
+    {
+      ifc_error("Request URI service point trigger should include hostport only",
+                  server_name, SASEvent::IFC_UNUSUAL, 0, trail);
```
AlexHockey: This text is inaccurate. The Request URI should be a *regex* that matches either on the hostport  (for a SIP URI) or a telephone number (for a Tel URI).

src/ifc.cpp
```diff
@@ -493,63 +500,106 @@ bool Ifc::filter_matches(const SessionCase& session_case,
...
+
+        if (spt_matched)
+        {
+          ifc_match_result.append("  SPT in group ").append(std::to_string(group_id)).append(" is matched.\n");
+        }
+        else
         {
-          groups[group] = val;
+          ifc_match_result.append("  SPT in group ").append(std::to_string(group_id)).append(" is not matched.\n");
+        }
```
AlexHockey: This if block could be condensed down into one statement.

src/ifc.cpp
```diff
@@ -493,63 +500,106 @@ bool Ifc::filter_matches(const SessionCase& session_case,
...
+        int32_t group_id = XMLUtils::parse_integer(group_node,
+                                                   "Group ID",
+                                                   0,
+                                                   std::numeric_limits<int32_t>::max());
+        TRC_DEBUG("This SPT in group %d is %s", (int)group_id,
+                  spt_matched ? "matched" : "not matched");
+
+        if (spt_matched)
+        {
+          ifc_match_result.append("  SPT in group ").append(std::to_string(group_id)).append(" is matched.\n");
```
AlexHockey: It feels like it would be useful to put the SPT index in here.

src/ifc.cpp
```diff
@@ -493,63 +500,106 @@ bool Ifc::filter_matches(const SessionCase& session_case,
...
+      TRC_DEBUG("Overall result group %d is %s", (int)group->first,
+                group->second ? "matched" : "not matched");
+      if (group->second)
+      {
+        ifc_match_result.append("Overall result group ").append(std::to_string(group->first)).append(" is matched.\n");
+      }
+      else
+      {
+        ifc_match_result.append("Overall result group ").append(std::to_string(group->first)).append(" is not matched.\n");
+      }
```
AlexHockey: Ditto - could be one statement.

src/ifc.cpp
```diff
@@ -493,63 +500,106 @@ bool Ifc::filter_matches(const SessionCase& session_case,
...
+    if (cnf)
+    {
+      ifc_match_result.append("\nAND result from all group to determine overall iFC match.\n");
+    }
+    else
     {
-      TRC_DEBUG("Result group %d val %s", (int)it->first, it->second ? "true" : "false");
-      ret = cnf ? (ret && it->second) : (ret || it->second);
+      ifc_match_result.append("\nOR result from all group to determine overall iFC match.\n");
     }
```
AlexHockey: Again, could be one statement.

AlexHockey: Also, should we TRC_* the `ifc_match_result`?

Ying-metaswitch: Yep removed individual TRC_ and just TRC the `ifc_match`.

src/ifc.cpp
Ying-metaswitch: I couldn't find such an index? Can't get it from looping either - one SPT from group 1 may be followed by an SPT from group 0.

src/ifc.cpp
Ying-metaswitch: Condensed accordingly.

src/ifc.cpp
```diff
@@ -111,7 +110,7 @@ bool Ifc::spt_matches(const SessionCase& session_case,  //< The session case
 
   if (!node)
   {
-    invalid_ifc("Missing class for service point trigger", server_name, SASEvent::IFC_INVALID, 0, trail);
+    ifc_error("Missing class for service point trigger", server_name, SASEvent::IFC_INVALID, 0, trail);
```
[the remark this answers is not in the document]
Ying-metaswitch: Yep I shouldn't have changed that logic. `handle_invalid_ifc` will continue to throw `xml_error`, while `handle_unusal_ifc` won't.

include/ifc.h
```diff
@@ -83,28 +83,28 @@ class Ifc
       rapidxml::xml_document<>* ifc_doc);
 
   bool filter_matches(const SessionCase& session_case,
-                      bool is_registered,
-                      bool is_initial_registration,
+                      const bool& is_registered,
```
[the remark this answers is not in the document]
Ying-metaswitch: It wasn't motivated by saving memory, but rather to reinforce the point that this bool isn't going to be changed in the function. Passing by value wouldn't change the value of this bool in calling function either, but it seems less apparent in style?

Ying-metaswitch:
Applied markups. Attached here is two SAS trace: for ifc unusual and ifc match, for ifc invalid and ifc mismatch.  \\ad\csd\bigfiles\clearwater\issue-attachments\clearwater-issues\2511
Note that `handle_invalid_ifc` raises `xml_error` and thus does SAS logging twice. The reason for keeping this is that XMLUtils raise `xml_error`, so we have to leave get `xml_error` to do SAS logging. Yet those invalid iFC that our code catches will have more specific SAS event, and we do that logging in `handle_invalid_ifc` itself. 

`make full_test` passed.

src/ifc.cpp
AlexHockey: Right but if I have two SPT in a group, and one matches and one doesn't how do I tell which one has matched?

include/ifc.h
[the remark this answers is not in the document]
AlexHockey: Are you trying to convey that the value won't be changed in the function, making the parameter const (i.e. not a reference) would be sufficient and IMHO more idiomatic.

src/ifc.cpp
Ying-metaswitch: It's printed out in the order of appearance in the iFC, so when the person looks at the iFC it should be apparent enough.

src/ifc.cpp
```diff
@@ -493,78 +508,96 @@ bool Ifc::filter_matches(const SessionCase& session_case,
     // we AND all the groups together. In DNF we do the converse.
     std::map<int32_t, bool> groups;
 
+    std:: string spt_relation = cnf ? "OR" : "AND";
+    std:: string group_relation = cnf ? "AND" : "OR";
```
AlexHockey: Nice! Very clean.

Ying-metaswitch:
Yep I used `xml_error` to catch all syntactic error and raise SAS event `INVALID_IFC_IGNORED` (which I want to rename to `INVALID_XML_IGNORED`); while `handle_invalid_ifc` catches semantic error and throws `ifc_error` to quit processing, but only raise SAS event `IFC_INVALID` once. See the screenshot in folder for only one SAS event. I couldn't export the SAS possibly because the AS is missing in the iFC. 

I would like to remove `IFC_INVALID_NOAS` and just use `IFC_INVALID`. It doesn't look like there is anything special about them, the error will be recorded in SAS param.

include/ifc.h
```diff
@@ -83,28 +83,47 @@ class Ifc
...
+class ifc_error : public std::exception
+{
+public:
+  ifc_error()
+  {
+  }
+
+  virtual ~ifc_error() throw ()
+  {
+  }
```
AlexHockey: These are the default implementions (basically) so you should be able to omit them. So the class is just 

```
class ifc_error : public std::exception {};
```

🙂

Ying-metaswitch: That's much clearer, thanks!

Ying-metaswitch:
In the end it's `INVALID_IFC_IGNORED`, `INVALID_XML_IGNORED` and `IFC_UNUSUAL` to log each case only once. `make full test` passed.

PR #2012, opened 2017-11-14
[Reviewer: Krista] Check that the IMPUs are valid

eleanor-merry:
We support request URIs that are URN uris, but only if we never have to use them for routing/making subscriber requests (they should always be handled by an application server).

This PR tidies up the diagnostics/behaviour when we receive requests with URN URIs that don't get sent to any application servers.

I've tested the mainline case live, and there's UTs.


src/ut/icscfsproutlet_test.cpp
```diff
@@ -3738,3 +3738,125 @@ TEST_F(ICSCFSproutletTest, ICSCFHandlesUrnUri)
...
+  // Inject a SIP MESSAGE.
+  Message msg1;
+  msg1._first_hop = true;
+  msg1._method = "MESSAGE";
+  msg1._requri = "urn:service:sos";
+  msg1._full_to_header = "To: <urn:service:sos>";
+  msg1._route = "Route: <sip:homedomain>";
+  inject_msg(msg1.get_request(), tp);
+
+  // Check that the MESSAGE is rejected
```
KristaHyer: I think this is rejected since there was no data found in the HSS?
Can you add a comment to make this clear?

src/ut/icscfsproutlet_test.cpp
```diff
@@ -3738,3 +3738,125 @@ TEST_F(ICSCFSproutletTest, ICSCFHandlesUrnUri)
...
+                                        ICSCF_PORT,
+                                        "1.2.3.4",
+                                        49152);
+
+  // Inject an INVITE request to a tel URI with a P-Served-User header.
+  Message msg1;
+  msg1._first_hop = true;
+  msg1._method = "MESSAGE";
+  msg1._toscheme = "tel";
+  msg1._to = "+1661000000";
```
KristaHyer: I think we have a rule set up where this will be replaced by urn:services:sos, which is invalid so can't be parsed, leading to the rejection.
Can you add a comment to make this clear?

src/ut/icscfsproutlet_test.cpp
```diff
@@ -3738,3 +3738,125 @@ TEST_F(ICSCFSproutletTest, ICSCFHandlesUrnUri)
...
+  // Create a TCP connection to the I-CSCF listening port.
+  TransportFlow* tp = new TransportFlow(TransportFlow::Protocol::TCP,
+                                        ICSCF_PORT,
+                                        "1.2.3.4",
+                                        49152);
+
+  // Inject a SIP MESSAGE.
+  Message msg1;
+  msg1._first_hop = true;
+  msg1._method = "REGISTER";
```
KristaHyer: Hmm I'm not sure I can work out why this is being rejected.
Is it just because this is a REGISTER, not a MESSAGE?
Can you add a comment please? :)

PR #2013, opened 2017-11-14
[Reviewer: Krista] Fix up uninteresting calls

eleanor-merry:
Fixes up the various uninteresting calls in the UTs

### DIFF
--- a/src/ifc.cpp
+++ b/src/ifc.cpp
@@ -576,7 +576,6 @@ bool Ifc::filter_matches(const SessionCase& session_case,
 AsInvocation Ifc::as_invocation() const
 {
   xml_node<>* as = _ifc->first_node(RegDataXMLUtils::APPLICATION_SERVER);
-  pj_assert(as != NULL);
 
   AsInvocation as_invocation;
   as_invocation.server_name = XMLUtils::get_first_node_value(as, RegDataXMLUtils::SERVER_NAME);
--- a/src/scscfselector.cpp
+++ b/src/scscfselector.cpp
@@ -279,8 +279,7 @@ std::string SCSCFSelector::get_scscf(const std::vector<int> &mandatory,
   // There are multiple S-CSCFs that match on all mandatory capabilities, the highest number of optional
   // capabilities, and the highest priority. Select one using a weighted random choice.
   srand(time(NULL));
-  int random;
-  random = rand() % sum;
+  int random = (sum != 0) ? rand() % sum : 0;
 
   int index = 0;
   int accumulator = matches[index].weight;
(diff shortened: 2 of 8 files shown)


src/scscfselector.cpp
```diff
@@ -279,8 +279,7 @@ std::string SCSCFSelector::get_scscf(const std::vector<int> &mandatory,
   // There are multiple S-CSCFs that match on all mandatory capabilities, the highest number of optional
   // capabilities, and the highest priority. Select one using a weighted random choice.
   srand(time(NULL));
-  int random;
-  random = rand() % sum;
+  int random = (sum != 0) ? rand() % sum : 0;
```
KristaHyer: So if `sum` is 0 does that just mean there are no valid S-CSCFs? So there is no point "selecting one"?

src/ut/handlers_test.cpp
```diff
@@ -398,9 +405,7 @@ TEST_F(DeregistrationTaskTest, ImpiClearedWhenBindingUnconditionallyDeregistered
   expect_sdm_updates(aor_ids, aors);
 
   // The corresponding IMPI is also deleted.
-  ImpiStore::Impi* impi = new ImpiStore::Impi("impi1");
-  EXPECT_CALL(*_local_impi_store, get_impi("impi1", _, false)).WillOnce(Return(impi));
-  EXPECT_CALL(*_local_impi_store, delete_impi(impi, _)).WillOnce(Return(Store::OK));
+  expect_gr_impi_deletes("impi1");
```
KristaHyer: Why is this gr? As far as I can see they are only deleting from the `_local_impi_store`?
Did we just used to ignore the call to the remote store?
(Same question for next block of changed code.)

src/ifc.cpp
```diff
@@ -576,7 +576,6 @@ bool Ifc::filter_matches(const SessionCase& session_case,
 AsInvocation Ifc::as_invocation() const
 {
   xml_node<>* as = _ifc->first_node(RegDataXMLUtils::APPLICATION_SERVER);
-  pj_assert(as != NULL);
```
KristaHyer: Why don't we `return` if it's NULL?
I can see why we don't want to assert it, but I don't see why we now continue through the rest of the function collecting info about an AS that doesn't exist?

src/ut/handlers_test.cpp
eleanor-merry: Yup - so there was an uninteresting call made in the UTs

src/ifc.cpp
eleanor-merry: pj_asserts don't give us any protection (if as is NULL, then we'll still crash), and I don't think it will give us a particularly useful error message in this case. It's also not possible for as to be NULL in this case.

PR #2014, opened 2017-11-14
[Reviewer: Alex] Resolve problems running at load with latency

richardwhiuk:
Alex,

This resolves two problems involved running at load with latency:

- It's very difficult to work out what is responsible for causing overload to kick in, because we don't have anything logging long requests. I've added some logs, which based on some runs I've done, I've logged requests which take 50 times are target latency. Testing showed that 10 times was too little.

- We don't correctly assign the stopwatch to the IO Hook, which negates it's purpose.

I've tested this live, and it resolves these two problems

### DIFF
--- a/src/thread_dispatcher.cpp
+++ b/src/thread_dispatcher.cpp
@@ -168,8 +168,8 @@ bool process_queue_element()
       pjsip_rx_data* rdata = qe.event_data.rdata;
 
       // Create an IO hook that pauses the stopwatch while blocked on IO.
-      Utils::IOHook io_hook(std::bind(pause_stopwatch, qe.stop_watch, std::placeholders::_1),
-                            std::bind(resume_stopwatch, qe.stop_watch, std::placeholders::_1));
+      Utils::IOHook io_hook(std::bind(pause_stopwatch, std::ref(qe.stop_watch), std::placeholders::_1),
+                            std::bind(resume_stopwatch, std::ref(qe.stop_watch), std::placeholders::_1));
 
       if (rdata)
       {
--- a/src/thread_dispatcher.cpp
+++ b/src/thread_dispatcher.cpp
@@ -124,30 +124,29 @@ static void resume_stopwatch(Utils::StopWatch& s, const std::string& reason)
 
 static void dump_message_details(pjsip_rx_data* rdata)
 {
-  if (rdata != NULL)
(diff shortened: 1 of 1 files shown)


src/thread_dispatcher.cpp
```diff
@@ -197,19 +228,7 @@ bool process_queue_element()
...
-                        ((pjsip_cid_hdr*)rdata->msg_info.cid)->id.ptr);
-            }
-            if (rdata->msg_info.cseq != NULL)
-            {
-              TRC_ERROR("Exception CSeq: %ld %.*s (maybe)",
-                        ((pjsip_cseq_hdr*)rdata->msg_info.cseq)->cseq,
-                        ((pjsip_cseq_hdr*)rdata->msg_info.cseq)->method.name.slen,
-                        ((pjsip_cseq_hdr*)rdata->msg_info.cseq)->method.name.ptr);
-            }
+            dump_message_details(rdata);
```
AlexHockey: The old logs talked about exceptions but the new ones don't. Should we have a log at the start of this branch saying "Hit exception in worker thread - message details to follow" of something?

src/thread_dispatcher.cpp
```diff
@@ -122,12 +122,43 @@ static void resume_stopwatch(Utils::StopWatch& s, const std::string& reason)
 }
 // LCOV_EXCL_STOP
 
+static void dump_message_details(pjsip_rx_data* rdata)
+{
+  if (rdata != NULL)
+  {
+    if (rdata->msg_info.cid != NULL)
+    {
+      TRC_ERROR("Call-Id: %.*s (maybe)",
```
AlexHockey: The "maybe" is a hang over from when these were in the exception path. Now that this is a subfunction, the call ID and cseq are definitive aren't they?

src/thread_dispatcher.cpp
```diff
@@ -122,12 +122,43 @@ static void resume_stopwatch(Utils::StopWatch& s, const std::string& reason)
 }
 // LCOV_EXCL_STOP
 
+static void dump_message_details(pjsip_rx_data* rdata)
+{
+  if (rdata != NULL)
```
AlexHockey: Bit weird that a function that takes an rdata and operates on an rdata silently does nothing if an rdata isn't passed in.

PR #2015, opened 2017-11-16
No IMS Subscription on Homestead request

Ying-metaswitch:
This PR fixes the issue that Sprout treats failure to find IMS Subscription in homestead XML as a SERVER_ERROR, in cases where IMS Subscription isn't expected. The work includes:
1. When Sprout gets/updates registration state in HSS via homestead, it may encounter two types of error: the connection to HSS itself failed, or the XML it gets back is invalid. So the functions are refactored to make it clear which step is failing.
2. A flag `IMSExpected` means the absence of IMSSubscription in the homestead XML should be treated as failure.  This flag is set based on `req_type` on a http PUT to homestead, and false on GET as we can't be sure if IMS will be present there or not. In `decode_homestead_xml`, always look for IMS. If there is no IMS, stop processing; return failure if IMSExpected, and success otherwise. Once the IMS is found, error in parsing it should be treated as failure. 

Screenshot of SAS call flow and timeline attached (couldn't export for some reason). In call flow, after 403 and an SAR for authentication-timeout, homestead is sending back 200 OK without IMS. This is the scenario we are testing. In timeline, Sprout now sends back 200 for this authentication-timeout, rather than incorrectly sending back a 500. 
fixed_sas_screenshot_timeline
fixed_sas_screenshot

### DIFF
--- a/src/hssconnection.cpp
+++ b/src/hssconnection.cpp
@@ -1,4 +1,4 @@
- /**
+/**
  * @file hssconnection.cpp HSSConnection class methods.
  *
  * Copyright (C) Metaswitch Networks 2017
@@ -318,7 +318,7 @@ bool decode_homestead_xml(const std::string& public_id,
                           HSSConnection::irs_info& irs_info,
                           const std::shared_ptr<rapidxml::xml_document<> > root,
                           SIFCService* sifc_service,
-                          const bool& IMSPresent,
+                          const bool& IMSExpected,
                           SAS::TrailId trail)
 {
   if (!root.get())
@@ -354,48 +354,56 @@ bool decode_homestead_xml(const std::string& public_id,
 
   irs_info._regstate = reg->value();
 
-  if ((IMSPresent) || (irs_info._regstate != RegDataXMLUtils::STATE_NOT_REGISTERED))
+  rapidxml::xml_node<>* prev_reg = cw->first_node(RegDataXMLUtils::PREVIOUS_REGISTRATION_STATE);
+  irs_info._prev_regstate = (!prev_reg) ? "" : prev_reg->value();
+
+  rapidxml::xml_node<>* imss = cw->first_node(RegDataXMLUtils::IMS_SUBSCRIPTION);
+
+  // IMS Subscription element may not be present on some request.
+  if (!imss)
   {
-    printf("\nfinding ims\n");
-    rapidxml::xml_node<>* prev_reg = cw->first_node(RegDataXMLUtils::PREVIOUS_REGISTRATION_STATE);
-    irs_info._prev_regstate = (!prev_reg) ? "" : prev_reg->value();
-
-    rapidxml::xml_node<>* imss = cw->first_node(RegDataXMLUtils::IMS_SUBSCRIPTION);
-
-    if (!imss)
+    std::string sp_str;
+    rapidxml::print(std::back_inserter(sp_str), *root, 0);
+    if (IMSExpected)
     {
-      std::string sp_str;
-      rapidxml::print(std::back_inserter(sp_str), *root, 0);
       TRC_WARNING("Malformed HSS XML for %s - no IMSSubscription element:\n%s",
                   public_id.c_str(),
                   sp_str.c_str());
       return false;
     }
-
-    if (!SproutXmlUtils::parse_ims_subscription(public_id,
-                                                root,
-                                                imss,
-                                                irs_info._service_profiles,
-                                                irs_info._associated_uris,
-                                                irs_info._aliases,
-                                                sifc_service,
-                                                trail))
+    else
     {
-      std::string sp_str;
-      rapidxml::print(std::back_inserter(sp_str), *root, 0);
-      TRC_WARNING("Malformed HSS XML for %s:\n%s",
-                  public_id.c_str(),
-                  sp_str.c_str());
-      return false;
-    }
-
-    rapidxml::xml_node<>* charging_addrs_node = cw->first_node("ChargingAddresses");
-
-    if (charging_addrs_node)
-    {
-      parse_charging_addrs_node(charging_addrs_node, irs_info._ccfs, irs_info._ecfs);
+      TRC_INFO("In HSS XML for %s, there is no IMSSubscription element:\n%s",
+               public_id.c_str(),
+               sp_str.c_str());
+      return true;
     }
   }
+
+  // If IMS Subscription is present, failure to parse it indicate real error.
+  if (!SproutXmlUtils::parse_ims_subscription(public_id,
+                                              root,
+                                              imss,
+                                              irs_info._service_profiles,
+                                              irs_info._associated_uris,
+                                              irs_info._aliases,
+                                              sifc_service,
+                                              trail))
+  {
+    std::string sp_str;
+    rapidxml::print(std::back_inserter(sp_str), *root, 0);
+    TRC_WARNING("Malformed HSS XML for %s:\n%s",
+                public_id.c_str(),
+                sp_str.c_str());
+    return false;
+  }
+
+  rapidxml::xml_node<>* charging_addrs_node = cw->first_node("ChargingAddresses");
+  if (charging_addrs_node)
+  {
+    parse_charging_addrs_node(charging_addrs_node, irs_info._ccfs, irs_info._ecfs);
+  }
+
   return true;
 }
 
@@ -469,9 +477,8 @@ HTTPCode HSSConnection::put_homestead_xml(const irs_query& irs_query,
 
   if (http_code != HTTP_OK)
   {
-    // We have either not found
-    // the subscriber on the HSS or been unable to communicate with
-    // the HSS successfully.
+    // We have either not found the subscriber on the HSS, or been unable to 
+    // communicate with the HSS successfully.
     TRC_ERROR("Could not get subscriber data from HSS");
   }
   return http_code;
@@ -491,11 +498,11 @@ HTTPCode HSSConnection::update_registration_state(const irs_query& irs_query,
   {
     bool IMSPresent = determine_ims_present(irs_query._req_type);
     http_code = decode_homestead_xml(irs_query._public_id,
-                              irs_info,
-                              root,
-                              _sifc_service,
-                              IMSPresent,
-                              trail) ? HTTP_OK : HTTP_SERVER_ERROR;
+                                     irs_info,
+                                     root,
+                                     _sifc_service,
+                                     IMSPresent,
+                                     trail) ? HTTP_OK : HTTP_SERVER_ERROR;
   }
   return http_code;
 }
@@ -512,12 +519,12 @@ HTTPCode HSSConnection::get_registration_data(const std::string& public_id,
   HTTPCode http_code = get_homestead_xml(public_id, root, trail);
   if (http_code == HTTP_OK)
   {
-  http_code = decode_homestead_xml(public_id,
-                              irs_info,
-                              root,
-                              _sifc_service,
-                              false,
-                              trail) ? HTTP_OK : HTTP_SERVER_ERROR;
+    http_code = decode_homestead_xml(public_id,
+                                     irs_info,
+                                     root,
+                                     _sifc_service,
+                                     false,
+                                     trail) ? HTTP_OK : HTTP_SERVER_ERROR;
   }
   return http_code;
 }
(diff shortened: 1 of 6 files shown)


src/hssconnection.cpp
```diff
@@ -309,16 +309,20 @@ void parse_charging_addrs_node(rapidxml::xml_node<>* charging_addrs_node,
   }
 }
 
+bool is_ims_expected(const std::string& req_type)
+{
+  return (req_type == HSSConnection::REG || req_type == HSSConnection::CALL);
```
AlexHockey: I think we need IMS subscription XML on more occasions than this. For example, when a UE deregisters we need the XML to work out what the primary IMPU is. Is it actually just the `AUTH_*` request type where we don't need XML. 

You might also want to add a UT to check this - deregister a user but have homestead not return any XML and see what happens.

src/hssconnection.cpp
```diff
@@ -309,16 +309,20 @@ void parse_charging_addrs_node(rapidxml::xml_node<>* charging_addrs_node,
...
+}
+
 bool decode_homestead_xml(const std::string& public_id,
                           HSSConnection::irs_info& irs_info,
-                          std::shared_ptr<rapidxml::xml_document<> > root,
+                          const std::shared_ptr<rapidxml::xml_document<> > root,
                           SIFCService* sifc_service,
-                          const bool& allowNoIMS,
+                          const bool& IMSExpected,
```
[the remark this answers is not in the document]
AlexHockey: Our coding standards use `snake_case` for function arguments. So this should be `ims_subscription_expected`.

src/hssconnection.cpp
```diff
@@ -350,30 +354,33 @@ bool decode_homestead_xml(const std::string& public_id,
...
   {
     std::string sp_str;
     rapidxml::print(std::back_inserter(sp_str), *root, 0);
-    TRC_WARNING("Malformed HSS XML for %s - no IMSSubscription element:\n%s",
-                public_id.c_str(),
-                sp_str.c_str());
-    return false;
+    if (IMSExpected)
+    {
+      TRC_WARNING("Malformed HSS XML for %s - no IMSSubscription element:\n%s",
```
AlexHockey: You should never have an error, warning, or status log that can be hit on every call. Please make this info or lower.

src/hssconnection.cpp
```diff
@@ -350,30 +354,33 @@ bool decode_homestead_xml(const std::string& public_id,
...
+                  public_id.c_str(),
+                  sp_str.c_str());
+      return false;
+    }
+    else
+    {
+      TRC_INFO("In HSS XML for %s, there is no IMSSubscription element:\n%s",
```
AlexHockey: If this is an allowed case, it would be better to make this a debug log.

src/hssconnection.cpp
```diff
@@ -472,28 +477,62 @@ HTTPCode HSSConnection::update_registration_state(const irs_query& irs_query,
 
   if (http_code != HTTP_OK)
   {
-    // If get_xml_object has returned a HTTP error code, we have either not found
-    // the subscriber on the HSS or been unable to communicate with
-    // the HSS successfully. In either case we should fail.
+    // We have either not found the subscriber on the HSS, or been unable to 
+    // communicate with the HSS successfully.
     TRC_ERROR("Could not get subscriber data from HSS");
```
AlexHockey: I know this isn't your change, but this shouldn't be an error log either.

src/hssconnection.cpp
```diff
@@ -472,28 +477,62 @@ HTTPCode HSSConnection::update_registration_state(const irs_query& irs_query,
...
+{
+  // Needs to be a shared pointer - multiple Ifcs objects will need a reference
+  // to it, so we want to delete the underlying pointer when they all go out
+  // of scope.
+  std::shared_ptr<rapidxml::xml_document<>> root;
+
+  HTTPCode http_code = put_homestead_xml(irs_query, root, trail);
+  if (http_code == HTTP_OK)
+  {
+    bool IMSExpected = is_ims_expected(irs_query._req_type);
```
AlexHockey: Ditto. Should be `ims_expected`.

src/hssconnection.cpp
```diff
@@ -524,22 +566,12 @@ HTTPCode HSSConnection::get_registration_data(const std::string& public_id,
 
   if (http_code != HTTP_OK)
   {
-    // If get_xml_object has returned a HTTP error code, we have either not found
-    // the subscriber on the HSS or been unable to communicate with
-    // the HSS successfully. In either case we should fail.
+    // We have either not found the subscriber on the HSS, or been unable to 
+    // communicate with the HSS successfully.
     TRC_ERROR("Could not get subscriber data from HSS");
```
AlexHockey: Ditto

src/hssconnection.cpp
Ying-metaswitch: It's a genuine error case that when IMSSubscription is expected yet not present. The original code uses WARNING for this, so does every case within this function that returns false (corresponds to Sprout sending 500).

src/hssconnection.cpp
Ying-metaswitch: Changed to WARNING, reasoning as above

src/hssconnection.cpp
AlexHockey: Logs that appear in the log file default should be infrequent and refer to the service as a whole, not to individual calls (we have SAS to debug calls). If we ever hit this line we're going to print out a log containing the entire service profile *on potentially every call*. This is going to swamp the log file and make it difficult to spot more problems like crashes.

Ying-metaswitch:
TS29.228 Table S-CSCF registration/deregistration notification response:
User Profile.
It shall be present when Server-Assignment-Type in the request is equal to
NO_ASSIGNMENT, REGISTRATION, RE_REGISTRATION or
UNREGISTERED_USER.

At that time I took it as the User Profile should just be absent for all other SARs....Needs more checking

src/hssconnection.cpp
Ying-metaswitch: Ah I see, changed every such per-call warning in the file to DEBUG

src/hssconnection.cpp
```diff
@@ -309,17 +309,21 @@ void parse_charging_addrs_node(rapidxml::xml_node<>* charging_addrs_node,
   }
 }
 
+bool is_ims_subscription_expected(const std::string& req_type)
+{
+  return (req_type == HSSConnection::REG || req_type == HSSConnection::CALL);
```
sathiyan-sivathas: I think you still need to address @AlexHockey's comment on this line (not sure why it's gone missing).

In you comment on this PR you correctly identify that the user data is only only required when the server-assignment-type is "NO_ASSIGNMENT, REGISTRATION, RE_REGISTRATION or
UNREGISTERED_USER" so you need to make the check the appropriate values here too.

Alex suggests that the IMS Subscription needs to be present except when the `req_type` is `AUTH_*` but I'm not sure because SAR type ADMINISTRATIVE_DEREGISTRATION isn't in the spec but we expect to have the iFCs there to deregister with application servers.

Ying-metaswitch: I think we agreed that the spec means user data isn't strictly required for AUTH_* and DEREG_*, and existing UTs pass. But we need to be sure in the DEREG_* this doesn't break the logic of 'deregister_with_application_server' where user data is needed. I wrote the UT for dereg_admin, and need to write for other two dereg cases.

src/hssconnection.cpp
```diff
@@ -309,17 +309,25 @@ void parse_charging_addrs_node(rapidxml::xml_node<>* charging_addrs_node,
   }
 }
 
+bool is_ims_subscription_expected(const std::string& req_type)
+{
+  // Not expected for AUTH cases
+  return (req_type == HSSConnection::REG || req_type == HSSConnection::CALL || 
+          req_type == HSSConnection::DEREG_USER || 
+          req_type == HSSConnection::DEREG_ADMIN ||
+          req_type == HSSConnection::DEREG_TIMEOUT);
```
sathiyan-sivathas: Can you be consistent with your alignment here by moving `HSSConnection::Call` onto a new line.

src/ut/chronoshandlers_test.cpp
```diff
@@ -75,6 +75,7 @@ TEST_F(ChronosAoRTimeoutTasksTest, MainlineTest)
 
   HSSConnection::irs_info irs_info;
   irs_info._associated_uris = associated_uris;
+  irs_info._service_profiles = {};
```
sathiyan-sivathas: I don't think you need this. As we sort of discussed, you want to test `get_registration_data` and `update_registration_state`. Since those functions are mocked in this test suite, you're not gaining anything by putting this line in.

src/ut/hssconnection_test.cpp
```diff
@@ -212,6 +225,16 @@ class HssConnectionTest : public BaseTest
...
+    fakecurl_responses_with_body[std::make_pair("http://10.42.42.42:80/impu/missingelement2/reg-data", "{\"reqtype\": \"dereg-admin\", \"server_name\": \"server_name\"}")] =
+      "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+      "<ClearwaterRegData>"
+        "<RegistrationState>NOT_REGISTERED</RegistrationState>"
+      "</ClearwaterRegData>";
+    fakecurl_responses_with_body[std::make_pair("http://10.42.42.42:80/impu/missingelement2/reg-data", "{\"reqtype\": \"dereg-user\", \"server_name\": \"server_name\"}")] =
+      "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+      "<ClearwaterRegData>"
+        "<RegistrationState>NOT_REGISTERED</RegistrationState>"
+      "</ClearwaterRegData>";
```
sathiyan-sivathas: What about `dereg-timeout`and `call`? I think you want to test those cases too.

src/ut/hssconnection_test.cpp
```diff
@@ -427,6 +450,36 @@ TEST_F(HssConnectionTest, SimpleNotRegisteredGet)
...
+  irs_query._public_id = "pubid48";
+  irs_query._req_type = HSSConnection::AUTH_TIMEOUT;
+  irs_query._server_name = "server_name";
+  HSSConnection::irs_info irs_info;
+
+  HTTPCode rc =_hss.update_registration_state(irs_query, irs_info, 0);
+
+  EXPECT_EQ("NOT_REGISTERED", irs_info._regstate);
+  EXPECT_TRUE(irs_info._service_profiles.empty());
+  ASSERT_TRUE(rc == 200);
```
sathiyan-sivathas: I know some of the existing tests do this, but in general you should use `EXPECT_*` instead of `ASSERT_*` because it means that the rest of the tests can run even if this one fails. The place where `ASSERT_*` becomes useful is when you *need* a statement to be true for the rest of the test to work e.g. if you need to dereference a pointer, you should assert that it's not NULL.

Can you replace `ASSERT_TRUE` with `EXPECT_TRUE` here and in the rest of the tests you've added.

src/ut/hssconnection_test.cpp
```diff
@@ -654,12 +708,47 @@ TEST_F(HssConnectionTest, BadXML_MissingIMSSubscription)
...
-  _hss.update_registration_state(irs_query, irs_info, 0);
+  HTTPCode rc = _hss.update_registration_state(irs_query, irs_info, 0);
 
   EXPECT_TRUE(irs_info._associated_uris.get_unbarred_uris().empty());
+  EXPECT_TRUE(irs_info._service_profiles.empty());
+  EXPECT_TRUE(log.contains("Malformed HSS XML"));
+  ASSERT_TRUE(rc == 500);
+}
+
+TEST_F(HssConnectionTest, BadXML_MissingIMSSubscription2)
```
sathiyan-sivathas: Can you give these tests more descriptive names e.g. `BadXML_MissingIMSSubscription_DeregAdmin`

src/ut/hssconnection_test.cpp
```diff
@@ -654,12 +708,47 @@ TEST_F(HssConnectionTest, BadXML_MissingIMSSubscription)
...
+  ASSERT_TRUE(rc == 500);
 }
 
+TEST_F(HssConnectionTest, BadXML_MissingIMSSubscription3)
+{
+  CapturingTestLogger log;
+
+  HSSConnection::irs_query irs_query;
+  irs_query._public_id = "missingelement2";
+  irs_query._req_type = HSSConnection::DEREG_USER;
```
sathiyan-sivathas: You've covered `DEREG_ADMIN` and `DEREG_USER` and there's already a test for `REG`. Can you add tests for `DEREG_TIMEOUT` and `CALL` please.

src/hssconnection.cpp
```diff
@@ -309,17 +309,25 @@ void parse_charging_addrs_node(rapidxml::xml_node<>* charging_addrs_node,
   }
 }
 
+bool is_ims_subscription_expected(const std::string& req_type)
+{
+  // Not expected for AUTH cases
```
sathiyan-sivathas: Can you expand on this comment to give more context i.e. refer to the spec reference that you found and state that we expect homestead to have cached the user data in the `DEREG_*` cases.

Ying-metaswitch:
full_test passed. Added UT for missing_ims_subscription for every type, and removed checking the logs as other EXPECTs are more robust.

PR #2017, opened 2017-11-17
Fix register correlation

MatMeredith:
This (along with https://github.com/Metaswitch/cpp-common/pull/717 and https://github.com/Metaswitch/sas-client/pull/78) fixes https://github.com/Metaswitch/clearwater-issues/issues/2730 and https://github.com/Metaswitch/clearwater-issues/issues/2505 and (probably) https://github.com/Metaswitch/clearwater-issues/issues/2602

See discussion of fix in https://github.com/Metaswitch/clearwater-issues/issues/2505.

UTs have been enhanced to cover this fix, and it has also been tested live.

### DIFF
--- a/src/authenticationsproutlet.cpp
+++ b/src/authenticationsproutlet.cpp
@@ -670,6 +670,17 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
     // Digest authentication).
     hdr->scheme = STR_DIGEST;
     pj_pool_t* rsp_pool = get_pool(rsp);
+    pj_create_random_string(buf, sizeof(buf));
+    pj_strdup(rsp_pool, &hdr->challenge.digest.opaque, &random);
+
+    // Log the opaque value to SAS to enable us to correlate this challenge
+    // with the subsequent transaction containing the challenge response.
+    std::string opaque;
+    opaque.assign(buf, sizeof(buf));
+    TRC_DEBUG("Log opaque value %s to SAS as a generic correlator", opaque.c_str());
+    SAS::Marker opaque_marker(trail(), MARKED_ID_GENERIC_CORRELATOR, 1u);
+    opaque_marker.add_var_param(opaque);
+    SAS::report_marker(opaque_marker, SAS::Marker::Scope::Trace);
 
     ImpiStore::AuthChallenge* auth_challenge;
     if (av->is_aka())
@@ -686,8 +697,6 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
       hdr->challenge.digest.algorithm = ((aka->akaversion == 2) ? STR_AKAV2_MD5 : STR_AKAV1_MD5);
 
       pj_strdup2(rsp_pool, &hdr->challenge.digest.nonce, aka->nonce.c_str());
-      pj_create_random_string(buf, sizeof(buf));
-      pj_strdup(rsp_pool, &hdr->challenge.digest.opaque, &random);
       hdr->challenge.digest.qop = STR_AUTH;
       hdr->challenge.digest.stale = stale;
 
@@ -770,8 +779,6 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
       pj_create_random_string(buf, sizeof(buf));
       nonce.assign(buf, sizeof(buf));
       pj_strdup(rsp_pool, &hdr->challenge.digest.nonce, &random);
-      pj_create_random_string(buf, sizeof(buf));
-      pj_strdup(rsp_pool, &hdr->challenge.digest.opaque, &random);
       pj_strdup2(rsp_pool, &hdr->challenge.digest.qop, digest->qop.c_str());
       hdr->challenge.digest.stale = stale;
 
@@ -786,11 +793,10 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
     // Add the header to the message.
     pjsip_msg_add_hdr(rsp, (pjsip_hdr*)hdr);
 
-    // Store the branch parameter in memcached for correlation purposes
-    pjsip_via_hdr* via_hdr = (pjsip_via_hdr*)pjsip_msg_find_hdr(req, PJSIP_H_VIA, NULL);
-    auth_challenge->set_correlator((via_hdr != NULL) ?
-                                   PJUtils::pj_str_to_string(&via_hdr->branch_param) :
-                                   "");
+    // Store the opaque value that we generated in the IMPI store.  This is
+    // needed if the challenge times out so that the SAS trace there can be
+    // correlated with the one for this initial REGISTER.
+    auth_challenge->set_correlator(opaque);
 
     // Add the IMPU to the challenge
     auth_challenge->set_impu(impu_for_hss);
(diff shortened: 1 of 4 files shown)


src/authenticationsproutlet.cpp
```diff
@@ -670,6 +670,17 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
...
     pj_pool_t* rsp_pool = get_pool(rsp);
+    pj_create_random_string(buf, sizeof(buf));
+    pj_strdup(rsp_pool, &hdr->challenge.digest.opaque, &random);
+
+    // Log the opaque value to SAS to enable us to correlate this challenge
+    // with the subsequent transaction containing the challenge response.
+    std::string opaque;
+    opaque.assign(buf, sizeof(buf));
+    TRC_DEBUG("Log opaque value %s to SAS as a generic correlator", opaque.c_str());
+    SAS::Marker opaque_marker(trail(), MARKED_ID_GENERIC_CORRELATOR, 1u);
```
AlexHockey: By convention generic correlators have a static parameter which specifies at what scope the first var data is unique. This avoids accidental clashes. See for example https://github.com/Metaswitch/cpp-common/blob/master/src/httpstack.cpp#L550. Can you do something similar?

src/authenticationsproutlet.cpp
```diff
@@ -786,11 +793,15 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
...
     pjsip_msg_add_hdr(rsp, (pjsip_hdr*)hdr);
 
-    // Store the branch parameter in memcached for correlation purposes
-    pjsip_via_hdr* via_hdr = (pjsip_via_hdr*)pjsip_msg_find_hdr(req, PJSIP_H_VIA, NULL);
-    auth_challenge->set_correlator((via_hdr != NULL) ?
-                                   PJUtils::pj_str_to_string(&via_hdr->branch_param) :
-                                   "");
+    // Store the opaque value that we generated in the IMPI store.  This is
+    // needed if the challenge times out so that the SAS trace there can be
+    // correlated with the one for this initial REGISTER.
```
AlexHockey: Does this paragraph parse?

src/authenticationsproutlet.cpp
```diff
@@ -786,11 +793,15 @@ void AuthenticationSproutletTsx::create_challenge(pjsip_digest_credential* crede
...
+    // Note we aren't dependent on this to correlate the challenged REGISTER
+    // transaction and the transaction containing the challenge response
+    // because we use the opaque value from the SIP messages directly (meaning
+    // that transactions get correlated even in the case of IMPI store
+    // unavailability.
```
AlexHockey: Missing closing bracket

src/authenticationsproutlet.cpp
```diff
@@ -978,6 +989,19 @@ void AuthenticationSproutletTsx::on_rx_initial_request(pjsip_msg* req)
     unsigned long nonce_count = pj_strtoul2(&credentials->nc, NULL, 16);
     nonce_count = (nonce_count == 0) ? 1 : nonce_count;
 
+    // If this is the first response to the challenge then log the value of
+    // opaque to SAS as a marker. We also do this when we challenge the initial
+    // REGISTER and in this way the two transactions (the challenge and the
+    // challenge response) get correlated in SAS.
+    if (nonce_count == 1)
```
AlexHockey: Using the nonce count from the message could be a security vulnerability. It would allow an attacker to snoop a challenge answer REGISTER, send a bunch of identical REGISTERs, and make the trace in SAS that is too big to load. Should we wait until we know the nonce count is acceptable before logging the correlator?

src/impistore.cpp
```diff
@@ -296,12 +296,13 @@ void correlate_trail_to_challenge(ImpiStore::AuthChallenge* auth_challenge,
   // Report the correlator as a SAS marker, if it exists.
   if (auth_challenge->get_correlator() != "")
   {
-    SAS::Marker via_marker(trail, MARKER_ID_VIA_BRANCH_PARAM, 1u);
-    via_marker.add_var_param(auth_challenge->get_correlator());
-    SAS::report_marker(via_marker, SAS::Marker::Scope::Trace);
+    SAS::Marker marker(trail, MARKED_ID_GENERIC_CORRELATOR, 3u);
```
AlexHockey: This technically breaks correlation across upgrade because we're moving to using a different marker. Presumably we don't care.

src/ut/authentication_test.cpp
```diff
@@ -152,6 +156,35 @@ class BaseAuthenticationTest : public SipTest
...
+    std::vector<MockSASMessage*> markers = mock_sas_find_marker_multiple(MARKED_ID_GENERIC_CORRELATOR);
+    for (MockSASMessage* marker : markers)
+    {
+      EXPECT_EQ(marker->var_params.size(), 1u);
+      if (marker->var_params[0] == value)
+      {
+        found_value = true;
+        break;
+      }
+    }
```
AlexHockey: You could use `find_if` instead of this for loop 😄

src/ut/authentication_test.cpp
```diff
@@ -152,6 +156,35 @@ class BaseAuthenticationTest : public SipTest
...
+    EXPECT_EQ(found_value, present);
+  }
+
+  // Chech that the AV for an IMPI in the ImpiStore has a particular stored
+  // correlator.
+  void check_impi_store_correlator(std::string impi_id, std::string nonce, std::string expected_correlator)
+  {
+    ImpiStore::Impi* impi = _impi_store->get_impi(impi_id, 0L);
+    ImpiStore::AuthChallenge* challenge = impi->get_auth_challenge(nonce);
+    EXPECT_NE(challenge, (void *)NULL);
```
AlexHockey: If you use `nullptr` you could avoid the cast (nullptr is preferred in C++11)

src/ut/authentication_test.cpp
```diff
@@ -760,6 +793,7 @@ TEST_F(AuthenticationTest, DigestAuthSuccess)
   // Test a successful SIP Digest authentication flow.
   pjsip_tx_data* tdata;
 
+
```
AlexHockey: Random blank line?

src/ut/authentication_test.cpp
```diff
@@ -1588,6 +1664,16 @@ TEST_F(AuthenticationTest, AKAv2AuthSuccess)
...
+  // Check that we logged the opaque value to SAS as a GENERIC_CORRELATOR.
+  // This is needed so that SAS can correlate this transaction with the
+  // subsequent challenge response below.
+  check_sas_correlator(auth_params["opaque"]);
+  mock_sas_discard_messages();
+
+  // Also check that we wrote the opaque value to the IMPI Store as the
+  // correlator.
+  check_impi_store_correlator("6505550001@homedomain", auth_params["nonce"], auth_params["opaque"]);
+
```
AlexHockey: Do we really need to check the correlators in all these tests? It adds a lot of noise to the testcases.

src/authenticationsproutlet.cpp
MatMeredith: So I did think about this.  I think!
- I don't think we can check that it is valid -- that depends on being able to read the AV from the IMPI Store right, and the whole point of making this change was so that we would correlate the SAS traces even in this kind of error scenario?
- But also, have we not already logged the actual SIP message to SAS by this point?   True, it won't be correlated, but someone already has the ability to spam arbitrarily large numbers of messages into SAS don't they?   Is this really much worse?

src/ut/authentication_test.cpp
MatMeredith: Thanks :-)   I may not make this markup since it's just in UT code, but will definitely remember next time!

src/ut/authentication_test.cpp
MatMeredith: Possibly not.   I wanted to check a range of error conditions (since that's where the issue was), the case where we correlate 3 transactions and both AKA / Digest.   There might be one or two too many though...

MatMeredith:
I've made the markups that I think really _need_ making (in the context of having already spent more time on this that it deserves).   Can I check you're happy?

AlexHockey:
Looks good to me.

PR #2020, opened 2017-11-21
Add diagnostics when Sprout sends a CANCEL

metaswitch-colin-haywood:
- Add new SAS logs and diags for cancel_fork and other places where CANCELs are generated
- Improve diags for app server failures
- Improve diags for "client not responding" failures


metaswitch-colin-haywood:
I've live-tested this fix and run `make full_test`. (against V10 anyway - can't get dev to build atm)

metaswitch-colin-haywood:
Got dev building, `make full_test` now passes.

include/sproutlet.h
```diff
@@ -284,13 +284,13 @@ class SproutletTsxHelper
   ///
   /// @param fork_id       - The identifier of the fork to cancel.
```
KristaHyer: Missing other `@param`s? Same for cancel_pending_forks and cancel_fork

src/basicproxy.cpp
```diff
@@ -325,7 +325,7 @@ void BasicProxy::on_cancel_request(pjsip_rx_data* rdata)
   // Send CANCEL to cancel the UAC transactions.
   // The UAS INVITE transaction will get final response when
   // we receive final response from the UAC INVITE transaction.
-  uas_tsx->process_cancel_request(rdata);
+  uas_tsx->process_cancel_request(rdata, "CANCEL request received from peer");
```
KristaHyer: We should pull out all of these strings and define them in the header.

src/basicproxy.cpp
```diff
@@ -680,14 +680,15 @@ void BasicProxy::UASTsx::process_tsx_request(pjsip_rx_data* rdata)
...
-void BasicProxy::UASTsx::process_cancel_request(pjsip_rx_data* rdata)
+void BasicProxy::UASTsx::process_cancel_request(pjsip_rx_data* rdata, const std::string& reason)
 {
   TRC_DEBUG("%s - Cancel for UAS transaction", name());
 
   // Send CANCEL to cancel the UAC transactions.
   // The UAS INVITE transaction will get final response when
   // we receive final response from the UAC INVITE transaction.
-  cancel_pending_uac_tsx(0, false);
+  // @@CPH what st_code to use here? maybe get it from rdata?
```
KristaHyer: So we currently put 0 as the st_code.
This means that [here](https://github.com/Metaswitch/sprout-bgcf-archive/blob/2075cd29a5e55735d5fed6cffe95b2e2bdd54fd9/src/pjutils.cpp#L926) we just don't add a reason to the CANCEL we send.
Is there something wrong with this behaviour? Can we not just leave it as 0?
(Same comment on next `@@CPH`.)

src/sproutletproxy.cpp
```diff
@@ -1686,25 +1698,27 @@ void SproutletWrapper::send_response(pjsip_msg*& rsp)
...
+                  _id.c_str(), fork_id, st_code, reason.c_str());
       _forks[fork_id].pending_cancel = true;
+      _forks[fork_id].cancel_st_code = st_code;
       _forks[fork_id].cancel_reason = reason;
     }
   }
 }
 
-void SproutletWrapper::cancel_pending_forks(int reason)
+// @@CPH should this not call into cancel_fork rather than duplicating code?
```
KristaHyer: I agree, it should

src/basicproxy.cpp
metaswitch-colin-haywood: I know this is the standard thing to do with constants, but there are arguments against doing that here: 
* they're used only for diagnostics and are unique
* the general use case would be: someone sees the string in the SAS, and searches for it in the code. When it's defined inline, they have to jump through one fewer hoop.

src/basicproxy.cpp
metaswitch-colin-haywood: I've left it as zero. Just thinking, the more diags, the better.

include/sproutlet.h
metaswitch-colin-haywood: Fixed.

src/sproutletproxy.cpp
metaswitch-colin-haywood: Well actually there is a slight difference between the two: `_forks[ii].state.tsx_state != PJSIP_TSX_STATE_NULL` is required in `cancel_pending_forks`. And ideally you'd want to make a function that cancels an actual `ForkStatus` rather than doing another unnecessary lookup by index... So overall I can't really be bothered to fix this right now as it's more effort than it's worth, but shout if you disagree.

src/basicproxy.cpp
richardwhiuk: Yeah, I agree - having them as a header just increases the distance you have to look when reading the code and adds no value here.

KristaHyer: I'm also happy for you to leave these as strings

src/sproutletproxy.cpp
KristaHyer: Sounds sensible to me, happy for you to check this in

PR #2022, opened 2017-11-22
[Reviewer: Richard] S-CSCF should reject unrouteable URIs

sebrexmetaswitch:
This is the other half of the fix for https://github.com/Metaswitch/clearwater-issues/issues/2432

PR is against dev, but depends on Ellie's changes in the `validate_impus` branch, so the more useful comparison here is https://github.com/Metaswitch/sprout/compare/validate_impus...sccsf_reject_unrouteable

This PR adds 2 commits:

1) When we've finished originating processing, reject the request if we have a non-sip or -tel URI as the request URI, where we previously assumed that if was valid and hence sent it to the I-CSCF

2) If we decide not to apply any services to a request, and the next-hop URI is non-sip or -tel, then reject the request (where we would previously attempt to route it and that would fail without sending a response)

The former handles originating requests that have (because of some AS misconfiguraion) not been handled by an AS. The latter handles terminating requests.

I've tested this by adding 3 UTs (originating with and without ENUM, and terminating).

`make full_test` passes.


src/scscfsproutlet.cpp
```diff
@@ -1440,19 +1458,46 @@ void SCSCFSproutletTsx::apply_originating_services(pjsip_msg* req)
...
+      URIClass uri_class = URIClassifier::classify_uri(req->line.req.uri, true, false);
+
+      if (uri_class != UNKNOWN)
+      {
+        TRC_DEBUG("No ENUM lookup available - routing to I-CSCF");
+        route_to_icscf(req);
+      }
+      else
+      {
+        // Invalid URI, so just reject the request
```
richardwhiuk: This block is repeated roughly three times, can we add a 'reject_request_due_to_invalid_uri()` function? Can be static in this file.

include/sproutsasevent.h
```diff
@@ -192,6 +192,7 @@ namespace SASEvent
   const int NO_AS_CHAIN_ROUTE = SPROUT_BASE + 0x0000EB;
   const int ORIG_SIP_TO_TEL = SPROUT_BASE + 0x0000EC;
   const int ICSCF_INVALID_IMPU = SPROUT_BASE + 0x0000ED;
+  const int SCSCF_INVALID_URI = SPROUT_BASE + 0x0000EF;
```
richardwhiuk: What's wrong with EE, OOI?

sebrexmetaswitch: Hmm, I had thought that `EE` was added by the branch this was based off. Looks like I was wrong about that though

PR #2025, opened 2017-11-24
Log basic stack trace unless aborting

AlexHockey:
Ellie, these are the co-req sprout changes for https://github.com/Metaswitch/cpp-common/pull/721. Please can you review and, if you;re happy, I will apply these to the rest of the codebase.


AlexHockey:
As discussed, commonalizing the signal handling turned out to be a more destabilizing change than I anticipated, so I'm going to track this as tech debt instead.

PR #2026, opened 2017-11-27
[Reviewer: RJW2] Allow non-PJSIP threads to place work on the worker thread queue

sebrexmetaswitch:
As discussed, the recent change to try to move Chronos timer pop handling to a worker thread wasn't working because only the transport thread was allowed to place work on the queue.

This PR removes that restriction. The queue is already protected by a lock, so we just don't report the queue size when adding work this way, which is OK because the transport thread will report the correct size when it next places a message on the queu.

It's not tested other than by `make test`, so we should run load through this change before we merge it.

### DIFF
--- a/src/pjutils.cpp
+++ b/src/pjutils.cpp
@@ -1233,15 +1233,16 @@ static void on_tsx_state(pjsip_transaction* tsx, pjsip_event* event)
 
 /// Runs a Callback object on a worker thread.
 /// Takes ownership of the Callback and is responsible for deleting it
-void PJUtils::run_callback_on_worker_thread(PJUtils::Callback* cb)
+void PJUtils::run_callback_on_worker_thread(PJUtils::Callback* cb,
+                                            bool is_pjsip_thread)
 {
   // The UTs have a different threading model - in those we run the callback
   // directly on whatever thread we're on
 #ifndef UNIT_TEST
-  if (is_pjsip_transport_thread())
+  if (is_pjsip_transport_thread() || !is_pjsip_thread)
   {
-    // We're on the transport thread, so we must add the callback to the worker
-    // thread's queue
+    // We're either on the transport thread or on a non-PJSIP owned thread, so
+    // add the callback to the worker thread queue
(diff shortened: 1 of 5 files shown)


src/pjutils.cpp
```diff
@@ -1233,15 +1233,16 @@ static void on_tsx_state(pjsip_transaction* tsx, pjsip_event* event)
...
 /// Takes ownership of the Callback and is responsible for deleting it
-void PJUtils::run_callback_on_worker_thread(PJUtils::Callback* cb)
+void PJUtils::run_callback_on_worker_thread(PJUtils::Callback* cb,
+                                            bool is_pjsip_thread)
 {
   // The UTs have a different threading model - in those we run the callback
   // directly on whatever thread we're on
 #ifndef UNIT_TEST
-  if (is_pjsip_transport_thread())
+  if (is_pjsip_transport_thread() || !is_pjsip_thread)
```
richardwhiuk: I'd reverse the check to allow us to short circuit the (relatively) expensive is_pjsip_transport_thread() call.

src/thread_dispatcher.cpp
```diff
@@ -735,12 +735,9 @@ void add_callback_to_queue(PJUtils::Callback* cb)
   // future we may want to look at prioritizing them
   qe.priority = SIPEventPriorityLevel::NORMAL_PRIORITY;
 
-  // Track the current queue size
-  if (queue_size_table)
-  {
-    queue_size_table->accumulate(sip_event_queue.size()); // LCOV_EXCL_LINE
-  }
```
richardwhiuk: Worth reinstating this given we know that it's safe? Alternatively, should have a comment why we aren't accumulating the stat.

sebrexmetaswitch: Comment added below

sebrexmetaswitch:
We've now run load over this, including a timer storm after the end of the load run when all subscribers' registrations timed out, and Sprout didn't crash so happy this is working

PR #2027, opened 2017-11-28
[Reviewer EM] Send out terminate+create contact in NOTIFY if Contact URI changes

steelyvoid:
Hi Ellie

Can you review this "terminate and recreate Contact in NOTIFY if Contact URI changes" fix?  I think the customer is going to want it in 11.1 - can we get it in despite the code freeze?

Some notes

- I tested live using sipp scripts to sending in a UDP REGISTER to a given IMPU first from one client box and then from another.  I confirmed that the P-CSCF was apparently happy to delete the old contact
- I haven't changed the logic for analytics, which means that (I think), we'll get both a call to

_analytics->registration(classified_binding->_b->_address_of_record,
                               classified_binding->_id,
                               classified_binding->_b->_uri,
                               0);

and to 

      _analytics->registration(classified_binding->_b->_address_of_record,
                               classified_binding->_id,
                               classified_binding->_b->_uri,
                               classified_binding->_b->_expires - now);
Is this acceptable?

### DIFF
--- a/src/ut/registrar_test.cpp
+++ b/src/ut/registrar_test.cpp
@@ -306,7 +306,8 @@ public:
   void check_notify(pjsip_msg* out,
                     std::string expected_aor,
                     std::string reg_state,
-                    std::pair<std::string, std::string> contact_values)
+                    std::pair<std::string, std::string> contact_values,
+                    bool check_last = false)
   {
     char buf[16384];
     int n = out->body->print_body(out->body, buf, sizeof(buf));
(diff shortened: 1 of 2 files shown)


src/ut/registrar_test.cpp
```diff
@@ -331,7 +332,15 @@ class RegistrarTest : public SipTest
     ASSERT_TRUE(reg_info);
     rapidxml::xml_node<> *registration = reg_info->first_node("registration");
     ASSERT_TRUE(registration);
-    rapidxml::xml_node<> *contact = registration->first_node("contact");
+    rapidxml::xml_node<> *contact;
```
eleanor-merry: I'm not a fan of this - it feels somewhat arbitrary that we check the first or last contact node. I'd rather see a check that looks at each contact node and checks them in turn.

steelyvoid: I've changed the code to check a particular contact node (defaulting to 0) starting from 0 (so that, to check all of 3 contacts you would call

  check_notify(..., 0);
  check_notify(..., 1);
  check_notify(..., 2);

PR #2028, opened 2017-11-29
[Reviewer: EM] Ignore load monitor for emergency requests and AS requests

sathiyan-sivathas:
Ignores load monitor for:
 - Requests containing an ODI token.
 - all registrations.
 - MESSAGEs with "urn:service:sos" in the request URI.

SAS PR to follow.

### DIFF
--- a/src/thread_dispatcher.cpp
+++ b/src/thread_dispatcher.cpp
@@ -373,8 +373,8 @@ enum IGNORE_LOAD_MONITOR_REASON
   IN_DIALOG,
   ACK,
   SUBSCRIBE,
+  REGISTER,
   ODI_TOKEN,
-  EMERGENCY_REGISTRATION,
   URN_SERVICE_SOS
 };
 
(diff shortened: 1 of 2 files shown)


src/thread_dispatcher.cpp
```diff
@@ -443,6 +453,34 @@ static bool ignore_load_monitor(pjsip_rx_data* rdata,
...
+  pjsip_uri* req_uri = rdata->msg_info.msg->line.req.uri;
+  if ((pjsip_method_cmp(&method, &message_method) == 0) &&
+      (PJSIP_URI_SCHEME_IS_URN(req_uri)))
+  {
+    std::string req_uri_str = PJUtils::uri_to_string(PJSIP_URI_IN_REQ_URI, req_uri);
+    if (req_uri_str == "urn:service:sos")
+    {
+      log_ignore_load_monitor(trail, URN_SERVICE_SOS);
+      return true;
+    } // LCOV_EXCL_LINE - For some reason this isn't covered.
```
eleanor-merry: Not a fan of this - presumably this needs a UT with a MESSAGE with an URN Req URI that isn't urn:services:sos.

src/thread_dispatcher.cpp
```diff
@@ -443,6 +453,34 @@ static bool ignore_load_monitor(pjsip_rx_data* rdata,
...
+    return true;
+  }
+
+  // Always accept MESSAGEs with "urn:service:sos" in the request URI.
+  const pjsip_method message_method= { PJSIP_OTHER_METHOD, { "MESSAGE", 7 }};
+  pjsip_uri* req_uri = rdata->msg_info.msg->line.req.uri;
+  if ((pjsip_method_cmp(&method, &message_method) == 0) &&
+      (PJSIP_URI_SCHEME_IS_URN(req_uri)))
+  {
+    std::string req_uri_str = PJUtils::uri_to_string(PJSIP_URI_IN_REQ_URI, req_uri);
```
eleanor-merry: Is this the right check? Why not just check if the Req URI is urn:services:sos? It doesn't matter about the MESSAGE or the URN URIness. I'd also like this not to be hard coded (and be a configuration option of whitelisted URIs), but happy enough for now.

PR #2029, opened 2017-11-29
[Reviewer: SS5] Recover from CSeq desync

tomwatson1024:
The CSeq counter for a SUBSCRIBE/NOTIFY dialog is stored on the AoR corresponding to that dialog, and this AoR is stored in memcached both locally and at remote sites. Any time the local SDM increments the CSeq it relies on the client also calling into the remote SDMs to increment the CSeqs on the remote copies of the AoR. Whenever a NOTIFY is sent from the local site, it uses the local CSeq value and increments both the remote and local values, and similarly when a NOTIFY is sent from the local site. 

Due to a netsplit or perhaps another unknown condition (edit: while testing this, I ran into two other existing bugs that consistently cause out-of-sync CSeqs - I think it's only luck that we don't hit them more frequently), the counters can become out of sync between the local and remote sites. This results in NOTIFYs with incorrect CSeq values being sent, as each site uses its local value - the problem persists until the all registrations on the AoR expire.

This pull request adds function allowing Sprout to recover from this case - if the local and remote CSeq counters differ on an AoR when adding/updating/removing a subscription (note that any subscription must be periodically refreshed by the client), the remote CSeq values are overwritten with the local one. (edit: the PR also adds basic fixes for the two bugs described above)

If the local CSeq is larger than the remote ones, the problem is resolved at this point. Otherwise, we may send a number of NOTIFYs with incorrect CSeqs, but the problem will resolve after a small number of these.

It is important to note that the changes made here are **not** intended to remain in the code for long. The SDM and its clients is an area of code known to require refactoring. A cohesive fix for this issue essentially consists of a good chunk of this refactoring, which we do not want to do at this stage, so to avoid throw away work the fix is minimal and by no means ideal.

We hope to schedule this refactoring work shortly, at which point we will be able to resolve these issues in a more satisfactory way. I have put aside my initial work on this fix, PR #1999, in favor of this simpler short-term fix, as it was becoming a sink of time with no real way of finishing short of a complete refactor.

### DIFF
--- a/src/registrarsproutlet.cpp
+++ b/src/registrarsproutlet.cpp
@@ -439,6 +439,7 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
   // what Homestead thought the previous regstate was.
   bool is_initial_registration = (irs_info._prev_regstate == RegDataXMLUtils::STATE_NOT_REGISTERED);
   bool no_existing_bindings_found = false;
+  int initial_notify_cseq = 0;
   AoRPair* aor_pair = write_to_store(_registrar->_sdm,
                                      aor,
                                      &(irs_info._associated_uris),
@@ -450,7 +451,8 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
                                      NULL,
                                      _registrar->_remote_sdms,
                                      private_id_for_binding,
-                                     all_bindings_expired);
+                                     all_bindings_expired,
+                                     initial_notify_cseq);
 
   // Update our view of whether this was in fact an initial registration based
   // on whether we found any bindings. There are race conditions where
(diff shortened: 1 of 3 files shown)


tomwatson1024:
Details of the two bugs I hit during testing are as follows.

- When first registering, we write to the local store, incrementing the CSeq on an AoR pair, and then write to the remote stores, incrementing the CSeq on the same AoR pair. This means that the CSeq written to each store differs by one from the CSeq written to the previous store. I have fixed this by storing the initial CSeq and writing it to the AoR pair before each store write.

- When subscribing, if no subscriptions are present on the AoR in a remote store we overwrite the remote AoR with a copy of the local AoR. However, the CSeq on the local AoR has already been incremented, and the CSeq on the copy is incremented again during the write. This causes the CSeqs in the remote stores to be one higher than the CSeq in the local store. I have fixed this by storing the remote CSeq and writing it back to the copied AoR.

tomwatson1024:
`make full_test` passes, as to the live tests. I have also verified the fix by injecting REGISTER/SUBSCRIBE messages directly into Sprout nodes and observing the response CSeqs, while introducing a netsplit. I have checked that the CSeq values written to the memcached stores are consistent across sites without a netsplit, and recover after a netsplit.

src/registrarsproutlet.cpp
```diff
@@ -481,6 +481,13 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
     // Log the bindings.
     log_bindings(aor, aor_pair->get_current());
 
+    // We want to write the same CSeq value to the remote stores as we have just
+    // written to the local one. Writing to the store causes the CSeq to be
+    // incremented, so we correct for that by decrementing it.
+    // This isn't an ideal solution - this entire area of code is due to be
+    // refactored, at which point a more permanent fix should be implemented.
+    int local_notify_cseq = aor_pair->get_current()->_notify_cseq - 1;
```
sathiyan-sivathas: Instead of writing to the store once and then subtracting -1 to get back to the original cseq, why not store off the initial cseq before writing to any stores, and reset the cseq before writes? I think that would be clearer.

tomwatson1024: I agree, I've made this change :)

tomwatson1024:
I've made the change suggested and done some basic retesting - `make full_test`, the live tests, and my custom tests pass. Could you take a cursory glace over the changes?

PR #2031, opened 2017-11-30
[Reviewer: KH1] Add case insensitive compare function to RPH map

sathiyan-sivathas:
Adds a case insensitive key_compare function for the RPH map so that an RPH value with different case to the one configured still returns the correct priority.

Tested by changing the cases of random characters in RPH values in the UTs. I did this in the config files to check that config validation still works, and in the tests to check key lookups work with different cases.

### DIFF
--- a/include/rphservice.h
+++ b/include/rphservice.h
@@ -51,8 +51,8 @@ private:
   Alarm* _alarm;
   std::string _configuration;
 
-  // RPH values are case insensitive, so write a case insensitive compare
-  // function to pass to the map.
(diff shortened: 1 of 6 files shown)


include/rphservice.h
```diff
@@ -49,7 +50,19 @@ class RPHService
 private:
   Alarm* _alarm;
   std::string _configuration;
-  std::map<std::string, SIPEventPriorityLevel> _rph_map;
+
+  // RPH values are case insensitive, so write a case insensitive compare
```
KristaHyer: I would ref the spec this is stated in (phil put it in the issue for us)

PR #2036, opened 2017-12-04
Add missing dependency

huws:
Previous check-in (https://github.com/Metaswitch/sprout/pull/2034) linked against `boost-locale`, but failed to list the library as a dependency.


huws:
@MatMeredith - checking in without review to fix build breakage.

PR #2039, opened 2017-12-05
Deregister AS in AoR Timeout Task

Ying-metaswitch:
This PR makes AoRTimeoutTask sends third party REGISTER when all bindings have expired, similar to DeregistrationTask and DeleteImpuTask and consistent with TS24.229.
1. The task calls deregister_with_application_server after updating hss, similar to RegistrationUtils::remove_bindings. Spec p.235-238 suggests that ordering as well.
2. AoRTimeoutTask now gets fifc_service and ifc_configuration in its Config, for the use of deregister_with_application_server.
2. UT checks deregister_with_application_server is called by log, which isn't ideal. Unfortunately deregister_with_application_server is an utility function and does not return success or not (I feel it should), so handler_test is checking this call by log as well.

full_test passed

### DIFF
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -456,12 +456,13 @@ AoRPair* DeregistrationTask::deregister_bindings(
 {
   AoRPair* aor_pair = NULL;
   bool all_bindings_expired = false;
+  bool got_ifcs;
   Store::Status set_rc;
   std::vector<std::string> impis_to_dereg;
 
   // Get registration data
   HSSConnection::irs_info irs_info;
-  bool got_ifcs = get_reg_data(_cfg->_hss, aor_id, irs_info, trail());
+  got_ifcs = get_reg_data(_cfg->_hss, aor_id, irs_info, trail());
 
   do
   {
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -233,7 +233,7 @@ void AoRTimeoutTask::process_aor_timeout(std::string aor_id)
 
   // Determine the set of IMPUs in the Implicit Registration Set
   HSSConnection::irs_info irs_info;
-  get_reg_data(_cfg->_hss, aor_id, irs_info, trail());
+  bool got_ifcs = get_reg_data(_cfg->_hss, aor_id, irs_info, trail());
 
   bool all_bindings_expired = false;
   AoRPair* aor_pair = get_and_set_local_aor_data(_cfg->_sdm,
@@ -261,6 +261,18 @@ void AoRTimeoutTask::process_aor_timeout(std::string aor_id)
                                *aor_pair,
                                _cfg->_hss,
                                trail());
+
+      if (got_ifcs)
+      {
+        RegistrationUtils::deregister_with_application_servers(irs_info._service_profiles[aor_id],
+                                                               _cfg->_fifc_service,
+                                                               _cfg->_ifc_configuration,
+                                                               _cfg->_sdm,
+                                                               _cfg->_remote_sdms,
+                                                               _cfg->_hss,
+                                                               aor_id,
+                                                               trail());
+      }
     }
   }
   else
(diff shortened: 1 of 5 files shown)


src/handlers.cpp
```diff
@@ -444,13 +456,12 @@ AoRPair* DeregistrationTask::deregister_bindings(
...
   AoRPair* aor_pair = NULL;
   bool all_bindings_expired = false;
-  bool got_ifcs;
   Store::Status set_rc;
   std::vector<std::string> impis_to_dereg;
 
   // Get registration data
   HSSConnection::irs_info irs_info;
-  got_ifcs = get_reg_data(_cfg->_hss, aor_id, irs_info, trail());
+  bool got_ifcs = get_reg_data(_cfg->_hss, aor_id, irs_info, trail());
```
sathiyan-sivathas: This function is declaring variables at the top, so I think it makes sense to leave the declaration where it was in this case.

src/ut/chronoshandlers_test.cpp
```diff
@@ -235,6 +240,8 @@ TEST_F(ChronosAoRTimeoutTasksTest, LocalAoRNoBindingsTest)
 // Test with a remote store, and both AoRs with no bindings
 TEST_F(ChronosAoRTimeoutTasksTest, NoBindingsTest)
 {
+  CapturingTestLogger log(5);
```
sathiyan-sivathas: Haven't you put this in the wrong test?

src/ut/chronoshandlers_test.cpp
```diff
@@ -314,6 +321,7 @@ TEST_F(ChronosAoRTimeoutTasksTest, NoBindingsTest)
   ASSERT_EQ(irs_query._req_type, HSSConnection::DEREG_TIMEOUT);
   ASSERT_EQ(irs_query._server_name, "sip:scscf.sprout.homedomain:5058;transport=TCP");
 
+  EXPECT_TRUE(log.contains("Creating third party deregistration"));
```
sathiyan-sivathas: It's not ideal if we can only test this by looking at logs, but I think there is another way.

You can check for received SIP messages by making this test class a subclass of the `SipTest` class. Have a look at `DeregistrationTaskTest` and in particular the `ClearMultipleImpis` test to see an example of this. If you make this change and add some user data in the `irs_info` that you return via `get_registration_data`, you can check that a 3rd party deregister is sent to the right application server.

I think we should try and make this change as part of this fix, but if it's taking too long, we can consider improving the UTs in the next release.

Ying-metaswitch: Well I wrapped a fake hss around mock hss to set_impu_result, but still can't use it in the config without failing all existing tests. Will come and talk to you about it.

src/ut/chronoshandlers_test.cpp
```diff
@@ -390,6 +398,64 @@ class ChronosAoRTimeoutTasksMockStoreTest : public SipTest
 
 };
 
+
+TEST_F(ChronosAoRTimeoutTasksMockStoreTest, DeregisterAS)
+{
+  // Test that deregister is sent to AS when all bindings expired
```
sathiyan-sivathas: Can you put this comment above the `TEST_F` line to match the other tests.

src/ut/chronoshandlers_test.cpp
```diff
@@ -390,6 +398,64 @@ class ChronosAoRTimeoutTasksMockStoreTest : public SipTest
...
+                              "        <Extension></Extension>\n"
+                              "      </SPT>\n"
+                              "    </TriggerPoint>\n"
+                              "    <ApplicationServer>\n"
+                              "      <ServerName>sip:1.2.3.4:56789;transport=UDP</ServerName>\n"
+                              "      <DefaultHandling>1</DefaultHandling>\n"
+                              "    </ApplicationServer>\n"
+                              "  </InitialFilterCriteria>\n"
+                              "</ServiceProfile></IMSSubscription>");
+  TransportFlow tpAS(TransportFlow::Protocol::UDP, stack_data.scscf_port, "1.2.3.4", 56789);
```
sathiyan-sivathas: You don't use this transport object in this test, so you can delete it.

src/ut/chronoshandlers_test.cpp
```diff
@@ -390,6 +398,64 @@ class ChronosAoRTimeoutTasksMockStoreTest : public SipTest
...
+
+  // Parse and handle the request
+  std::string body = "{\"aor_id\": \"sip:6505550231@homedomain\"}";
+  int status = handler->parse_response(body);
+
+  ASSERT_EQ(status, 200);
+
+  handler->handle_response();
+
+  // Expect a 3rd-party deregister to be sent to the AS in the iFCs
```
sathiyan-sivathas: You haven't actually checked that the deregister is sent to the right AS. You can do that by checking that the top route header in the received message contains `1.2.3.4:56789`.

PR #2040, opened 2017-12-05
Fix up expect_target function in UTs

KristaHyer:
This is the start of the fix to Metaswitch/clearwater-issues#2825.

I have edited expect_target() so that it **always** checks the target, (either strictly, or less strictly).
I have also fixed up some tests, which were incorrect and so failed once the new function was added.

This fix cannot be checked in, as around 30 tests still fail.
More work is required to look into those failures and either:
- fix up the tests so they pass
- fix the bugs in the code causing the tests to fail


KristaHyer:
New PR with edits on correct branch - closing this PR

PR #2042, opened 2017-12-07
[Reviewer RJW2] Quote pcfa params

steelyvoid:
Hi Richard

Are you OK to review this change for me (we discussed last week - its the fact that we don't quote "gen-values" of Charging function params in PCFA headers, contrary to the BNF in RFCs 3588 and 3261)? It would be good if you could find time to do this this week as the MMF workaround we gave to the customer might not work (as it would need to differentiate between ';' chars that are part of the values and not ccf/ecf parm separators).

I've taken care to quote only values that contain non-host/token characters (making the assumption that any value that starts with '[' is an IPv6 address and therefore doesn't need quoting, as is made elsewhere in sip_parser).

Note that I had to add a _hss_connection->set_impu_result to TEST_F(RegistrarTest, SimpleMainlinePassthrough), rather than change the existing _hss_connection->set_impu_result_with_prev call.  It seems that Sprout queries the regdata _without_ the ?private_id parameter when its trying to get CF values.

Tested live with a CCF value of aaa://hostname;transport=TCP as well as the UT updates below.

cpp-common PR to follow

### DIFF
--- a/src/pjutils.cpp
+++ b/src/pjutils.cpp
@@ -2166,12 +2166,7 @@ void PJUtils::add_pcfa_header(pjsip_msg* msg,
          ++it)
     {
       TRC_DEBUG("Adding CCF %s to PCFA header", it->c_str());
-      pjsip_param* new_param =
-        (pjsip_param*)pj_pool_alloc(pool, sizeof(pjsip_param));
-      new_param->name = STR_CCF;
-      new_param->value = pj_strdup3(pool, it->c_str());
-
-      pj_list_insert_before(&pcfa_hdr->ccf, new_param);
+      add_pcfa_param(&pcfa_hdr->ccf, pool, STR_CCF, *it);      
     }
 
     for (std::deque<std::string>::const_iterator it = ecfs.begin();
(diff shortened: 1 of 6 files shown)


steelyvoid:
One possible tweak - https://github.com/Metaswitch/sprout/pull/2042/files#diff-c6f4a4a957deddf8a50e0882c2bfd44aR1017 pre-quotes the ccf value, but I spotted in a SAS flow that XML would normally have &quot; escape sequences here.

I've checked that we don't get spurious double quoting in this case.  I'll commit a tweak so that we test this

src/pjutils.cpp
```diff
@@ -2166,30 +2166,69 @@ void PJUtils::add_pcfa_header(pjsip_msg* msg,
...
-      pj_list_insert_before(&pcfa_hdr->ecf, new_param);
+      add_pcfa_param(&pcfa_hdr->ecf, pool, STR_ECF, *it);      
     }
+
     pjsip_msg_add_hdr(msg, (pjsip_hdr*)pcfa_hdr);
   }
 }
 
+// Add Changing Function param to the list for a PCFA header
+void PJUtils::add_pcfa_param(pj_list_type *cf_list,
```
richardwhiuk: We could add some UTs for this function directly, instead of having to go through the rest of the header code.

steelyvoid: Will do

PR #2049, opened 2017-12-08
Trim whitespace when matching contact filtering tokens

metaswitch-colin-haywood:
What the title says. From an internal Metaswitch issue:

> When a Polycom VVX300 (and possibly other Polycom phones) registers, it signals RFC3840 feature
> support in the Contact header, by sending a Contact of the form:
> 
> Contact: <sip:5932224461@10.248.255.248:5060>;methods="INVITE, ACK, BYE, CANCEL, OPTIONS, INFO, MESSAGE, SUBSCRIBE, NOTIFY, PRACK, UPDATE, REFER"
> 
> ...
> 
> This causes trouble in an IMS network, where Clearwater fails to match some types of method to the 
> supported methods list (on the grounds that, say, " NOTIFY" with the leading space isn't the same as 
> "NOTIFY" without the space), concludes that the Polycom doesn't support this method, and doesn't send 
> the message to that contact address (and, if the Polycom is the only registered contact address, the 
> message gets rejected).

### DIFF
--- a/src/ut/contact_filtering_test.cpp
+++ b/src/ut/contact_filtering_test.cpp
@@ -104,7 +104,13 @@ TEST_F(ContactFilteringMatchNumericTest, InvalidNumericBackwardsRange) { EXPECT_
 typedef ContactFilteringTest ContactFilteringMatchTokensTest;
 TEST_F(ContactFilteringMatchTokensTest, MatchingTokens) { EXPECT_EQ(YES, match_tokens("hello,world", "goodbye,cruel,world")); }
 TEST_F(ContactFilteringMatchTokensTest, NonMatchingTokens) { EXPECT_EQ(NO, match_tokens("hello,dave", "i,cant,let,you,do,that")); }
-TEST_F(ContactFilteringMatchTokensTest, MatchingTokensCaseInsensitive) { EXPECT_EQ(YES, match_tokens("hello,dave", "Hello,is,it,me,your,looking,for")); }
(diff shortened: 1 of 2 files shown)


src/ut/contact_filtering_test.cpp
```diff
@@ -104,7 +104,13 @@ TEST_F(ContactFilteringMatchNumericTest, InvalidNumericBackwardsRange) { EXPECT_
 typedef ContactFilteringTest ContactFilteringMatchTokensTest;
 TEST_F(ContactFilteringMatchTokensTest, MatchingTokens) { EXPECT_EQ(YES, match_tokens("hello,world", "goodbye,cruel,world")); }
 TEST_F(ContactFilteringMatchTokensTest, NonMatchingTokens) { EXPECT_EQ(NO, match_tokens("hello,dave", "i,cant,let,you,do,that")); }
-TEST_F(ContactFilteringMatchTokensTest, MatchingTokensCaseInsensitive) { EXPECT_EQ(YES, match_tokens("hello,dave", "Hello,is,it,me,your,looking,for")); }
+TEST_F(ContactFilteringMatchTokensTest, MatchingTokensCaseInsensitive) { EXPECT_EQ(YES, match_tokens("hello,dave", "Hello,is,it,me,youre,looking,for")); }
+TEST_F(ContactFilteringMatchTokensTest, MatchingTokensWithWhitespace)
+{
+  EXPECT_EQ(YES, match_tokens("hello, goodbye , yes,no ,maybe", "goodbye,norma,jean"));
+  EXPECT_EQ(YES, match_tokens("hello, goodbye , yes,no ,maybe", "you,say,yes"));
+  EXPECT_EQ(YES, match_tokens("hello, goodbye , yes,no ,maybe", "I,say,no"));
```
AlexHockey: Worth a test that the 2nd argument also has whitespace stripped from it?

PR #2050, opened 2017-12-08
[Reviewer: AJH] Make sure get_routing_uri returns a SIP URI

sathiyan-sivathas:
Alex, this PR provides a fix for a scenario where we could crash in the Sproutlet Proxy if we receive a certain type of SIP message. The conditions required for the crash are:
 - We receive a SIP message with no route header and a Tel URI in the request line.
 - A Sproutlet is matched using a port since there is no SIP URI to match on (and Sproutlets can't be matched on Tel URIs).
 - The Sproutlet is uninterested and forwards it onto another sproutlet by calling `get_routing_uri()`.

Before this fix,  `get_routing_uri()` would return the Tel URI in the request line and cast to a SIP URI which would cause a crash. The fix is to fall back to `_root_uri` in the cases where `get_routing_uri()` would return a Tel URI. Tested in UT by writing a test that crashes/passes before/after my fix and running the scenario that caused the original crash on a live system and verifying that there is no longer a crash.

As discussed, we should also be checking that the URI we return is reflexive. I've added a TODO for that, which I'll address next week.

### DIFF
--- a/src/ut/sproutletproxy_test.cpp
+++ b/src/ut/sproutletproxy_test.cpp
@@ -14,6 +14,7 @@
 #include "test_interposer.hpp"
 #include "sproutletproxy.h"
 #include "compositesproutlet.h"
+#include "custom_headers.h"
 #include "pjutils.h"
 #include "pjsip.h"
 #include "pjsip_simple.h"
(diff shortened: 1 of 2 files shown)


src/ut/sproutletproxy_test.cpp
```diff
@@ -2430,6 +2461,65 @@ TEST_F(SproutletProxyTest, CompositeNetworkFunctionTelURI)
...
+  ASSERT_EQ(1, txdata_count());
+
+  // Check the 200 OK.
+  tdata = current_txdata();
+  RespMatcher(200).matches(tdata->msg);
+  tp->expect_target(tdata);
+  free_txdata();
+
+  // All done!
+  ASSERT_EQ(0, txdata_count());
```
AlexHockey: Does this test check that the request has actually been routed to the teltest2 sproutlet? Should it?

PR #2053, opened 2017-12-11
[Reviewer: AJH] Options polling to Sprout with URI parameter "user=phone" included

metaswitch-ak10:
Hi Alex, would you mind reviewing this fix?

This is a simple fix to URI classifier. Sprout classified local node URIs, such as those used for Options polls, incorrectly when the "user=phone" URI parameter was spuriously specified. 
I have also had to modify the UT framework a bit by allowing to send requests containing no user part in the SIP URI (so that the '@' sign is not included).

I have written a new UT to test my changes and all UTs passed, as well as run a live test on our L3 test system (running v11.0.2).

### DIFF
--- a/src/uri_classifier.cpp
+++ b/src/uri_classifier.cpp
@@ -160,17 +160,6 @@ URIClass URIClassifier::classify_uri(const pjsip_uri* uri, bool prefer_sip, bool
           ret = enforce_global ? LOCAL_PHONE_NUMBER : GLOBAL_PHONE_NUMBER;
           classified = true;
         }
-        else
-        {
-          classified = false;
-          ret = (local_to_node) ? NODE_LOCAL_SIP_URI : HOME_DOMAIN_SIP_URI;
-        }
-      }
-      else
-      {
-        classified = false;
-        ret = (local_to_node) ? NODE_LOCAL_SIP_URI : HOME_DOMAIN_SIP_URI;
-      }
     }
     if (!classified)
     {
--- a/src/uri_classifier.cpp
+++ b/src/uri_classifier.cpp
@@ -160,6 +160,7 @@ URIClass URIClassifier::classify_uri(const pjsip_uri* uri, bool prefer_sip, bool
           ret = enforce_global ? LOCAL_PHONE_NUMBER : GLOBAL_PHONE_NUMBER;
           classified = true;
         }
+      }
     }
     if (!classified)
     {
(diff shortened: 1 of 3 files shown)


src/ut/common_sip_processing_test.cpp
```diff
@@ -181,6 +181,105 @@ static pjsip_module mod_reject =
...
+  msg2._method = "OPTIONS";
+  msg2._requri = std::string("sip:poll-sip@127.0.0.1:") + std::to_string(ICSCF_PORT) + std::string(";user=phone");
+  msg2._via = "127.0.0.1";
+  msg2._todomain = std::string("127.0.0.1:") + std::to_string(ICSCF_PORT);
+  msg2._to = "poll-sip";
+  msg2._fromdomain = "127.0.0.1";
+  msg2._from = msg2._to;
+  msg2._contentlength = false;
+  msg2._extra = "Contact: <sip:127.0.0.1>\nAccept: application/sdp\nContent-Length: 0";
+  inject_msg(msg2.get_request(), _tp);
```
AlexHockey: This is pretty copy-paste. Might be better to just re-use the same `Message` object and modify the req-uri in place.

src/uri_classifier.cpp
```diff
@@ -159,12 +159,12 @@ URIClass URIClassifier::classify_uri(const pjsip_uri* uri, bool prefer_sip, bool
         }
         else
         {
-          ret = HOME_DOMAIN_SIP_URI;
+          ret = (local_to_node) ? NODE_LOCAL_SIP_URI : HOME_DOMAIN_SIP_URI;
```
AlexHockey: Should we really be implementing the logic that begins on line 171? i.e. 

* If home domain => HOME_DOMAIN_SIP_URI
* Else if local to node => NODE_LOCAL_SIP_URI
* Else => OFFNET_SIP_URI

AlexHockey:
p.s. sorry it's taken me a few days to get round to this - it's been a busy few days 😓

src/uri_classifier.cpp
metaswitch-ak10: I have checked with Steve and he agrees to make the behaviour consistent, so I will do that.

src/ut/common_sip_processing_test.cpp
metaswitch-ak10: Good idea.

metaswitch-ak10:
Hi @AlexHockey, I have made the changes you suggested. Would you mind reviewing them?

PR #2054, opened 2017-12-13
Move Memento from a submodule to a plugin [Reviewer: EM]

tomwatson1024:
This pull request moves the Memento sproutlet from a submodule of the Sprout repo to a plugin used in the same way as greeter/houdini. This means that we'll no longer have to build thrift/cassandra and memento itself when building Sprout, unless we actually want Memento.


src/stack.cpp
```diff
@@ -114,6 +114,8 @@ const static std::string _known_statnames[] = {
   "cdiv_not_registered",
   "cdiv_no_answer",
   "cdiv_not_reachable",
+  // The following stats are used by memento-as. Since memento has not yet been
```
eleanor-merry: Why?

PR #2055, opened 2017-12-13
[Reviewer: Seb] Update rogers dependency

richardwhiuk:
The AIO node doesn't currently install rogers - as it installs `ellis-node bono-node restund sprout-node homer-node homestead-node clearwater-prov-tools` (as per https://github.com/Metaswitch/clearwater-infrastructure/blob/master/scripts/clearwater-aio-install.sh#L48 ).

This updates sprout-node to reference `rogers`, which it should anyway, which fixes that.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -24,7 +24,7 @@ Description: Debugging symbols for sprout-libs
 
 Package: sprout-node
 Architecture: any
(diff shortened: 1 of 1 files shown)


sebrexmetaswitch:
Looks like it'll work, but it's confusing that we still have the `sprout-node` etc. packages. Feels like maybe we should remove these `-node` packages completely, but that's probably a worry for another day.

PR #2057, opened 2017-12-15
[Reviewer: Tom] Fix up incorrect log lines

sebrexmetaswitch:
Noticed a couple of errors in log lines that I think you added:

1) typo in "Received"
2) both these logs erroneously claimed to be on a worker thread, when they're actually on the transport thread. Since we log the thread id anyway, I figured we may as well just remove that part of the log.


tomwatson1024:
Looks good 😺

PR #2059, opened 2017-12-18
[Reviewer: Richard] Ensure rx_fork_error is run on a worker thread

sebrexmetaswitch:
This is a fix for issue https://github.com/Metaswitch/clearwater-issues/issues/2846

When Sprout can't contact an AS, we get an `on_tsx_state` callback on the transport thread. This can then cause blocking work, as we call into the SproutletProxy's `on_client_not_responding` method.

This fix ensures that call ends up on a worker thread.

We create a `Callback` object which can be run on the worker thread, which takes the tsx and a std::function (which is the actual work to do). I went for a std::function as I suspect we may have to do more of this in the future if we find more places where we're doing blocking work on the transport thread. (A likely looking place is https://github.com/Metaswitch/sprout/blob/dev/src/basicproxy.cpp#L2122, but I couldn't get a reliable repro so didn't want to make the change if I couldn't test it.)

We add a count of `_pending_callbacks` to the `UASTsx` so that we can ensure the tsx isn't destoryed until we've run all outstanding callbacks.

### Testing
* `make full_test` passes
* I ran stress through a system with IFCs pointing at a non-existent AS, and saw:
    * before this change, we had loads of logs showing blocking work on the transport thread, and lots of calls rejected due to overload
    * after this change, we didn't have any


richardwhiuk:
Alex, this feels sufficiently interesting that another pair of eyes sounds like good idea - can you take a look?

src/sproutletproxy.cpp
```diff
@@ -830,14 +851,28 @@ void SproutletProxy::UASTsx::on_client_not_responding(UACTsx* uac_tsx,
...
-    // Schedule any requests generated by the Sproutlet.
-    schedule_requests();
+      // Schedule any requests generated by the Sproutlet.
+      this->schedule_requests();
+    };
+
+    Callback* cb = new Callback(this, fn);
+
+    // This prevents this UASTsx from being destroyed until this Callback is run
+    _pending_callbacks++;
```
AlexHockey: It would be nice to encapsulate this so that if you create a callback in the sproutlet proxy you cannot forget to increment the count (in the same way you can't forget to decrements the count because the callback does it for you).

src/sproutletproxy.cpp
```diff
@@ -830,14 +851,28 @@ void SproutletProxy::UASTsx::on_client_not_responding(UACTsx* uac_tsx,
...
 
-    TRC_VERBOSE("Notifying upstream sproutlet %s of client failure: %s",
-                 upstream_sproutlet->service_name().c_str(),
-                 fork_error_to_str(fork_error));
 
-    upstream_sproutlet->rx_fork_error(fork_error, upstream_fork);
+    // We must ensure that the upstream Sproutlet handles this on a worker
+    // thread, so we create a Callback which we'll place on the worker thread
+    // queue if we're not on a worker thread already
+    std::function<void()> fn = [this, upstream_sproutlet, fork_error, upstream_fork]() -> void
```
AlexHockey: It is probably more idionamtic to do

```cpp
Callback* cb = new Callback(this, [this, upstream_sproutlet, fork_error, upstream_fork]()
{
  // ...
});

PR #2060, opened 2017-12-20
[Reviewer: AJH] Check URI is reflexive when getting routing URI

sathiyan-sivathas:
Alex,

This PR enhances `get_routing_uri()` to check that the URI it finds on a request is reflexive before returning it as a routing URI. Please can you review. This addresses the scenario where a message can contain a non-local SIP URI that we then use to route between Sproutlets which will fail.

I've had to change the Sproutlet API slightly with this change but the impacts are quite small. I've added UTs to get full coverage in this area of code.

Thanks,

Sathiyan

### DIFF
--- a/include/sproutlet.h
+++ b/include/sproutlet.h
@@ -371,8 +371,7 @@ public:
   virtual SAS::TrailId trail() const = 0;
 
   /// Get the URI that caused us to be routed to this Sproutlet or if no such
-  /// URI exists e.g. if the Sproutlet was matched on a port, return the root
-  /// URI.
+  /// URI exists e.g. if the Sproutlet was matched on a port, return NULL.
   ///
   /// @returns            - The URI that routed to this Sproutlet.
   ///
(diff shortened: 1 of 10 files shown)


AlexHockey:
I thought when we discussed this we were going to have `get_routing_uri` return NULL if the routing URI wasn't reflexive. That was why we deferred the fix - due to the risk of causing crashes if callers didn't check the return value of the function. 

I think the behavior better matches the description of the function: "Get the URI that caused us to be routed to this Sproutlet." If there was no such URI it should return nothing, not make one up.

sathiyan-sivathas:
I was kind of hoping that I wouldn't have to check all the calling code, but I agree that your idea better fits the description of this function. I'll make the change and resubmit.

sathiyan-sivathas:
I've made the changes you suggested. I had to make some of the calling code check the return value of the function and I changed one of the existing UTs because it wasn't working correctly before. Could you take another look?

PR #2061, opened 2017-12-20
[Reviewer: Krista] Fix up static analysis warnings

eleanor-merry:
Fixes to static analysis warnings and some general tidy up


eleanor-merry:
Yup - and I'm going to spin up a chef deployment and run the live tests over it.

PR #2062, opened 2018-01-02
[Reviewer: Krista] Allow URNs of the form service.*:sos.*

eleanor-merry:
This PR extends the set of emergency URNs we always accept, to match up with the RFCs.

Tested in UTs only


src/thread_dispatcher.cpp
```diff
@@ -48,6 +49,8 @@ extern "C" {
 #include "snmp_event_accumulator_by_scope_table.h"
 #include "thread_dispatcher.h"
 
+static const boost::regex EMERGENCY_SERVICES_URI = boost::regex("service.*:sos.*");
```
sathiyan-sivathas: Do you want a case-insensitive regex here? I remember I didn't do this when I wrote the original code and seeing this reminded me.

src/ut/thread_dispatcher_test.cpp
```diff
@@ -327,13 +294,76 @@ TEST_F(ThreadDispatcherTest, NeverRejectUrnServiceSosTest)
...
+  test_load_monitor_checks_on_requests(msg, false);
+}
+
+// On recieving a message with urn:services:sas in the request URI, the thread
+// dispatcher treat this message the same as any other non-priotised message.
+TEST_F(ThreadDispatcherTest, UrnNonServicesSosTest)
+{
+  TestingCommon::Message msg;
+  msg._method = "MESSAGE";
+  msg._requri = "urn:services:sas";
```
KristaHyer: Can we use something a bit different from "sos" so it's quickly clear what these tests are changing?

src/ut/thread_dispatcher_test.cpp
```diff
@@ -309,13 +282,7 @@ TEST_F(ThreadDispatcherTest, NeverRejectRegisterTest)
...
   msg._method = "REGISTER";
 
-  EXPECT_CALL(*mod_mock, on_rx_request(_)).WillOnce(Return(PJ_TRUE));
-  EXPECT_CALL(load_monitor, get_target_latency_us()).WillOnce(Return(100000));
-  EXPECT_CALL(load_monitor, admit_request(_, true)).WillOnce(Return(true));
-  EXPECT_CALL(load_monitor, request_complete(_, _));
-
-  inject_msg_thread(msg.get_request());
-  process_queue_element();
+  test_load_monitor_checks_on_requests(msg, true);
```
KristaHyer: Love this tidying!

src/thread_dispatcher.cpp
eleanor-merry: Good point, changed

src/ut/thread_dispatcher_test.cpp
eleanor-merry: I've changed these to notsos

PR #2064, opened 2018-01-04
[Reviewer: Krista] Accept-Contact fixes

rkd-msw:
Picking you because I think you've done a chunk of Sprout UT work recently and I think the UT changes are the most interesting part of this PR - if I've misremembered or you have too much on your plate, let me know.

L3 raised on Slack (I don't think there's an issue yet) that our Accept-Contact header parsing doesn't accept comma-separated values - `Accept-Contact: *;foo,*;bar` should be equivalent to:

```
Accept-Contact: *;foo
Accept-Contact: *;bar
```

but it fails to parse. Generally I'm a bit worried about this - I think we've hit a few header parsing bugs in our custom headers recently - so I've tried to figure out how to improve the testing here.

This PR:

* adds a specific test for comma-separated values, `AcceptContactMultiple`
* fixes up the Accept-Contact parser to make it work, using the same loop approach as PJSIP's code for Contact headers, which already correctly handles comma-separated values
* adds some helper methods to the tests, like `parse_and_print_multi`, to make them shorter and more readable
* adds several randomly-generated Accept-Contact tests created with abnfgen (http://www.quut.com/abnfgen/) to try and ensure there aren't any further parser bugs lurking - some of these do fail, so they're disabled, and I'll raise issues for those.

If this looks useful I might give a lightning talk on abnfgen.

Tagging @steelyvoid who raised the issue and @metaswitch-ak10 who I think is doing something similar for Reject-Contact, which has a similar bug.

### DIFF
--- a/src/ut/sip_parser_test.cpp
+++ b/src/ut/sip_parser_test.cpp
@@ -22,6 +22,13 @@ using namespace std;
 
 #define EXPECT_PJEQ(X, Y) EXPECT_EQ(PJUtils::pj_str_to_string(&X), string(Y))
 
+enum CloneType
+{
+  None,
+  Shallow,
+  Full
+};
+
 /// Fixture for SIP Parser testing
 class SipParserTest : public SipTest
 {
@@ -35,6 +42,10 @@ public:
   {
     SipTest::TearDownTestCase();
   }
+
+  std::vector<std::string> parse_and_print_multi(std::string header, std::string hname, CloneType ct = CloneType::None);
+  std::string parse_and_print_one(std::string header, std::string hname, CloneType ct = CloneType::None);
+  bool verify_parses(std::string header, std::string hname);
 };
 
 TEST_F(SipParserTest, PChargingVector)
@@ -543,7 +554,13 @@ TEST_F(SipParserTest, SessionExpiresUAC)
   EXPECT_STREQ("Session-Expires: 600;refresher=uac;other-param=10;more-param=42", buf);
 }
 
-TEST_F(SipParserTest, AcceptContact)
+std::string SipParserTest::parse_and_print_one(std::string header, std::string hname, CloneType ct)
+{
+  std::vector<std::string> ret = parse_and_print_multi(header, hname, ct);
+  return ret[0];
+}
+
+std::vector<std::string> SipParserTest::parse_and_print_multi(std::string header, std::string hname, CloneType ct)
 {
   pj_pool_t *main_pool = pjsip_endpt_create_pool(stack_data.endpt, "rtd%p",
                                                  PJSIP_POOL_RDATA_LEN,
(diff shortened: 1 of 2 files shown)


rkd-msw:
https://github.com/Metaswitch/sprout/pull/2064/files?w=1 makes the custom_headers.cpp change look neater - a lot of it is indentation.

KristaHyer:
I can review this :) when do you want it by?

rkd-msw:
No particular deadline - whenever's convenient.

src/ut/sip_parser_test.cpp
```diff
@@ -35,6 +42,9 @@ class SipParserTest : public SipTest
   {
     SipTest::TearDownTestCase();
   }
+
+  std::vector<std::string> parse_and_print_multi(std::string header, std::string hname, CloneType ct = CloneType::None);
+  std::string parse_and_print_one(std::string header, std::string hname, CloneType ct = CloneType::None);
```
KristaHyer: I don't fully like these names, since we're not really printing anything in my eyes. Maybe `parse_and_return_one`?

src/ut/sip_parser_test.cpp
```diff
@@ -543,14 +553,123 @@ TEST_F(SipParserTest, SessionExpiresUAC)
...
+
+  if (ct != CloneType::Full)
+  {
+    pj_pool_release(main_pool);
+  }
+
+  return ret;
+}
+
+TEST_F(SipParserTest, AcceptContact)
```
KristaHyer: Can this test, and the one immediately below it (`AcceptContactMultiple`), not use the new parse_and_print functions?

src/ut/sip_parser_test.cpp
```diff
@@ -575,57 +694,186 @@ TEST_F(SipParserTest, AcceptContact)
...
+TEST_F(SipParserTest, DISABLED_AcceptContact_VariousABNF_NullCharacter)
+{
+  std::string with_null("A: *;b=\"\\\0qwertyuiopasdfghjkl\"\n", 31);
+  parse_and_print_multi(with_null,
+                        "Accept-Contact");
+}
+
+TEST_F(SipParserTest, AcceptContact_VariousABNF_ManyCommas)
+{
+      std::vector<std::string> expected = { "Accept-Contact: *;explicit", "Accept-Contact: *", "Accept-Contact: *", "Accept-Contact: *", "Accept-Contact: *" };
```
KristaHyer: alignment?

src/ut/sip_parser_test.cpp
```diff
@@ -575,57 +694,186 @@ TEST_F(SipParserTest, AcceptContact)
...
+}
+
+TEST_F(SipParserTest, AcceptContactQuotedPair)
+{
+  EXPECT_EQ("Accept-Contact: *;c=\"\\j\"",
+          parse_and_print_one("Accept-Contact: *;c=\"\\j\"\n",
+                              "Accept-Contact"));
+}
+
+TEST_F(SipParserTest, DISABLED_AcceptContact_QuotedStringLWS)
```
KristaHyer: For this test (and several others), I was wondering why there's no `EXPECT_EQ`?
Is the idea that they just don't throw an exception when parsing, and that is the pass criteria?
Or is it that these are the tests in which you have used abnfgen, so the output is randomly generated, so we can't check it against a hard-coded value?
Can you please add a comment to make it clear above these tests?

Also, isn't it a hole in our testing if we never confirm that the headers have been parsed correctly by using `EXPECT_EQ`? What if the parsing doesn't throw an exception, but has still been done incorrectly?

src/ut/sip_parser_test.cpp
```diff
@@ -575,57 +694,186 @@ TEST_F(SipParserTest, AcceptContact)
...
+  EXPECT_EQ(true, hdr->required_match);
   EXPECT_NE(true, hdr->explicit_match);
-  EXPECT_EQ(0u, pj_list_size(&hdr->feature_set));
+  EXPECT_EQ(1u, pj_list_size(&hdr->feature_set));
 
   pj_pool_release(main_pool);
 }
 
+
+
```
KristaHyer: Too many blank lines.

src/ut/sip_parser_test.cpp
rkd-msw: We are printing - on line 646 we call the header's print_on function to turn it back into a string. We're not printing to screen, but we're doing the same sort of thing that e.g. snprintf is doing.

I'd think a function called `parse_and_return_one` would return the parsed pjsip_hdr struct - I haven't done that because it's more clunky to check (and doesn't test that the print_on function works properly!).

src/ut/sip_parser_test.cpp
rkd-msw: It could, but I think it's worth having a couple of (longer) tests where we verify that the parsed pjsip_hdr structs have the right fields and worked as expected. (For example, you could imagine a bug where 'explicit' and 'require' were flipped around in both the parsing and the print code, so it looked OK when printed again but hdr->required was false when it should be true.)

src/ut/sip_parser_test.cpp
rkd-msw: So, I *think* the only tests where that should be true are disabled tests that have exposed bugs. This is because:

* the print function normalises the header, so `A :	 *\n	 ;REqUIRe` becomes `Accept-Contact: *;require`
* rather than figure out what it will normalise it to, I've been running the test once, seeing what it does normalise it to, checking that's sensible, and adding an EXPECT_EQ
* where the header crashes the parser, I can't really do that

I'll add a comment to that effect, saying we should fix up the EXPECT_EQ if and when we fix the parser bug.

src/ut/sip_parser_test.cpp
KristaHyer: Fair enough, I'm happy with this name then

src/ut/sip_parser_test.cpp
KristaHyer: Sounds like a sensible reason to me

src/ut/sip_parser_test.cpp
KristaHyer: Ok this makes sense, thanks!

PR #2066, opened 2018-01-04
[Reviewer: AJH] Locality Optimization

tomwatson1024:
**Summary**

This pull request modifies the SIP routing behavior of Sprout so that the terminating half of call setup is handled by the site in which the callee is registered. This is to deal with potential conflicts between Clearwater's GR model and that of application servers, where there may be a significant cost associated with invoking an app server in a site different from that in which the user was registered.

It also maintains redundancy in the case that a site has failed, routing the INVITE to an alternate site if required.

**Altered Routing Behaviour**

Hostnames for which the node can process traffic are split (conceptually, and in configuration) into two categories. 'local' hostnames are those that correspond to services provided by this node that should be invoked in the mainline case; as an example, for a node in site 1, scscf.site1.homedomain would be a local hostname. 'remote' hostnames are those for which the node can handle traffic in the event of site failure; for a node in site 2, scscf.site1.homedomain would be a remote hostname.

When receiving SIP messages off the wire, a message with a route header matching either a remote or local hostname will be processed locally. The expectation is that a message with a remote hostname in the route header has been sent to this node as a result of a site failure.

When receiving SIP messages from a local Sproutlet, we consider whether or not the message will cross a network function boundary (or cross from the originating side of the S-CSCF to the terminating side). If so, only messages with route headers matching local hostnames will be processed locally, and other messages routed off the node. If not, the message is processed locally as before.

**Configuration**

Three config options are added. 

The options 'remote_alias_list' and 'local_alias_list', replace the existing 'alias_list' option. If the alias list option is used, a warning is presented to the user (during config upload and Sprout initialization) and the entries are processed as if they were on the local_alias_list. These options each contain a comma-separated list of hostnames. The local alias list is expected to contain the Sprout hostname of the local site, and the remote alias list the Sprout hostnames of remote sites. These options are set automatically when spinning up a new GR deployment from images; the remote alias list must be set manually when adding a new site to an existing deployment.

The option 'always_serve_remote_sites' is added. This option can be set to 'Y' or 'N'. If set to 'Y', the site will handle messages on behalf of remote sites in all cases, reproducing the current routing behavior; internally, the contents of both the local and remote alias lists in shared_config are added to the local_alias_list variable. If set to 'N', the altered routing behavior defined above will be used; internally, the contents of the local and remote alias lists will be added to the corresponding variables.

If not set, this option defaults to 'Y', maintaining the existing behavior.

**Monitoring**

Two new statistics are added:
- a simple count of the number of times a message has been routed off the node to a hostname that is a remote alias;
- a simple count of the number of times a message with a route header corresponding to a remote alias has been accepted by the node.

**Testing**

The existing unit tests pass under Valgrind, as do five additional tests. The Houdini UTs pass, using the companion PR for Houdini. Live tests of this feature have been added to the pipeline, and pass, along with the existing pipeline tests.


src/sproutletproxy.cpp
```diff
@@ -236,52 +260,85 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
...
+
   if (sproutlet == NULL)
   {
+    if (remote_match_rejected)
+    {
+      match_type = ALIAS_MATCH_TYPE::REMOTE_MATCH_REJECTED;
+    }
+    else
+    {
+    match_type = ALIAS_MATCH_TYPE::NO_MATCH;
```
AlexHockey: Alignment.

src/sproutletproxy.cpp
```diff
@@ -236,52 +260,85 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
...
     SAS::Event event(trail, SASEvent::NO_SPROUTLET_SELECTED, 0);
     SAS::report_event(event);
   }
+  else
+  {
+    match_type = ALIAS_MATCH_TYPE::MATCH;
+  }
 
-  return sproutlet;
+  return SproutletMatch(sproutlet, match_type);
```
AlexHockey: Would it be easier to keep a `SproutletMatch match` variable up to date as you go through the function, rather than maintain a flag and create a return variable just at the end?

src/sproutletproxy.cpp
```diff
@@ -236,52 +260,85 @@ Sproutlet* SproutletProxy::target_sproutlet(pjsip_msg* req,
...
-Sproutlet* SproutletProxy::match_sproutlet_from_uri(const pjsip_uri* uri,
-                                                    std::string& alias,
-                                                    std::string& local_hostname,
-                                                    SPROUTLET_SELECTION_TYPES& selection_type) const
+SproutletProxy::SproutletMatch SproutletProxy::match_sproutlet_from_uri(
+  const pjsip_uri* uri,
+  std::string& alias,
+  std::string& local_hostname,
+  SPROUTLET_SELECTION_TYPES& selection_type,
+  bool allow_remote_aliases) const
```
AlexHockey: Again, alignment looks weird.

src/sproutletproxy.cpp
```diff
@@ -377,9 +445,29 @@ Sproutlet* SproutletProxy::match_sproutlet_from_uri(const pjsip_uri* uri,
...
+  else if (remote_match_rejected)
+  {
+    match_type = ALIAS_MATCH_TYPE::REMOTE_MATCH_REJECTED; // LCOV_EXCL_LINE TJW2 TODO
+  }
+  else
+  {
+    match_type = ALIAS_MATCH_TYPE::NO_MATCH;
+  }
+
+  return SproutletMatch(sproutlet, match_type);
```
AlexHockey: Ditto. Can you maintain a `SproutletMatch` as you go along?

src/sproutletproxy.cpp
```diff
@@ -517,7 +603,8 @@ std::string SproutletProxy::get_local_hostname(const pjsip_sip_uri* uri) const
   (void*)match_sproutlet_from_uri((pjsip_uri*)uri,
                                   unused_alias,
                                   local_hostname,
-                                  unused_selection_type);
+                                  unused_selection_type,
+                                  true).first;
```
AlexHockey: Why do we not need to check the match type here? I can believe that we don't, but it wasn't obvious to me, so a comment would be useful at the very least.

src/sproutletproxy.cpp
```diff
@@ -530,26 +617,60 @@ std::string SproutletProxy::get_local_hostname(const pjsip_sip_uri* uri) const
...
+// on hosts that are 'remote' but not 'local' aliases.
+//
+// 'local aliases' are hostnames which correspond to services provided by this
+// node. For example, if this node is in site 1, scscf.site-1.homedomain.com is
+// a local alias.
+// Requests for local aliases of this node will be treated as local throughout.
+//
+// 'remote aliases' are hostnames which correspond to services provided by
+// remote nodes for which this node can accept traffic. For example, if this
+// node is in site 2, scscf.site-2.homedomain.com may be a GR alias. Requests
```
AlexHockey: Surely if the node is in site2, a site2 domain is a local alias?

AlexHockey: Again, you're switching between GR and remote.

src/sproutletproxy.cpp
```diff
@@ -679,9 +803,17 @@ pj_status_t SproutletProxy::UASTsx::init(pjsip_rx_data* rdata)
     Sproutlet* sproutlet = NULL;
     pjsip_route_hdr* route = (pjsip_route_hdr*)
                 pjsip_msg_find_hdr(rdata->msg_info.msg, PJSIP_H_ROUTE, NULL);
+
+    // For the purpose of redundancy, we allow GR aliases here, that is, accept
+    // requests intended for a GR alias of this site off the wire. This means
+    // that this node can handle traffic destined for a remote site in the case
+    // that the remote site has failed.
+    bool discard;
```
AlexHockey: A name like "unused" might be better here.

src/sproutletproxy.cpp
```diff
@@ -1235,15 +1381,27 @@ void SproutletProxy::UASTsx::check_destroy()
...
 
   // Do an initial lookup for the target sproutlet.
-  Sproutlet* sproutlet = _sproutlet_proxy->target_sproutlet(req->msg,
+  SproutletMatch match = _sproutlet_proxy->target_sproutlet(req->msg,
                                                             port,
                                                             alias,
+                                                            allow_remote_aliases,
                                                             trail());
+  if ((match.first == NULL) &&
+      (match.second == ALIAS_MATCH_TYPE::REMOTE_MATCH_REJECTED))
```
AlexHockey: Why do you even need to check the first parameter?

src/sproutletproxy.cpp
```diff
@@ -1235,15 +1381,27 @@ void SproutletProxy::UASTsx::check_destroy()
...
+  {
+    remote_match_rejected = true;
```
AlexHockey: This variable doesn't seem to be used.

src/sproutletproxy.cpp
```diff
@@ -1841,7 +2006,7 @@ bool SproutletWrapper::is_uri_reflexive(const pjsip_uri* uri) const
 
 bool SproutletWrapper::is_uri_local(const pjsip_uri* uri) const
 {
-  return _proxy->is_uri_local(uri);
+  return _proxy->is_uri_local(uri, true);
```
AlexHockey: Why is it right to assume this?

src/ut/session_expires_helper_test.cpp
```diff
@@ -269,10 +275,12 @@ class SessionExpiresHelperTest : public SipTest
 
 protected:
   TransportFlow* _tp;
+  static MockSnmpCounterTable* _mock_counter_table;
   static SproutletProxy* _proxy;
   static std::list<Sproutlet*> _sproutlets;
 };
 
+MockSnmpCounterTable* SessionExpiresHelperTest::_mock_counter_table;
```
AlexHockey: FFR, it might have been easier to allow the SproutletPRoxy to be passed nullptr for the new stat, and have it check before incrementing the count. That way you wouldn't have had to updated all these UTs.

src/ut/sproutletproxy_test.cpp
```diff
@@ -654,6 +655,37 @@ class FakeSproutletURIForwarder : public CompositeSproutletTsx
...
+  FakeSproutletGRForwarder(Sproutlet* sproutlet) :
+   CompositeSproutletTsx(sproutlet, static_cast<FakeSproutlet<FakeSproutletGRForwarder>*>(sproutlet)->_next_hop)
+  {
+  }
+
+  void on_rx_initial_request(pjsip_msg* req)
+  {
+    // Regardless of what we receive, forward the request on using a Route
+    // header to route the message to the specified (external) URI.  This is
+    // used to test Tel URIs, which are not themselves routable.
```
AlexHockey: Duff reference to Tel URIs.

src/ut/sproutletproxy_test.cpp
```diff
@@ -654,6 +655,37 @@ class FakeSproutletURIForwarder : public CompositeSproutletTsx
...
+    string forwarding_uri = "sip:proxy1.homedomain-remote;transport=TCP;lr";
```
AlexHockey: In order to test your code is working correctly, you need this URI to correspond to a sproutlet that would be matched if it were received on the wire. I don't think this URI matches anything though.

include/sproutletproxy.h
```diff
@@ -72,6 +78,16 @@ class SproutletProxy : public BasicProxy, SproutletHelper
...
   class UASTsx;
 
+  enum ALIAS_MATCH_TYPE
+  {
+    NO_MATCH,
+    MATCH,
+    REMOTE_MATCH_REJECTED
+  };
+
+  typedef std::pair<bool, ALIAS_MATCH_TYPE> AliasMatch;
```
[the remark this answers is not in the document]
AlexHockey: OK I figured it out. You shouldn't use this type. The first field is completely determinable from the second, so just pass the ALIAS_MATCH_TYPE around, and write a function to return whether it represents a match.

include/sproutletproxy.h
```diff
@@ -72,6 +78,16 @@ class SproutletProxy : public BasicProxy, SproutletHelper
   /// Pre-declaration
   class UASTsx;
 
+  enum ALIAS_MATCH_TYPE
```
AlexHockey: This is a type so it should be in CamelCase (`AliasMatchType`). We use SCREAMING_SNAKE_CASE for constants.

include/stack.h
```diff
@@ -145,7 +146,9 @@ extern pj_status_t init_stack(const std::string& sas_system_name,
                               const std::string& additional_home_domains,
                               const std::string& sproutlet_uri,
                               const std::string& sprout_hostname,
-                              const std::string& alias_hosts,
+                              const std::string& deprecated_alias_hosts,
```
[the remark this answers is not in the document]
tomwatson1024: I agree - the issue is that splitting the option out into lists is handled in stack.cpp, and I wanted to keep the code together.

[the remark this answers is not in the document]
AlexHockey: That's annoying. I don't care too much then.

sprout-base.root/etc/init.d/sprout
```diff
@@ -187,6 +189,8 @@ get_daemon_args()
         [ -z "$sprout_chronos_callback_uri" ] || sprout_chronos_callback_uri_arg="--sprout-chronos-callback-uri=$sprout_chronos_callback_uri"
         [ -z "$dummy_app_server" ] || dummy_app_server_arg="--dummy-app-server=$dummy_app_server"
         [ -z "$sprout_request_on_queue_timeout" ] || request_on_queue_timeout_arg="--request-on-queue-timeout=$sprout_request_on_queue_timeout"
+        [ -z "$alias_list" ] || deprecated_alias_list_arg="--alias=$alias_list"
+        [ "$always_serve_remote_aliases" != "Y" ] || always_serve_remote_aliases_arg ="--always-serve-remote-aliases"
```
graemerobertson: @tomwatson1024 there is a space in this line which means Sprout never starts with the option `--always-serve-remote-aliases`! Please can you fix?

include/sproutletproxy.h
```diff
@@ -30,22 +31,43 @@ class SproutletProxy : public BasicProxy, SproutletHelper
...
+  /// @param[in]  route_to_remote_alias_tbl    SNMP table for counting the
+  ///                                          number of requests routed to
+  ///                                          remote aliases.
+  /// @param[in]  accept_for_remote_alias_tbl  SNMP table for counting the
+  ///                                          number of requests for remote
+  ///                                          aliases accepted locally.
+  /// @param[in]  max_sproutlet_depth          The maximum number of Sproutlets
+  ///                                          that can be invoked in a row
+  ///                                          before we break the loop.
+  ///
```
AlexHockey: OK, so I see you've bundled the doxygen comments impediment task into this PR. This makes the PR harder to review because it is bigger. It also puts the actual story we care about at risk - I don't have time to review all your doxygen comments but I can't accept the PR with them in because I have seen some mistakes (and I would rather have no comments than misleading ones). Please can you revert the commit that adds them?

include/sproutletproxy.h
```diff
@@ -224,10 +435,74 @@ class SproutletProxy : public BasicProxy, SproutletHelper
...
+    /// @param[in]  alias                        The alias of match.sproutlet
+    ///                                          used to obtain the match.
+    /// @param[out] tsx_match_locality           The locality of the tsx match.
+    ///
+    /// @return     True if the match should be rejected, and false otherwise.
+    ///
+    bool accept_sproutlet_match(SproutletMatch match,
+                                bool always_serve_remote_aliases,
+                                std::string& alias,
+                                SproutletTsxMatchLocality& tsx_match_locality);
```
AlexHockey: I don't really see why we need this function. We only need to make a decision about whether to accept the match in one place in the code, and it is really trivial, so this function just obfuscates things.

include/sproutletproxy.h
```diff
@@ -224,10 +435,74 @@ class SproutletProxy : public BasicProxy, SproutletHelper
...
+    ///                                          locally in all cases.
+    ///
+    /// @return     The matched SproutletTsx, or NULL, and the match locality.
+    ///             See accept_sproutlet_match for details on tsx match
+    ///             localities.
+    ///
+    SproutletTsxMatch get_sproutlet_tsx(pjsip_tx_data* req,
+                                        int port,
+                                        std::string& alias,
+                                        bool always_serve_remote_aliases);
```
AlexHockey: To go along with my comments on the `SproutletTsxMatch`, I think you should reinstate the old version of this function.

include/sproutletproxy.h
tomwatson1024: Although this is only called from get_sproutlet_tsx, it is called in two places within that function - since there's now more complicated behavior here (determining if we're on a network function boundary, SAS logging etc) I didn't want to duplicate the code twice within get_sproutlet_tsx.

AlexHockey: Maybe it's the output parameter that is making this function seem overly complicated. If you can get rid of that, I think I'd be happy. .

include/sproutletproxy.h
tomwatson1024: I've reverted the a few of the doxygen comment changes. Most of them apply to functions which I've changed in order to implement locality optimization - since I'd have to update the comments anyway, I thought it made sense to do so in the doxygen style.

src/ut/sproutletproxy_test.cpp
```diff
@@ -60,6 +62,89 @@ class FakeSproutlet : public Sproutlet
...
 };
 
+// A SproutletTsx that routes the message to the specified forwarding URI.
+class FakeForwarderTsx: public SproutletTsx
+{
+public:
+  FakeForwarderTsx(Sproutlet* sproutlet,
+                   std::string forwarding_uri) :
+   SproutletTsx(sproutlet),
+   forwarding_uri(forwarding_uri)
```
AlexHockey: Wow - I had no idea C++ was smart enough for this to work - given the member and the argument have the same name.

src/ut/sproutletproxy_test.cpp
```diff
@@ -60,6 +62,89 @@ class FakeSproutlet : public Sproutlet
...
+
+    send_request(req);
+  }
+
+  std::string forwarding_uri;
+};
+
+// A Sproutlet that routes the message to a forwarding URI specified by a mock
+// function call. Mock functions also determine whether the Sproutlet is
+// interested, and if not, its next hop.
```
AlexHockey: Nice approach.

src/ut/sproutletproxy_test.cpp
tomwatson1024: should I change this?

AlexHockey: It works so no need to change it. I just didn't know that it was that smart! :-)

src/sproutletproxy.cpp
```diff
@@ -1269,17 +1331,102 @@ void SproutletProxy::UASTsx::check_destroy()
...
-                                                        std::string& alias)
+bool SproutletProxy::UASTsx::accept_sproutlet_match(SproutletMatch match,
+                                                    bool always_serve_remote_aliases,
+                                                    bool first_match_in_tsx,
+                                                    std::string& alias)
+{
+  if (match.match_locality == AliasMatchLocality::LOCAL)
+  {
+    // Local matches are always accepted.
+    SAS::Event event(trail(), SASEvent::SPROUTLET_ALIAS_MATCH_ACCEPT, 0);
```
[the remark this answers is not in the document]
tomwatson1024: I've modified this so that a SAS log is only generated when requests for remote aliases are accepted off the wire, i.e. not during Sproutlet processing.

AlexHockey:
LGTM

PR #2067, opened 2018-01-08
[Reviewer: EM] Define subscriber manager management API

sathiyan-sivathas:
Ellie,

I've defined the SM management API based on your design. Please could you review it. I've taken your draft file and made the following changes:
 - The binding/subscription constructors take binding/subscription IDs. They also set the underlying value for the ID before passing back i.e. `_uri` for bindings and `_to_tag` for subscriptions.
 - I've removed the timeout event trigger since all timeout events should be internal so S4 now.
 - Added the CSeq field to subscriptions.
 - Added a `get_id()` method to the binding and subscription objects.
 - Deleted the `_path_uris` field from binding because it is no longer required for upgrade.

Other than that I've added some SDM-REFACTOR-TODOs for things that I think will need to be done at some point. Probably worth checking you agree with these.


include/subscriber_manager.h
```diff
@@ -0,0 +1,200 @@
...
+  enum EventTrigger
+  {
+    USER,
+    ADMIN
+  };
+
+  class Binding
+  {
+  public:
+    Binding(std::string binding_id): _uri(binding_id) {};
```
eleanor-merry: I'm not *convinced* that this needs to take a binding ID

sathiyan-sivathas:
I've added a TODO for the doxygen comments. We discussed and agreed that the binding/subscription IDs didn't need to passed in the constructors.

PR #2069, opened 2018-01-08
first draft at commonising hss cache helper function

KristaHyer:
As we refactor the SDM, we are also refactoring the sproutlets.
Some of the sproutlets we refactor will need to use code from the scscf sproutlet.
This pull request moves that code from the scscf sproutlet and puts it in a common place so that it can be used by multiple sproutlets.

`make full_test` passes


KristaHyer:
@Ying-metaswitch 

I've looked into this more, and these are the edits that need to be made:
- change this from a class to just a set of functions
- change the owenrship of some data (such as the cached data) to be the sproulets
- change the read_hss_data function, so that it calls into the SM following the new architecture, instead of calling to the HSS iteself (via HSSConnection which goes through homestead)

I'm working on these now, alongside refactoring the SCSCFSproutlet.

src/contact_filtering.cpp
```diff
@@ -269,26 +268,10 @@ bool binding_to_target(const std::string& aor,
...
-        else
-        {
-          TRC_WARNING("Ignoring contact %s for target %s because of badly formed path URI %s",
-                      binding._uri.c_str(), aor.c_str(), (*path).c_str());
-          // TODO SAS log
-          valid = false;
-          break;
-        }
-      }
+      TRC_WARNING("Empty path headers field for contact %s, implying no path headers were present on a register",
```
KristaHyer: Not invalid to have no path headers. Change comment (after double checking with SS5)

PR #2077, opened 2018-01-12
Refactor S4 and dependencies into submodules.

AlexHockey:
Hi Ellie. This is the companion PR for https://github.com/Metaswitch/clearwater-s4/pull/1. It mostly just contains a whole bunch of deleted code, but it is worth you reviewing the added code. The most "controversial" thing I've done is removed from the Binding objects the functions for generating GRUUs. These now live as standalone functions in the `AoRUtils` namespace.


AlexHockey:
BTW, I have checked that sprout builds and the UTs run (barring some memory leaks in S4 ;-).

PR #2078, opened 2018-01-15
[Reviewer: RKD] Reject-Contact fixes

metaswitch-ak10:
Hi Rob,

This PR complements your Accept-Contact PR #2064. Consequently, I believe you are the best person to review this.

In this fix, I have:

- Fixed the Reject-Contact header parser to allow parsing of headers with multiple rc-values.

- Merged the routines for parsing Accept-Contact and Reject-Contact headers into one common routine.  I think this makes sense since the routines are virtually the same with only minor differences. having a common routine also reduces the possibility of an error when the parser get updated in the future.

- Added a UT for parsing Reject-Contact header with multiple rc-values. Note that by combining the two routines, the extensive UTs that you have added as part of your fix test the Reject-Contact code as well.

I have tested the fix by running make full_test;

### DIFF
--- a/src/ut/sip_parser_test.cpp
+++ b/src/ut/sip_parser_test.cpp
@@ -962,17 +962,17 @@ TEST_F(SipParserTest, RejectContactMultiple)
   parse_rxdata(rdata);
 
   pj_str_t header_name = pj_str("Reject-Contact");
-  pjsip_accept_contact_hdr* hdr =
-      (pjsip_accept_contact_hdr*)pjsip_msg_find_hdr_by_name(rdata->msg_info.msg,
+  pjsip_reject_contact_hdr* hdr =
+      (pjsip_reject_contact_hdr*)pjsip_msg_find_hdr_by_name(rdata->msg_info.msg,
                                                             &header_name,
                                                             NULL);
-  EXPECT_NE(hdr, (pjsip_accept_contact_hdr*)NULL);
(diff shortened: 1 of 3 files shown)


src/ut/sip_parser_test.cpp
```diff
@@ -941,6 +941,43 @@ TEST_F(SipParserTest, RejectContact)
...
+
+  string str("INVITE sip:6505554321@homedomain SIP/2.0\n"
+             "Via: SIP/2.0/TCP 10.0.0.1:5060;rport;branch=z9hG4bKPjPtVFjqo;alias\n"
+             "Max-Forwards: 63\n"
+             "From: <sip:6505551234@homedomain>;tag=1234\n"
+             "To: <sip:6505554321@homedomain>\n"
+             "Contact: <sip:6505551234@10.0.0.1:5060;transport=TCP;ob>\n"
+             "Call-ID: 1-13919@10.151.20.48\n"
+             "CSeq: 1 INVITE\n"
+             "Reject-Contact: *;+sip.instance=\"<i:am:a:robot>\";+xyz,*;+abcd\n"
```
rkd-msw: Might be worth testing 'explicit' and/or 'require' in here, as for Reject-Contact they should appear in the `feature_set`.

metaswitch-ak10: Ok, I will do that

PR #2082, opened 2018-01-16
[Reviewer: Seb] Don't leak PJSIP transactions sending SIP messages

richardwhiuk:
Seb,

If we send a SIP request to another SIP node using `PJUtils::send_request`, e.g. Third Party Registrations, if `pjsip_tsx_send_msg` fails, we can leak the `pjsip_tx_data`'s pool, which leaks a bunch of memory.

This is because we `pjsip_tsx_send_msg` will not decrement the reference count isn't decremented if pjsip_tsx_send_msg fails.

I've written further detail on exactly what semantics are of the memory management, and why this is the right thing to do inline. There's no current UTs for pjutils.

This is conceptually a follow on to https://github.com/Metaswitch/sprout/pull/2001

I've tested with a long low load, under massif, and I'm now running a longer stress test with a higher load to test this further.

### DIFF
--- a/src/pjutils.cpp
+++ b/src/pjutils.cpp
@@ -1648,6 +1648,18 @@ pj_status_t PJUtils::respond_stateful(pjsip_endpoint* endpt,
 
   status = pjsip_tsx_send_msg(uas_tsx, tdata);
 
+  if (status != PJ_SUCCESS)
+  {
+    // The message is owned by the transaction, which will get a on_tsx_state
+    // callback. However, we still have a reference count if tsx_send_msg
+    // fails, which we should decrement, to prevent a leak.
+    pjsip_tx_data_dec_ref(tdata);
+
+    // Even if we failed to send, we should treat it as a success, as the
+    // message may be resent by the transaction owner.
+    status = PJ_SUCCESS;
+  }
+
   return status;
 }
 
--- a/src/pjutils.cpp
+++ b/src/pjutils.cpp
@@ -1173,6 +1173,8 @@ static void on_tsx_state(pjsip_transaction* tsx, pjsip_event* event)
 
             // We no longer care about the old tdata.
             pjsip_tx_data_dec_ref(old_tdata);
+            old_tdata = nullptr;
+
             sss->tdata = tdata;
           }
           // LCOV_EXCL_STOP
@@ -1216,6 +1218,7 @@ static void on_tsx_state(pjsip_transaction* tsx, pjsip_event* event)
 
               // The same logic in send_request applies here too.
               pjsip_tx_data_dec_ref(tdata);
+              tdata = nullptr;
             }
           }
         }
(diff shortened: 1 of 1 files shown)


src/pjutils.cpp
```diff
@@ -1611,6 +1648,18 @@ pj_status_t PJUtils::respond_stateful(pjsip_endpoint* endpt,
 
   status = pjsip_tsx_send_msg(uas_tsx, tdata);
 
+  if (status != PJ_SUCCESS)
+  {
+    // The message is owned by the transaction, which will get a on_tsx_state
+    // callback. However, we still have a reference count if tsx_send_msg
+    // fails, which we should decrement, to prevent a leak.
+    pjsip_tx_data_dec_ref(tdata);
```
bossmc: Set to `NULL` to maximise the future explosion?

src/pjutils.cpp
```diff
@@ -1338,12 +1348,39 @@ pj_status_t PJUtils::send_request(pjsip_tx_data* tdata,
...
+      //   failure of tsx_send_msg
+      //
+      // - (3) The reference owned by the caller which was passed into this
+      //   function by the caller.
+      //
+      //   In the success case, tsx_send_msg will decrement this reference. In
+      //   the failure case, it won't. Thus, given on_tsx_state will handle
+      //   further processing, and to keep the interface to this function clean,
+      //   we should decrement the reference here.
+      pjsip_tx_data_dec_ref(tdata);
```
bossmc: Ditto (in reverse)

src/pjutils.cpp
```diff
@@ -1206,7 +1207,16 @@ static void on_tsx_state(pjsip_transaction* tsx, pjsip_event* event)
...
-            (void)pjsip_tsx_send_msg(retry_tsx, tdata);
+            pj_status_t tsx_status = pjsip_tsx_send_msg(retry_tsx, tdata);
+
+            if (tsx_status != PJ_SUCCESS)
+            {
+              TRC_DEBUG("Failed to to send retry: %s",
+                        PJUtils::pj_status_to_string(tsx_status).c_str());
+
+              // The same logic in send_request applies here too.
+              pjsip_tx_data_dec_ref(tdata);
```
bossmc: *cough*

sebrexmetaswitch:
As discussed, we've not tested that last fix, but we think that's ok because we think this is the correct fix anyway (and we've tested it in several other places)

src/pjutils.cpp
richardwhiuk: Done

src/pjutils.cpp
richardwhiuk: Done

src/pjutils.cpp
richardwhiuk: Done (well `nullptr` actually).

PR #2094, opened 2018-01-19
Add config option for ramrecorder function

kenferguson:


include/cfgoptions.h
```diff
@@ -141,6 +141,7 @@ struct options
   int                                  request_on_queue_timeout;
   std::set<std::string>                blacklisted_scscfs;
   bool                                 enable_orig_sip_to_tel_coerce;
+  bool                                 ramrecorder;
```
johadalin: Should this actually be a `pj_bool_t`? Not sure if the two are interchangable

johadalin:
Also, assuming we're doing some reasonable set of testing here to ensure it works as expected?

PR #2095, opened 2018-01-22
[Reviewer: AJH] fixed registrar behaviour regarding the inclusion of outbound tag in Require hdr

metaswitch-ak10:
Hi Alex, would you be happy reviewing this fix?

This PR fixes the behaviour of the registrar with respect to the inclusion of the "outbound" option tag in the Require header of a REGISTER response. The correct behaviour is described in RFC5626 (section 6).

I have tested my fix by running 'make full_test' and checking that all UTs pass and my modifications have 100% coverage.

Note that the registrar's current behaviour does not comply with the whole specification introduced in RFC5626. Here are some notable exceptions:

- When receiving a REGISTER request, the registrar MUST check from its
Via header field if the registrar is the first hop or not.  If the
registrar is not the first hop, it MUST examine the Path header of
the request.  If the Path header field is missing or it exists but
the first URI does not have an "ob" URI parameter, ... In this case,
the following processing applies: if the REGISTER request contains
the reg-id and the outbound option tag in a Supported header field,
then the registrar MUST respond to the REGISTER request with a 439
(First Hop Lacks Outbound Support) response.
- ... if the Contact header field contains
more than one header field value with a non-zero expiration and any
of these header field values contain a "reg-id" Contact header field
parameter, the entire registration SHOULD be rejected with a 400 (Bad
Request) response. The justification for recommending rejection
versus making it mandatory is that the receiver is allowed by
[RFC3261] to squelch (not respond to) excessively malformed or
malicious messages.


AlexHockey:
Before I review, can you remind me what bug this fixes? Or are we doing this just to be more spec-compliant?

metaswitch-ak10:
Prior to this fix, the registrar was incorrectly including the "outbound" option  tag in the "Require" header of a REGISTER response. I have fixed the behaviour according to RFC5626. 
This fixes the following bug:

In this flow, the UA does not indicate support for RFC5626 (no "Supported: outbound" header on the request).  Clearwater does send a "Require: outbound" on the 200 response, but Perimeta ignores it and just passes it through to the client because the client didn't advertise support for it.
 
This bug is part of an Issue raised by L3 support in response to Fixed Line IM testing with Telmex.

AlexHockey:
Planning to review tomorrow. Hope that's OK.

metaswitch-ak10:
That's fine, take your time, it's not that urgent.

src/registrarsproutlet.cpp
```diff
@@ -729,17 +749,54 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
+    {
+      contains_ob = true;
+    }
+
+    if (supported_outbound && contains_reg_id && contains_ob &&
+       !aor_pair->get_current()->bindings().empty())
+    {
+      // We include the outbound option tag in the Require header if and only if:
+      // 1. We have bindings with path headers
+      // 2. The first path header contains the "ob" parameter
```
AlexHockey: My reading of https://tools.ietf.org/html/rfc5626 is that the "first" path header refers to the first path header that was added to the message, not the top Path header. I base that on the description of the edge-proxy behavior from [section 5.1](https://tools.ietf.org/html/rfc5626#section-5.1):

>If the edge proxy is the first hop, as
   indicated by the Via header field, it MUST insert its URI in a Path
   header field value as described in [RFC3327].  If it is not the first
   hop, it might still decide to add itself to the Path header based on
   local policy

and 

>In addition, the first node MUST include an
   "ob" URI parameter in its Path header field value.

Summarizing this, the edge proxy must add a Path header with an "ob" parameter. Subsequent edge proxies may add path headers without the "ob" parameter. So the first Path header added will be the only one with the "ob" parameter, which will be at the end of the Path header list.

src/ut/registrar_test.cpp
```diff
@@ -517,8 +520,7 @@ class RegistrarObservedHssTest : public RegistrarTest
                 MatchesRegex("Contact: <sip:eeeebbbbaaaa11119c661a7acf228ed7@10.114.61.111:5061;transport=tcp;ob>;expires=(300|[1-2][0-9][0-9]|[1-9][0-9]|[1-9]);\\+sip.ice;\\+sip.instance=\"<urn:uuid:00000000-0000-0000-0000-a55444444440>\";reg-id=1;pub-gruu=\"sip:6505550231@homedomain;gr=urn:uuid:00000000-0000-0000-0000-a55444444440\"\r\n"
                              "Contact: <sip:f5cc3de4334589d89c661a7acf228ed7@10.114.61.213:5061;transport=tcp;ob>;expires=(300|[1-2][0-9][0-9]|[1-9][0-9]|[1-9]);\\+sip.ice;\\+sip.instance=\"<urn:uuid:00000000-0000-0000-0000-b665231f1213>\";reg-id=1;pub-gruu=\"sip:6505550231@homedomain;gr=urn:uuid:00000000-0000-0000-0000-b665231f1213\"\r\n"
                              "Contact: <sip:f5cc3de4334589d89c661a7acf228ed7@10.114.61.213:5061;transport=tcp;ob>;expires=(300|[1-2][0-9][0-9]|[1-9][0-9]|[1-9]);\\+sip.ice;reg-id=1"));
-
-    EXPECT_EQ("Require: outbound", get_headers(out, "Require")); // because we have path
+    EXPECT_EQ("", get_headers(out, "Require")); // We don't have +sip.instance
```
AlexHockey: It won;t be clear to someone reading this code that this is intended to test absence of a "Require: outbound" header. The comment should explain this (e.g. "We don't have +sip.instance so outbound not required")

src/ut/registrar_test.cpp
```diff
@@ -543,7 +545,7 @@ class RegistrarObservedHssTest : public RegistrarTest
                 MatchesRegex("Contact: <sip:eeeebbbbaaaa11119c661a7acf228ed7@10.114.61.111:5061;transport=tcp;ob>;expires=(300|[1-2][0-9][0-9]|[1-9][0-9]|[1-9]);\\+sip.ice;\\+sip.instance=\"<urn:uuid:00000000-0000-0000-0000-a55444444440>\";reg-id=1;pub-gruu=\"sip:6505550231@homedomain;gr=urn:uuid:00000000-0000-0000-0000-a55444444440\"\r\n"
                              "Contact: <sip:f5cc3de4334589d89c661a7acf228ed7@10.114.61.213:5061;transport=tcp;ob>;expires=(300|[1-2][0-9][0-9]|[1-9][0-9]|[1-9]);\\+sip.ice;\\+sip.instance=\"<urn:uuid:00000000-0000-0000-0000-b665231f1213>\";reg-id=1;pub-gruu=\"sip:6505550231@homedomain;gr=urn:uuid:00000000-0000-0000-0000-b665231f1213\"\r\n"
                              "Contact: <sip:f5cc3de4334589d89c661a7acf228ed7@10.114.61.213:5061;transport=tcp;ob>;expires=(300|[1-2][0-9][0-9]|[1-9][0-9]|[1-9]);\\+sip.ice;reg-id=1"));
-    EXPECT_EQ("Require: outbound", get_headers(out, "Require")); // because we have path
+    EXPECT_EQ("", get_headers(out, "Require")); // We don't have +sip.instance
```
AlexHockey: Ditto.

AlexHockey: And so on in this file.

src/registrarsproutlet.cpp
AlexHockey: You should also add a UT for this case.

src/ut/registrar_test.cpp
```diff
@@ -3332,6 +3334,127 @@ TEST_F(RegistrarTest, AllIMPUsBarred)
...
+  EXPECT_EQ("OK", str_pj(out->line.status.reason));
+  EXPECT_EQ("Supported: outbound", get_headers(out, "Supported"));
+  EXPECT_EQ("Contact: <sip:f5cc3de4334589d89c661a7acf228ed7@10.114.61.213:5061;transport=tcp;ob>;expires=300;+sip.ice;+sip.instance=\"<urn:uuid:00000000-0000-0000-0000-b665231f1213>\";reg-id=1;pub-gruu=\"sip:6505550231@homedomain;gr=urn:uuid:00000000-0000-0000-0000-b665231f1213\"",
+            get_headers(out, "Contact"));
+  EXPECT_EQ("Require: outbound", get_headers(out, "Require")); // We expect the outbound options tag in the Require header
+  EXPECT_EQ(msg._path, get_headers(out, "Path"));
+  EXPECT_EQ("P-Associated-URI: <sip:6505550231@homedomain>", get_headers(out, "P-Associated-URI"));
+  EXPECT_EQ("Service-Route: <sip:scscf.sprout.homedomain:5058;transport=TCP;lr;orig>", get_headers(out, "Service-Route"));
+  EXPECT_EQ(1,((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_REGISTRATION_STATS_TABLES.init_reg_tbl)->_attempts);
+  EXPECT_EQ(1,((SNMP::FakeSuccessFailCountTable*)SNMP::FAKE_REGISTRATION_STATS_TABLES.init_reg_tbl)->_successes);
```
AlexHockey: We have a problem in Clearwater that our tests are way over-specified - i.e. they test things they don't care about. All this test really cares about is that the Require header is present on a 200 OK. So I would be tempted to just check the status code and the requires header.

AlexHockey: The same goes for the other tests you've added.

metaswitch-ak10:
@AlexHockey Hi Alex, I have implemented the changes you suggested. Notable changes are:

- Added two helper routines in pjutils to make the main code more manageable and readable. The first, `is_param_in_generic_array_hdr`, checks whether a given parameter is present in a specific header of a message (I made it a bit more general to allow for any `pjsip_generic_array_hdr`). 
The second routine, `msg_get_first_routing_hdr_by_name`, finds the _first_ header (of type `pjsip_routing_hdr`) in a message.

- I have corrected my mistake regarding the checking of the "ob" parameter in the _first_ Path header. 

- Added a UT to test the case where the "ob" parameter is not present in the _first_ Path header, but it is present in a subsequent Path header.

- I have also made a correction to the code that checks for the presence of the "instance-id" and "reg-id" parameters in the Contact header. Prior to my correction, once found, the code continued checking for their presence, resulting in the booleans `contains_instance_id` and `contains_reg_id` being set to false.

src/pjutils.cpp
```diff
@@ -2827,6 +2827,42 @@ void PJUtils::add_top_header(pjsip_msg* msg, pjsip_hdr* hdr)
...
+  }
+  return found;
+}
+
+pjsip_routing_hdr* PJUtils::msg_get_first_routing_hdr_by_name(pjsip_msg* msg, const pj_str_t* name)
+{
+  pjsip_routing_hdr* h;
+  pjsip_routing_hdr* hdr;
+  h = hdr = (pjsip_routing_hdr*)pjsip_msg_find_hdr_by_name(msg,
+                                name, NULL);
```
AlexHockey: Why not put all these parameters on one line?

src/registrarsproutlet.cpp
```diff
@@ -729,22 +743,46 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
   }
 
-  // Deal with path header related fields in the response.
-  pjsip_routing_hdr* path_hdr = (pjsip_routing_hdr*)
-                              pjsip_msg_find_hdr_by_name(req, &STR_PATH, NULL);
-  if ((path_hdr != NULL) &&
-      (!aor_pair->get_current()->bindings().empty()))
+  // We check if the UE that sent this REGISTER supports "outbound" (RFC5626)
+  bool supported_outbound = PJUtils::is_param_in_generic_array_hdr(req,
+                          PJSIP_H_SUPPORTED, &STR_OUTBOUND);
```
AlexHockey: This alignment is odd. We try to put all parameters on the same line, or aligned in the same column. 

If you use vim, this vim setting will do that:

```vim
set cino=(0,g0,:0,N-s
```

src/registrarsproutlet.cpp
```diff
@@ -729,22 +743,46 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
   {
-    // We have bindings with path headers so we must require outbound.
-    pjsip_require_hdr* require_hdr = pjsip_require_hdr_create(get_pool(rsp));
-    require_hdr->count = 1;
-    require_hdr->values[0] = STR_OUTBOUND;
-    pjsip_msg_add_hdr(rsp, (pjsip_hdr*)require_hdr);
+    // Check for the presence of an "ob" parameter in the URI of the first path header
+    pjsip_sip_uri* uri = (first_path_hdr->name_addr.uri != NULL) ? 
+                       (pjsip_sip_uri*)pjsip_uri_get_uri(first_path_hdr->name_addr.uri) : NULL; 
+    bool contains_ob = ( (uri != NULL) &&
```
AlexHockey: We don't typically put spaces between brackets.

src/registrarsproutlet.cpp
```diff
@@ -729,22 +743,46 @@ void RegistrarSproutletTsx::process_register_request(pjsip_msg *req)
...
-    // We have bindings with path headers so we must require outbound.
-    pjsip_require_hdr* require_hdr = pjsip_require_hdr_create(get_pool(rsp));
-    require_hdr->count = 1;
-    require_hdr->values[0] = STR_OUTBOUND;
-    pjsip_msg_add_hdr(rsp, (pjsip_hdr*)require_hdr);
+    // Check for the presence of an "ob" parameter in the URI of the first path header
+    pjsip_sip_uri* uri = (first_path_hdr->name_addr.uri != NULL) ? 
+                       (pjsip_sip_uri*)pjsip_uri_get_uri(first_path_hdr->name_addr.uri) : NULL; 
+    bool contains_ob = ( (uri != NULL) && 
+                         (pjsip_param_find(&uri->other_param, &STR_OB) != NULL) );
```
AlexHockey: Ditto.

src/pjutils.cpp
```diff
@@ -2827,6 +2827,42 @@ void PJUtils::add_top_header(pjsip_msg* msg, pjsip_hdr* hdr)
...
+  if (h != NULL)
+  {
+    while ( (h=(pjsip_routing_hdr*)pjsip_msg_find_hdr_by_name(msg,
```
AlexHockey: Unnecessary space between the brackets.

src/pjutils.cpp
```diff
@@ -2827,6 +2827,42 @@ void PJUtils::add_top_header(pjsip_msg* msg, pjsip_hdr* hdr)
...
+  if (h != NULL)
+  {
+    while ( (h=(pjsip_routing_hdr*)pjsip_msg_find_hdr_by_name(msg, 
+                                  name, h->next)) != NULL )
+    {
+        hdr = h;
+    }
+  }
+  return hdr;
+}
```
AlexHockey: A few things in this function that I think are going to trip some people up. 

* The `x = y = z` construct is not one we use much. 
* Similarly, the `if ((x = y) != z)` is not often used. 

I would be tempted to rework this into a single for loop. It feels to me like this is less likely to be misread by a future developer.

```cpp
for (pjsip_routing_hdr* h = (pjsip_routing_hdr*)pjsip_msg_find_hdr_by_name(msg, name, NULL);
     h != NULL;
     h = (pjsip_routing_hdr*)pjsip_msg_find_hdr_by_name(msg, name, h->next))
{
  hdr = h;
}
```

include/pjutils.h
```diff
@@ -388,6 +388,24 @@ bool is_param_in_top_route(const pjsip_msg* req,
...
+/// header of a message.
+///
+/// @param msg       the message to be searched
+/// @param htype     type of pjsip_generic_array_hdr header to be serached
+/// @param pname     name of the parameter to be searched for
+///
+/// @return          boolean indicating whether the parameter was found (true if found)
+bool is_param_in_generic_array_hdr(pjsip_msg* msg, pjsip_hdr_e htype, const pj_str_t* param_name);
+
+/// Get the first (as added to the message) pjsip_routing_hdr header from a message
```
AlexHockey: I get what you're saying, but I think most people will read the function name and think that "first" means first in the message, rather than added first. 

Also, "added first" means something different depending on the header. Take for example, the History Info header (which you may not be familiar with). This is header where each proxy that a message traverses may add a new copy of the header. But for history info, new headers are added to the end of the message, not to the start! 

So I think the safest thing to do is just call this `msg_get_last_routing_hdr_by_name`, and just add a comment to the registrar explaining why we get the last header when the RFC talks about the first header.

AlexHockey:
I'm approving the PR because I don;t need to see it again after these markups (I would like you apply these markups though :-)

eleanor-merry:
Closing as merged in https://github.com/Metaswitch/sprout/pull/2163

PR #2104, opened 2018-01-24
[Reviewer: AJH] Added 'no served user' sas logs

metaswitch-ak10:
Hi Alex, could you please review this fix?

This PR adds a SAS log in the case when the _served user_ cannot be determined in flows between the SCSCF and AS. The SAS log itself is in the PR referenced bellow. This is a more detailed log than log 0x8100EB, giving the reason why the served user could not be determined: 

- either "Session case is terminating, but the request contains an overriding route header"

- or "URI is not locally hosted"

I have tested the fix by sending specific INVITE requests in order to to trigger both logs.

There are a couple of things I would like to check with you:

- I don't think we need to log also success cases, i.e. when served user can be determined, as these seem to be logged higher up (logs 0x8100DA - 0x8100DC).

- SL thinks the severity of the logs should be set to 80, since the logs are helpful in understanding the routing of the requests. Do you agree with this?


metaswitch-ak10:
Is this ok?

include/sproutsasevent.h
```diff
@@ -198,6 +198,8 @@ namespace SASEvent
   const int NO_CCFS_FOR_ACR = SPROUT_BASE + 0xF0;
 
   const int SCSCF_INVALID_URI = SPROUT_BASE + 0x0000F1;
+  const int NO_SERVED_USER_TERMINATING = SPROUT_BASE + 0x0000F2;
```
AlexHockey: It's not the fact that the session case is terminating that's important, it's that there's an overriding route header. How about NO_SERVED_USER_OVERRIDING_ROUTE?

metaswitch-ak10: That makes more sense, thanks

AlexHockey:
Looks good to me.

metaswitch-ak10:
Hi Alex,

I have made the SAS logs more detailed as you suggested.
I have also made the debug log "Classified URI as ..." more detailed by including the URI. I found that this saved me some time when testing my fix, as the log appears for multiple URIs in a single SIP message, so it helps to specify which one is being classified at a given moment.

AlexHockey:
:+1:

PR #2125, opened 2018-01-31
Handlers markup

Ying-metaswitch:
Hey Sathiyan thanks for reviewing the handlers. This PR is for the markups.

1. All the UT that shouldn't be deleted have been put back
2. `extract_impu` becomes a helper function;
    all comments in header file has been updated;
    `MockWithSM` inherits from `BaseTest` instead of `SipTest`;
    SM clients clear map of pointers in Bindings
3. Other markups as instructed

Memory leak in send_notifys and possible leak in SM UT, which don't seem to be mine.


src/ut/handlers_test.cpp
```diff
@@ -663,6 +699,24 @@ TEST_F(GetSubscriptionsTest, TwoSubscriptions)
   EXPECT_TRUE(document["subscriptions"].HasMember("789"));
 }
 
+// Test getting an IMPU when the local store is down.
+TEST_F(GetSubscriptionsTest, LocalStoreDown)
```
sathiyan-sivathas: This test doesn't talk to the local store anymore. Can you rename it so something about SM returning an error e.g. `SubscriberManagerFails.`

src/ut/handlers_test.h
```diff
@@ -29,7 +28,7 @@
 
 
 // Base class used for testing handlers with Mock SDMs.
```
sathiyan-sivathas: Can you update this comment.

src/handlers.cpp
```diff
@@ -207,38 +207,44 @@ HTTPCode DeregistrationTask::deregister_bindings(
...
     }
   }
 
-  Bindings unused_bindings;
-  return _cfg->_sm->remove_bindings(aor_id,
-                                    binding_ids,
-                                    SubscriberDataUtils::EventTrigger::ADMIN,
-                                    unused_bindings,
-                                    trail());
+  bindings.clear();
```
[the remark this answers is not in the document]
Ying-metaswitch: Very detailed! Thanks :)

Ying-metaswitch:
Yeah now that Binding is freed in the production code correctly, I changed UT so that they don't delete the binding. full_test passed and passing back for review

src/handlers.cpp
```diff
@@ -207,38 +208,45 @@ HTTPCode DeregistrationTask::deregister_bindings(
...
-                                    binding_ids,
-                                    SubscriberDataUtils::EventTrigger::HSS,
-                                    unused_bindings,
-                                    trail());
+  SubscriberDataUtils::delete_bindings(bindings);
+
+  // Remove these bindings via subscriber manager
+  rc = _cfg->_sm->remove_bindings(aor_id,
+                                  binding_ids,
+                                  SubscriberDataUtils::EventTrigger::ADMIN,
```
sathiyan-sivathas: I think you've changed this back from `HSS` to `ADMIN` when doing a catchup merge maybe? This needs to be `HSS` since the DeregistrationTask is an operation triggered by the HSS.

This is how we get SM makes sure it doesn't trigger an SAR when removing bindings because of an RTR.

src/ut/handlers_test.cpp
```diff
@@ -107,15 +107,17 @@ class DeregistrationTaskTest : public SipTest
...
+                         const Bindings& bindings,
                          std::vector<std::string>& binding_ids)
   {
     EXPECT_CALL(*_subscriber_manager, get_bindings(aor_id, _, _))
       .WillOnce(DoAll(SetArgReferee<1>(bindings), Return(HTTP_OK)));
 
     EXPECT_CALL(*_subscriber_manager,
-                remove_bindings(aor_id, _, SubscriberDataUtils::EventTrigger::HSS, _, _))
-          .WillOnce(DoAll(SaveArg<1>(&binding_ids), Return(HTTP_OK)));
+                remove_bindings(aor_id, _, SubscriberDataUtils::EventTrigger::ADMIN, _, _))
```
sathiyan-sivathas: Same as above, this needs to be `HSS`.

src/ut/handlers_test.cpp
```diff
@@ -229,18 +231,18 @@ TEST_F(DeregistrationTaskTest, SubscriberManagerWritesFail)
...
-  bindings[""] = &binding;
+  bindings[""] = binding;
   EXPECT_CALL(*_subscriber_manager,
               get_bindings(aor_id, _, _))
     .WillOnce(DoAll(SetArgReferee<1>(bindings),
                     Return(HTTP_OK)));
 
   EXPECT_CALL(*_subscriber_manager,
-              remove_bindings(aor_id, _, SubscriberDataUtils::EventTrigger::HSS, _, _))
+              remove_bindings(aor_id, _, SubscriberDataUtils::EventTrigger::ADMIN, _, _))
```
sathiyan-sivathas: Same again.

PR #2149, opened 2018-02-13
Add comment about aliases to code

KristaHyer:
We talked about aliases last week, and you mentioned that we should add a TODO to the code.
Is this the sort of thing you were looking for?

The only file which needs reviewing is src/sprout_xml_utils.cpp (I'm just storing the rest of my stuff here so I don't lose it).


KristaHyer:
@AlexHockey had to push this branch in - can you please review the file `src/sprout_xml_utils.cpp` even though this request is now closed? Thanks!

PR #2151, opened 2018-02-13
Increased sprout error logs from EANTC work

BennettAllen1:
This is the apple scrumble branch but with the error codes from the two commits reverted out


BennettAllen1:
added more logging (including error ones), also upgraded some logging levels

BennettAllen1:
similar to #2138 but with the reverts

BennettAllen1:
made changes to this based on https://metacom2.metaswitch.com/confluence/pages/viewpage.action?spaceKey=MobileEngineering&title=Clearwater+Engineering+Logging+Guidelines

BennettAllen1:
Still need to update cpp-common after it is merged

PR #2164, opened 2018-02-20
Change Subscriptions from a typedef to a class

KristaHyer:
Need to deal with:
- [ ] still getting errors
- [ ] worried that sometimes we assume there is only one subscription in the map (check we only do this in the one test, which comments this, and nowhere else)
- [ ] need to update comments! (for example in doxygen comments in headers where talking about how memory is passed around)
- [ ] tidy by creating new typedef
- [x] put delete expired into own funtion
- [x] find all `inserts` on subscription class - possible memory issue with them (see commit for fix)


KristaHyer:
Moved to git.datcon

PR #2165, opened 2018-02-20
[Reviewer: Seb] Route off-net calls to the BGCF when not doing ENUM

richardwhiuk:
If we aren't doing ENUM, and we have a URI which is already known to be
off-net, don't bother routing it to the I-CSCF which will need to perform an
LIR.

### DIFF
--- a/include/scscfsproutlet.h
+++ b/include/scscfsproutlet.h
@@ -349,7 +349,10 @@ private:
   void route_to_icscf(pjsip_msg* req);
 
   /// Route the request to the BGCF.
-  void route_to_bgcf(pjsip_msg* req);
+  ///
+  /// @param req      - The request to route
+  /// @param reason   - The SAS Event ID to log as the reason
+  void route_to_bgcf(pjsip_msg* req, int reason);
 
   /// Route the request to the terminating side S-CSCF.
   void route_to_term_scscf(pjsip_msg* req);
@@ -605,8 +608,7 @@ private:
   /// 400 Bad Request error (which also frees the request).
   ///
   /// @param req      - The request to reject
-  /// @param uri_str  - The URI string to add to the SAS log
-  void reject_invalid_uri(pjsip_msg* req, const std::string& uri_str);
+  void reject_invalid_uri(pjsip_msg* req);
 
   /// The S-CSCF URI for this transaction. This is used in the SAR sent to the
   /// HSS. This field should not be changed once it has been set by the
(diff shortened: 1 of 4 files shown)


include/scscfsproutlet.h
```diff
@@ -605,8 +608,7 @@ class SCSCFSproutletTsx : public CompositeSproutletTsx
   /// 400 Bad Request error (which also frees the request).
   ///
   /// @param req      - The request to reject
-  /// @param uri_str  - The URI string to add to the SAS log
-  void reject_invalid_uri(pjsip_msg* req, const std::string& uri_str);
+  void reject_invalid_uri(pjsip_msg* req);
```
sebrexmetaswitch: See below comment, but I don't think this is right

src/ut/scscf_test.cpp
```diff
@@ -2267,11 +2267,37 @@ TEST_F(SCSCFTest, TestWithoutEnum)
...
+  msg._todomain = "ut.cw-ngv.com";
+
+  // IFCs for originating call
+  HSSConnection::irs_info irs_info_1;
+  setup_irs_info(irs_info_1, "+16505551000", "homedomain");
+  expect_get_subscriber_state(irs_info_1, "sip:+16505551000@homedomain");
+
+  // We only do ENUM on originating calls
+  msg._route = "Route: <sip:sprout.homedomain;orig>";
+  msg._extra = "Record-Route: <sip:homedomain>\nP-Asserted-Identity: <sip:+16505551000@homedomain>";
```
sebrexmetaswitch: Do we need the Record-Route header here? If so, can we add a comment explaining why?

src/scscfsproutlet.cpp
```diff
@@ -2369,8 +2370,10 @@ pjsip_msg* SCSCFSproutletTsx::get_base_request()
   }
 }
 
-void SCSCFSproutletTsx::reject_invalid_uri(pjsip_msg* req, const std::string& uri_str)
+void SCSCFSproutletTsx::reject_invalid_uri(pjsip_msg* req)
 {
+  std::string uri_str = PJUtils::uri_to_string(PJSIP_URI_IN_REQ_URI, req->line.req.uri);
```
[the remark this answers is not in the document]
sebrexmetaswitch: I think?

richardwhiuk:
I've tweaked the invocation of `reject_invalid_uri()` to allow it to take a URI string if you don't want the request URI.

PR #2166, opened 2018-02-23
[Reviewer: Seb] HttpClient rework

johadalin:
These are the changes to Sprout to take the new HttpClient/HttpRequest changes. As part of plumbing this through, we've updated:

* RalfProcessor
* HssConnection
* XDMConnection
* Constructors for the RalfProcessor and ChronosConnection in main.cpp

In the UTs:

* Created a MockXDMConnection rather than the old fake, which called through to a fake http connection.
* Fixed/updated mmtel/xdm tests in scscf_test
* Updated RalfProcessor tests

Changes currently pass `make full_test`, and build into a debian. Pipeline is currently running, so we'll check in on that soon.


src/ut/sprout_fv_test.cpp
```diff
@@ -299,7 +299,7 @@ class SproutFVTest : public SipTest
   NotifySender* _notify_sender;
   AnalyticsLogger* _analytics;
   FakeHSSConnection* _hss_connection;
-  FakeXDMConnection* _xdm_connection;
+  MockXDMConnection* _xdm_connection;
```
sebrexmetaswitch: This seems odd - we create a `MockXDMConnection` but never `EXPECT` on it... Does that meant that we can just get rid of it?

src/ut/scscf_test.cpp
```diff
@@ -6752,19 +6761,14 @@ TEST_F(SCSCFTest, MmtelThenExternal)
...
-                            <communication-diversion active="false"/>
-                            <incoming-communication-barring active="false"/>
-                            <outgoing-communication-barring active="false"/>
-                          </simservs>)");  // "
+  EXPECT_CALL(*_xdm_connection, get_simservs("sip:6505551000@homedomain",_,_,_))
+    .WillOnce(DoAll(SetArgReferee<1>(simservs),
+                    Return(true)));
+  EXPECT_CALL(*_xdm_connection, get_simservs("sip:6505551234@homedomain",_,_,_))
+    .WillOnce(DoAll(SetArgReferee<1>(simservs),
+                    Return(true)));
```
sebrexmetaswitch: It's not clear to me why we're returning the string `simservs` here, I think because it's not clear to me what this test is actually ttrying to do. If you happen to know, it would be good if we could have some comments like:
```
// Expect that the request for XDM (whatever that is) for user 6505551000
// returns simservs (whatever they are) with active barring "false" (or whatever)
...
// Then expect that the simservs are requested for user blah, which also returns
// blah...
```

As it's completely opaque to me *why* we expect these calls

src/ut/scscf_test.cpp
```diff
@@ -6337,7 +6336,11 @@ TEST_F(SCSCFTest, MmtelCdiv)
                               </communication-diversion>
                             <incoming-communication-barring active="false"/>
                             <outgoing-communication-barring active="false"/>
-                          </simservs>)");  // "
+                          </simservs>)";
+  EXPECT_CALL(*_xdm_connection, get_simservs("sip:6505551234@homedomain",_,_,_))
```
sebrexmetaswitch: Similar to my other comment, if you know *why* we expect this, it would be good to have a comment

src/ut/scscf_test.cpp
```diff
@@ -6457,45 +6460,49 @@ TEST_F(SCSCFTest, MmtelDoubleCdiv)
...
+                                   <cp:rule id="rule1">
+                                     <cp:conditions/>
+                                     <cp:actions><forward-to><target>sip:6505559012@homedomain</target></forward-to></cp:actions>
+                                   </cp:rule>
+                                 </cp:ruleset>
+                               </communication-diversion>
+                             <incoming-communication-barring active="false"/>
+                             <outgoing-communication-barring active="false"/>
+                           </simservs>)";
+  EXPECT_CALL(*_xdm_connection, get_simservs("sip:6505555678@homedomain",_,_,_))
```
sebrexmetaswitch: (and again, I can't see why we expect this not what we're really returning)

src/ut/scscf_test.cpp
```diff
@@ -6917,19 +6920,14 @@ TEST_F(SCSCFTest, MultipleMmtelFlow)
...
-                              <default-behaviour>presentation-restricted</default-behaviour>
-                            </originating-identity-presentation-restriction>
-                            <communication-diversion active="false"/>
-                            <incoming-communication-barring active="false"/>
-                            <outgoing-communication-barring active="false"/>
-                          </simservs>)");  // "
+  EXPECT_CALL(*_xdm_connection, get_simservs("sip:6505551000@homedomain",_,_,_))
+    .Times(2).WillRepeatedly(DoAll(SetArgReferee<1>(simservs),
+                                   Return(true)));
+  EXPECT_CALL(*_xdm_connection, get_simservs("sip:6505551234@homedomain",_,_,_))
```
sebrexmetaswitch: Again with these two

src/ut/scscf_test.cpp
```diff
@@ -6917,19 +6920,14 @@ TEST_F(SCSCFTest, MultipleMmtelFlow)
...
+    .Times(2).WillRepeatedly(DoAll(SetArgReferee<1>(simservs),
```
sebrexmetaswitch: Why twice? (I assume there's a reason, but it's not obviuos what it is)

src/ut/scscf_test.cpp
```diff
@@ -10300,15 +10298,13 @@ class SCSCFTestWithRalf : public SCSCFTestBase
   static void SetUpTestCase()
   {
     SCSCFTestBase::SetUpTestCase();
-    _ralf_connection = new NiceMock<MockHttpConnection>();
-    _ralf_processor = new NiceMock<MockRalfProcessor>(_ralf_connection);
+    _ralf_processor = new NiceMock<MockRalfProcessor>();
```
sebrexmetaswitch: Hmm, I see we had a `NiceMock` before, but I'm not sure why. Should all the `...WithRalf` tests be expecting explicitly on the `_ralf_processor`? If anything, I'd have thought we want a `StrictMock` here

PR #2167, opened 2018-02-23
[Reviewer RJW2] Ensure PJSIP timers that are cancelled after they've popped do not make it into the sproutlets

steelyvoid:
Hi Richard

Are you OK to review this two-pronged timer cancel fix?  

No UTs, but I managed to test first the timer handling code in scscfsproutlet with this "slow DNS" repro scenario (which stopped the crash, but a random leg got CANCELled and the call failed) and then the "cancel timer even if popped" code in sproutletproxy with the same slow repro (this time the call completed fine, with the slow timer being properly cancelled) altogether.

I haven't been able to come up with a repro scenario for the "extra response" case in scscfsproutlet, but I'm going to claim that the backstop code is good by inspection (and "make full_test" passes).

I'll reference the associated SAS resource bundle fix shortly

### DIFF
--- a/include/sproutletproxy.h
+++ b/include/sproutletproxy.h
@@ -302,7 +302,8 @@ protected:
     virtual void on_tsx_state(pjsip_event* event);
 
     // A count of the number of pending Callbacks that are queued for this
-    // UASTsx. A non-zero count prevents the UASTsx from being destroyed
+    // UASTsx plus the number of outstanding PJSIP timers.
+    // A non-zero count prevents the UASTsx from being destroyed
     int _pending_callbacks = 0;
 
   private:
--- a/include/sproutletproxy.h
+++ b/include/sproutletproxy.h
@@ -302,7 +302,10 @@ protected:
     virtual void on_tsx_state(pjsip_event* event);
 
     // A count of the number of pending Callbacks that are queued for this
-    // UASTsx plus the number of outstanding PJSIP timers.
+    // UASTsx plus the number of outstanding timer callbacks, where this
+    // includes timer callbacks outstanding in PJSIP, and timer callbacks
+    // which have been popped by PJSIP, but whose callback is awaiting 
+    // scheduling on a worker thread.
     // A non-zero count prevents the UASTsx from being destroyed
     int _pending_callbacks = 0;
 
--- a/src/sproutletproxy.cpp
+++ b/src/sproutletproxy.cpp
@@ -672,6 +672,7 @@ SproutletProxy::UASTsx::Callback::Callback(UASTsx* tsx, std::function<void()> ru
   // Whenever we create a Callback we expect it to be run, so increase the count
   // of _pending_callbacks on the UASTsx
   _tsx->_pending_callbacks++;
+  TRC_DEBUG("Increment pending callbacks to: %d", _tsx->_pending_callbacks);
 }
 
 void SproutletProxy::UASTsx::Callback::run()
@@ -681,6 +682,7 @@ void SproutletProxy::UASTsx::Callback::run()
   // Now that we are running a Callback, we can decrement the count of pending
   // callbacks
   _tsx->_pending_callbacks--;
+  TRC_DEBUG("Decrement pending callbacks to: %d", _tsx->_pending_callbacks);
 
   // _run_fn() contains that actual work that we need to do for this Callback
   _run_fn();
@@ -1157,6 +1159,12 @@ bool SproutletProxy::UASTsx::schedule_timer(SproutletWrapper* tsx,
   if (scheduled)
   {
     _pending_timers.insert(tentry);
+
+    // We don't know when this timer is going to pop, so increment the
+    // _pending_callbacks count now and only decrement it when the timer
+    // pops or when we know we've cancelled it successfully.
+    _pending_callbacks++;
+    TRC_DEBUG("Increment pending callbacks to: %d", _pending_callbacks);
   }
   return scheduled;
 }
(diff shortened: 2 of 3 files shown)


steelyvoid:
Good spot - thanks!

steelyvoid:
Hi Richard

Can you give another look at the latest commit here to see whether it addresses your concern above.  I have run a "make full_test" so I'm confident that this doesn't leave any orphaned TSX objects.

One suggested tweak: I'm nervous that, thanks to a future bug in timers (or callbacks) logic, we might end up with _pending_callbacks going negative.  Should https://github.com/Metaswitch/sprout/blob/d2dca9b46ffe108b076bf5825ad2b06815d7df98/src/sproutletproxy.cpp#L1357 actually check whether the count is <= 0, and then make a TRC_ERROR in the branch if its negative?

steelyvoid:
As discussed on the phone, a solution which ties the pending callback increment to the lifetime of the TimerCallback necessitates use of atomic integers, as the increment needs to be done outside the transaction lock, so is arguably equally ugly.

I've left the existing fix in place, but beefed up the comments to cover the reasoning for the pending_callback increment/decrement and added a TRC_ERROR (with trail ID) to catch the "yikes - pending_callback went negative" case.  Are you OK to cast a (hopefully) final eye over it?

include/sproutletproxy.h
```diff
@@ -302,7 +302,8 @@ class SproutletProxy : public BasicProxy, SproutletHelper
     virtual void on_tsx_state(pjsip_event* event);
 
     // A count of the number of pending Callbacks that are queued for this
-    // UASTsx. A non-zero count prevents the UASTsx from being destroyed
+    // UASTsx plus the number of outstanding PJSIP timers.
```
richardwhiuk: Hmm, I think this is inaccurate, and I think the distinction is important. I think we should say 'the number of outstanding timer callbacks, where this includes timer callbacks outstanding in PJSIP, and timer callbacks which have been popped by PJSIP, but whose callback is awaiting scheduling on a worker thread'.

src/sproutletproxy.cpp
```diff
@@ -672,6 +672,7 @@ SproutletProxy::UASTsx::Callback::Callback(UASTsx* tsx, std::function<void()> ru
   // Whenever we create a Callback we expect it to be run, so increase the count
   // of _pending_callbacks on the UASTsx
   _tsx->_pending_callbacks++;
+  TRC_DEBUG("Increment pending callbacks to: %d", _tsx->_pending_callbacks);
```
richardwhiuk: `Incremented` I think, given we've already done the increment. We don't want to make the log reader believe the number is about to be increased.

src/sproutletproxy.cpp
```diff
@@ -681,6 +682,7 @@ void SproutletProxy::UASTsx::Callback::run()
   // Now that we are running a Callback, we can decrement the count of pending
   // callbacks
   _tsx->_pending_callbacks--;
+  TRC_DEBUG("Decrement pending callbacks to: %d", _tsx->_pending_callbacks);
```
richardwhiuk: Ditto, but decremented.

src/sproutletproxy.cpp
```diff
@@ -1157,18 +1159,51 @@ bool SproutletProxy::UASTsx::schedule_timer(SproutletWrapper* tsx,
...
+    // would have meant making _pending_callbacks an atomic integer because
+    // it would have needed to be incremented outside the transaction lock.
+    // Swings and roundabouts.
+    //
+    // We need _pending_callbacks to take account of timers in flight, because
+    // otherwise the Tsx object might be destroyed after the timer has popped
+    // but before its had a chance to run, in which case we'd end up
+    // threading a destroyed object.
+    _pending_callbacks++;
+    TRC_DEBUG("Increment pending callbacks to: %d", _pending_callbacks);
```
richardwhiuk: Ditto.

src/sproutletproxy.cpp
```diff
@@ -1157,18 +1159,51 @@ bool SproutletProxy::UASTsx::schedule_timer(SproutletWrapper* tsx,
...
+  // Cancel the timer at PJSIP
+  if (_sproutlet_proxy->cancel_timer(tentry))
   {
-    _pending_timers.erase(tentry);
+    // Successfully cancelled.  Decrement the pending callbacks count
+    // incremented in SproutletProxy::UASTsx::schedule_timer.  Note that
+    // if the cancel failed, it will be because the timer has popped, in which
+    // case _pending_callbacks will be decremented by process_timer_pop below
+    _pending_callbacks--;
+    TRC_DEBUG("Decrement pending callbacks to: %d", _pending_callbacks);
```
richardwhiuk: Ditto. I wonder if we should have a `->decrement_pending_callbacks()` function instead... - although we'd lose line information. Probably fine as is.

src/sproutletproxy.cpp
```diff
@@ -1197,6 +1232,14 @@ void SproutletProxy::UASTsx::process_timer_pop(pj_timer_entry* tentry)
 {
   enter_context();
 
+  // Timer callback has been made so decrement the pending callback count.
+  // Note that we need to decrement _pending_callbacks both in this routine
+  // and in SproutletProxy::UASTsx::cancel_timer, the latter IFF the PJSIP
+  // cancel call did not fail (if it failed, then the timer will have popped,
+  // in which case this routine will eventually get called).
+  _pending_callbacks--;
+  TRC_DEBUG("Decrement pending callbacks to: %d", _pending_callbacks);
```
richardwhiuk: Ditto.

include/sproutletproxy.h
steelyvoid: Agreed - thanks

src/sproutletproxy.cpp
steelyvoid: Will do

PR #2170, opened 2018-03-05
[Reviewer Krista] Correctly quote PCV header contents

steelyvoid:
Hi Krista

Can you review this fix for me this week (before the code slush)? Its a fix to Sprint-critical issue https://github.com/Metaswitch/clearwater-issues/issues/3646.

Fix here is to refactor the PCFA print code (which has the same quoting requirement) and use this code to correctly quote the parameters for the PCV, with a couple of exceptions

-  I've left the icid-value as "always quote", for two reasons:  Its what the code explicitly did already (possibly for interworking reasons) and we want to cope with the "icid-value is zero-length" case (icid-value is mandatory in the PCV, and I expect 'icid-value=;....' to be rejected by the downstream node as unparseable, whereas 'icid-value="";...' is less likely to cause problems)
-  the icid-generated-at parameter only takes "host" parameters, so I'm just reported what is already in the parameter unquoted.

Tested using the updated UT test script and also live (I fired in a test INVITE with the PCV header from the issue and checked that the header in the response was correctly quoted)

### DIFF
--- a/src/custom_headers.cpp
+++ b/src/custom_headers.cpp
@@ -20,6 +20,7 @@ extern "C" {
 #include "log.h"
 #include "constants.h"
 #include "custom_headers.h"
+#include "pjutils.h"
 
 /// Custom parser for Privacy header. This is registered with PJSIP below
 /// in register_custom_headers().
@@ -853,11 +854,12 @@ void *pjsip_p_c_v_hdr_shallow_clone(pj_pool_t* pool, const void* o)
 
 int pjsip_p_c_v_hdr_print_on(void* h, char* buf, pj_size_t len)
 {
-  const pjsip_parser_const_t *pc = pjsip_parser_const();
   pjsip_p_c_v_hdr* hdr = (pjsip_p_c_v_hdr*)h;
   char* p = buf;
 
   // Check the fixed parts of the header will fit.
+  // We always need an icid-value for the header to be valid (even if icid.slen
+  // is 0), so we always quote this to give us a guaranteed valid parameter value.
   int needed = 0;
   needed += hdr->name.slen; // Header name
   needed += 2;              // : and space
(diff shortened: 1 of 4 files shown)


steelyvoid:
Incidentally, I mentioned in the issue that according to RFC 7976, we shouldn't putting PCVs in 100 responses anyway.  AK10 is fixing this on a separate thread (we need to address the quoting issue for the other headers anyway)

src/custom_headers.cpp
```diff
@@ -853,38 +854,64 @@ void *pjsip_p_c_v_hdr_shallow_clone(pj_pool_t* pool, const void* o)
...
+      needed += 2;            // quotes
+    }
     needed += 1;              // ;
   }
+
+  // Quoting is not an option for the icid-gen-addr parameter
   if (hdr->icid_gen_addr.slen) {
     needed += 18;              // icid-generated-at=
     needed += hdr->icid_gen_addr.slen; // <icid-generated-at>
+    needed += 1;               // ;
```
KristaHyer: Why do we need this extra character?

src/pjutils.cpp
```diff
@@ -2288,37 +2288,46 @@ void PJUtils::add_pcfa_header(pjsip_msg* msg,
...
-
   // Check whether we need to quote the value.  We'll need to do this if
   // - its not already quoted
   // - it contains characters other than those allowed for a host or
   //   token (see RFC 3455, section 5.5)
   // Note that we assume for simplicity that if the value starts with '[',
   // its an ipv6 address (int_parse_host in sip_parser.c makes the same
   // assumption and the pjsip_HOST_SPEC doesn't cover IPv6 parsing).
-  const char *inbuf = value.c_str();
   bool quote = false;
```
KristaHyer: Can we rename this to `needs_quoting`? I think it's clearer

src/ut/sip_parser_test.cpp
```diff
@@ -54,7 +54,7 @@ TEST_F(SipParserTest, PChargingVector)
              "Max-Forwards: 63\n"
```
KristaHyer: Can you add a comment to this test stating what it tests?

src/custom_headers.cpp
steelyvoid: I think it was an omission in the original code.  You need to leave space for the semi-colon separator that needs to be appended if there are more parameters in the list after this one.  The previous parameters all made space for it, but not this one, so you'd make space for the ';' if there was a term_ioi but no icid-generated-at (say), but not if there wasn't a term_ioi but there was a icid-generated-at.

Its possible that we are now allowing for a final ';', which is not strictly necessary, but at least we're doing it consistently regardless of which parameters happen to be present.

src/pjutils.cpp
steelyvoid: OK

src/custom_headers.cpp
KristaHyer: This makes sense - thanks!

PR #2171, opened 2018-03-06
[Reviewer: AJH] Integrate Sprout with the new sas.json file

jack-atack:
This removes the reading/passing of the `sas_server` shared_config parameter from Sprout, instead using the new SasService module (associated PR: https://github.com/Metaswitch/cpp-common/pull/770)


src/main.cpp
```diff
@@ -787,14 +786,11 @@ static pj_status_t init_options(int argc, char* argv[], struct options* options)
         std::vector<std::string> sas_options;
```
AlexHockey: It doesn't feel worth doing the split on comma, or the check on the number of elements.

src/main.cpp
```diff
@@ -2069,9 +2059,12 @@ int main(int argc, char* argv[])
   quiescing_mgr = new QuiescingManager();
   quiescing_mgr->register_completion_handler(new QuiesceCompleteHandler());
 
+  // Initialise the SasService, to read the SAS config to pass into SAS::Init
+  SasService* sas_service = new SasService();
+
   // Initialize the PJSIP stack and associated subsystems.
   status = init_stack(opt.sas_system_name,
-                      opt.sas_server,
+                      sas_service->get_single_sas_server(),
```
AlexHockey: So I think what will happen here is:

* The `SasService` starts up and refreshes the SAS library's config. 
* The SAS library gets initialized with a single IP address. 

I don't know how the SAS libraries are implemented but I'm not convinced that will work properly, or continue to work properly in the future given how un-intuitive it is. 

I think there are two reasonable designs here:

* The SAS service takes over responsibility for configuring the client library, and sprout (homestead, ralf, etc.) never call `SAS::init` or `SAS::term` (dumb client library, smart SasService).
* The SAS library copes with being passed JSON configuration, and only connecting to the first address if appropriate (smart client library, dumb SasService). 

At the moment it feels like we have a halfway house where both the client library and the SasService are dumb, which either means that sprout has to pick up the slack, or it ends up doing dumb things (like reloading the config before the library is initialized).

PR #2173, opened 2018-03-07
[Reviewer: RJW] Added option for PCFA header to be included on provisional responses

metaswitch-ak10:
Hi Richard,

Here is the PR for the PCFA issue we have discussed earlier. The fix is staright forward and includes the following:
* Added shared_config option "include_pcfa_on_provisional_responses". When set to 'Y', S-CSCF adds the PCFA header to all 1xx (except 100) responses.
* Removed PCFA and PCV headers from 100 responses

**Note that even with sprout's behaviour prior to my fix, the PCFA header is always included on final responses (>=200).**

I have live tested my fix on our l3-vt test deployment. I have also added some UTs.

### DIFF
--- a/src/scscfsproutlet.cpp
+++ b/src/scscfsproutlet.cpp
@@ -744,11 +744,12 @@ void SCSCFSproutletTsx::on_rx_response(pjsip_msg* rsp, int fork_id)
     }
   }
 
-  if ((st_code >= PJSIP_SC_OK) && (_hss_data_cached))
+  if ((st_code > PJSIP_SC_TRYING) && _hss_data_cached)
   {
-    // Final response. Add a P-Charging-Function-Addresses header if one is
-    // not already present for some reason. We only do this if we have
-    // the charging addresses cached (which we should do).
+    // We add a P-Charging-Function-Addresses header if one is not already present.
+    //  We only do this if we have the charging addresses cached (which we should do).
+    // According to TS 24.229, PCFAs should not be added to responses being sent to ASes,
+    // except in some very specific cases, but this is unlikely to cause inter-op issues.
     PJUtils::add_pcfa_header(rsp,
                              get_pool(rsp),
                              _irs_info._ccfs,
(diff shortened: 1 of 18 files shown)


metaswitch-ak10:
Hi Richard, just to let you know, I still need to touch up the UTs since currently they don't pass.

metaswitch-ak10:
@richardwhiuk Hi Richard. It looks like we will be going forward with the current state of this fix as we have agreed to get it done by the 16th. I am going to cut an MR  for Sprint and if we later find that Sprint want a different behaviour (in regards to my previous email about the originating user's CCFs not being sent to the first originating AS), we will cut another MR. Could you please review this current fix?

src/scscfsproutlet.cpp
```diff
@@ -744,11 +744,12 @@ void SCSCFSproutletTsx::on_rx_response(pjsip_msg* rsp, int fork_id)
...
   }
 
-  if ((st_code >= PJSIP_SC_OK) && (_hss_data_cached))
+  if ((st_code > PJSIP_SC_TRYING) && _hss_data_cached)
   {
-    // Final response. Add a P-Charging-Function-Addresses header if one is
-    // not already present for some reason. We only do this if we have
-    // the charging addresses cached (which we should do).
+    // We add a P-Charging-Function-Addresses header if one is not already present.
+    //  We only do this if we have the charging addresses cached (which we should do).
```
richardwhiuk: Random space at the start of this line.

src/scscfsproutlet.cpp
```diff
@@ -744,11 +744,12 @@ void SCSCFSproutletTsx::on_rx_response(pjsip_msg* rsp, int fork_id)
     }
   }
 
-  if ((st_code >= PJSIP_SC_OK) && (_hss_data_cached))
+  if ((st_code > PJSIP_SC_TRYING) && _hss_data_cached)
   {
-    // Final response. Add a P-Charging-Function-Addresses header if one is
-    // not already present for some reason. We only do this if we have
-    // the charging addresses cached (which we should do).
+    // We add a P-Charging-Function-Addresses header if one is not already present.
```
richardwhiuk: I'd make it clear that we want to include a PCFA on provisional 1xx responses and final responses, and that we don't want to include it on 100 Trying responses.

PR #2176, opened 2018-03-09
[Reviewer: EM] Don't include NAPTR queries in latency calculation

sathiyan-sivathas:


sathiyan-sivathas:
Ellie says "LGTM" on slack.

PR #2178, opened 2018-03-13
[Reviewer: RND] Allow searching for SIP errors in SAS 

rkd-msw:
Turns out you were right, it wasn't that hard! Tested by building and deploying https://github.com/Metaswitch/sprout/commit/93c2babe56687bd4aea4bc7f2c49abd409740b6d, sending a REGISTER, and checking it came up in SAS when I searched for a SIP 503 but not when I searched for a SIP 423.

### DIFF
--- a/src/common_sip_processing.cpp
+++ b/src/common_sip_processing.cpp
@@ -277,8 +277,9 @@ static void sas_log_tx_msg(pjsip_tx_data *tdata)
   }
   else if (trail != 0)
   {
-    // Raise SAS markers on initial requests only - responses in the same
-    // transaction will have the same trail ID so don't need additional markers
+    // Raise SAS Call-ID, branch ID, To and From markers on initial requests
+    // only - responses in the same transaction will have the same trail ID so
+    // don't need additional markers.
     if (tdata->msg->type == PJSIP_REQUEST_MSG)
     {
       PJUtils::report_sas_to_from_markers(trail, tdata->msg);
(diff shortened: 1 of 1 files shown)


src/common_sip_processing.cpp
```diff
@@ -285,6 +285,18 @@ static void sas_log_tx_msg(pjsip_tx_data *tdata)
...
 
+    // Log a SIP protocol error marker for 4xx, 5xx or 6xx responses, so we can
+    // search by error code.
+    if (tdata->msg->type == PJSIP_RESPONSE_MSG && (tdata->msg->line.status.code > 399))
+    {
+      SAS::Marker error_marker(trail, MARKER_ID_PROTOCOL_ERROR, 1u);
+      error_marker.add_static_param(5); // SIP protocol
+      error_marker.add_static_param(tdata->msg->line.status.code);
+      // We can't give a more specific error string here - it would be a waste of bandwidth.
+      error_marker.add_var_param("");
```
RobDover: Why do we need to add a parameter here at all? Is it so that this marker is formatted in the same way as it is on other products?

And if this is used on other products, do you know why the parameter is there and (presumably) in use for them? It doesn't seem particularly useful to report if we're just searching on it...

src/common_sip_processing.cpp
```diff
@@ -285,6 +285,18 @@ static void sas_log_tx_msg(pjsip_tx_data *tdata)
       PJUtils::mark_sas_call_branch_ids(trail, tdata->msg);
     }
 
+    // Log a SIP protocol error marker for 4xx, 5xx or 6xx responses, so we can
```
RobDover: The comment on [line 280](https://github.com/Metaswitch/sprout/pull/2178/files#diff-287bb104cc53d2730d4870cfea31701aR280) is no longer accurate - can you update it?

src/common_sip_processing.cpp
```diff
@@ -285,6 +285,18 @@ static void sas_log_tx_msg(pjsip_tx_data *tdata)
...
+    // search by error code.
+    if (tdata->msg->type == PJSIP_RESPONSE_MSG && (tdata->msg->line.status.code > 399))
```
RobDover: This feels icky to me - would rather have a constant defined somewhere than a naked `399`. How about `tdata->msg->line.status.code >= PJSIP_SC_BAD_REQUEST`?

src/common_sip_processing.cpp
rkd-msw: The protocol error marker is defined as taking two static parameters (protocol and protocol-specific error code) and a variable parameter (error string) - and I'd like to raise it correctly. If we raised this marker throughout the code, we could be more specific, with errors like "Data missing from memcached" - and I suspect this is the text that shows up in SAS search results on the right.

PR #2180, opened 2018-03-15
180s received after a transaction has been terminated are mis-routed

kenferguson:
Can you review please? Feel free to be pretty picky - I've not worked in C++ or Sprout before so it's worth pointing out things even if they're just minor. 

I've written the UT and run make test. I'll run make deb before committing. I've talked to Ellie - I won't be merging into 11.2.


include/basicproxy.h
```diff
@@ -46,6 +46,10 @@ class BasicProxy
 
 protected:
 
+  /// Route a response received outside a transaction.
+  pj_bool_t route_rx_response(pjsip_tx_data *tdata);
+
+
```
richardwhiuk: Extra empty blank line.

include/sproutletproxy.h
```diff
@@ -79,6 +79,9 @@ class SproutletProxy : public BasicProxy, SproutletHelper
                               const pjsip_sip_uri* base_uri,
                               pj_pool_t* pool) const;
 
+  /// Handles responses that are received outside a transaction.
+  pj_bool_t on_rx_response(pjsip_rx_data *rdata);
```
richardwhiuk: This is overriding a method from basic proxy, so you should add ` override` prior to the `;` which will make the compiler check this is happening. This will cause a compile error if not true (e.g. if you someone changes the base signature).

include/sproutletproxy.h
```diff
@@ -286,6 +289,7 @@ class SproutletProxy : public BasicProxy, SproutletHelper
     /// Handle a timer pop.
     static void on_timer_pop(pj_timer_heap_t* th, pj_timer_entry* tentry);
 
+
```
richardwhiuk: Not sure why you've added this blank line.

src/basicproxy.cpp
```diff
@@ -73,6 +73,62 @@ pj_bool_t BasicProxy::on_rx_request(pjsip_rx_data* rdata)
 }
 
 
+// Statelessly route a response received outside of a transaction.
+pj_bool_t BasicProxy::route_rx_response(pjsip_tx_data *tdata)
```
richardwhiuk: This function always returns true, which suggests it's return value isn't very interesting to those functions above. Can we change it to just return void.

src/basicproxy.cpp
```diff
@@ -73,6 +73,62 @@ pj_bool_t BasicProxy::on_rx_request(pjsip_rx_data* rdata)
...
+    pjsip_tx_data_dec_ref(tdata);
+    return PJ_TRUE;
+  }
+
+  // Calculate the address to forward the response.
+  pjsip_response_addr res_addr;
+  pj_bzero(&res_addr, sizeof(res_addr));
+  res_addr.dst_host.type = pjsip_transport_get_type_from_name(&hvia->transport);
+  res_addr.dst_host.flag =
+                   pjsip_transport_get_flag_from_type(res_addr.dst_host.type);
```
richardwhiuk: Probably space for this to be one line given it's not nested in an if block.

src/basicproxy.cpp
```diff
@@ -81,8 +137,6 @@ pj_bool_t BasicProxy::on_rx_request(pjsip_rx_data* rdata)
 pj_bool_t BasicProxy::on_rx_response(pjsip_rx_data *rdata)
```
richardwhiuk: Huh, this method also always returns true. You could change on_rx_response to be void as well.

src/sproutletproxy.cpp
```diff
@@ -426,6 +426,64 @@ pjsip_sip_uri* SproutletProxy::next_hop_uri(const std::string& service,
 }
 
 
+pj_bool_t SproutletProxy::on_rx_response(pjsip_rx_data *rdata)
+{
+  TRC_DEBUG("Received response after transaction completed.");
```
richardwhiuk: I'd probably do `"... response(%p) after ... ", rdata` and include the response pointer

src/sproutletproxy.cpp
```diff
@@ -426,6 +426,64 @@ pjsip_sip_uri* SproutletProxy::next_hop_uri(const std::string& service,
...
+
+  // Only forward responses to INVITES.
```
richardwhiuk: Worth referencing the RFC which requires us to do this.

include/sproutletproxy.h
```diff
@@ -561,6 +565,7 @@ class SproutletWrapper : public SproutletTsxHelper
   pjsip_sip_uri* next_hop_uri(const std::string& service,
                               const pjsip_sip_uri* base_uri,
                               pj_pool_t* pool) const;
+  pj_bool_t on_rx_response(pjsip_rx_data *rdata);
```
richardwhiuk: Why does this exist?

src/sproutletproxy.cpp
```diff
@@ -426,6 +426,64 @@ pjsip_sip_uri* SproutletProxy::next_hop_uri(const std::string& service,
...
+                                                stack_data.pool,
+                                                false);
+      SproutletMatch match = match_sproutlet_from_uri(uri,
+                                                      alias,
+                                                      local_hostname_unused,
+                                                      selection_type);
+
+      if (match.sproutlet != NULL)
+      {
+        TRC_DEBUG("Removing internal Via");
```
richardwhiuk:
```
const std::string sproutlet = match.sproutlet.service_name();
TRC_DEBUG("Removing internal Via: %s", sproutlet.c_str());
```

Or if I was feeling really adventurous, I'd fix https://github.com/Metaswitch/sprout/blob/dev/include/sproutlet.h#L93 to return a const reference, so we can do:

```
TRC_DEBUG("Removing internal Via: %s", match.sproutlet.service_name().c_str());
```

src/sproutletproxy.cpp
```diff
@@ -426,6 +426,64 @@ pjsip_sip_uri* SproutletProxy::next_hop_uri(const std::string& service,
...
+                PJUtils::pj_status_to_string(status).c_str());
+      return PJ_TRUE;
+      // LCOV_EXCL_STOP
+    }
+
+    // This transaction is complete and so there's nothing more we can do
+    // locally. Remove any local Via headers, then pass on to BasicProxy to
+    // send on.
+    pjsip_via_hdr *hvia = (pjsip_via_hdr*)pjsip_msg_find_hdr(tdata->msg, PJSIP_H_VIA, NULL);
+    while (hvia != NULL)
```
richardwhiuk: I'd probably do:

```c++
for (pjsip_via_hdr *hvia = (pjsip_via_hdr*)pjsip_msg_find_hdr(tdata->msg, PJSIP_H_VIA, NULL);
     hvia != NULL;)
{
  ...
}
```

and remove the two find_hdr lines which are duplicated.

src/sproutletproxy.cpp
```diff
@@ -426,6 +426,64 @@ pjsip_sip_uri* SproutletProxy::next_hop_uri(const std::string& service,
...
+    {
+      std::string local_hostname_unused;
+      std::string alias;
+
+      SPROUTLET_SELECTION_TYPES selection_type = NONE_SELECTED;
+      std::string domain = PJUtils::pj_str_to_string(&hvia->sent_by.host);
+      pjsip_uri* uri = PJUtils::uri_from_string("sip:" + domain,
+                                                stack_data.pool,
+                                                false);
+      SproutletMatch match = match_sproutlet_from_uri(uri,
```
richardwhiuk: Urgh, this function has an irritatingly rubbish interface:

It takes:

- One input parameter
- Three output parameters (one of which isn't marked as such at https://github.com/Metaswitch/sprout/blob/c21d1e9c546edcb575fca0561c5d19d3ac054a42/include/sproutletproxy.h#L176)
- And it returns a custom structure.

As a bonus, the structure returned, has an object which you have to check (though that's almost defensible, given the lack of `std::optional` til C++17

I'd probably refactor `match_sproutlet_from_uri` to fix those flaws...

src/sproutletproxy.cpp
```diff
@@ -426,6 +426,64 @@ pjsip_sip_uri* SproutletProxy::next_hop_uri(const std::string& service,
...
+        // We've only got non-sproutlet Vias left so can continue routing now.
+        break;
+      }
+
+      hvia = (pjsip_via_hdr*)pjsip_msg_find_hdr(tdata->msg, PJSIP_H_VIA, NULL);
+    }
+
+    return  BasicProxy::route_rx_response(tdata);
+  }
+  return PJ_TRUE;
```
richardwhiuk: I might put a new line above this line.

src/ut/scscf_test.cpp
```diff
@@ -2942,6 +2942,121 @@ TEST_F(SCSCFTest, ISCMultipleResponses)
...
+
+  // Send the 200 OK through before the 180 to simulate the messages overtaking.
+  // Confirm stats are set correctly
+  send_response_back_through_dialog(respond_to_txdata(txdata, 200), 200, 2);
+  EXPECT_EQ(1, ((SNMP::FakeEventAccumulatorTable*)_scscf_sproutlet->_audio_session_setup_time_tbl)->_count);
+  EXPECT_EQ(0, ((SNMP::FakeEventAccumulatorTable*)_scscf_sproutlet->_video_session_setup_time_tbl)->_count);
+
+  inject_msg(respond_to_txdata(txdata, 180));
+  out = current_txdata()->msg;
+  RespMatcher(190).matches(out);
```
richardwhiuk: 190? really?? Either this test is failing, or not being run, or RespMatcher is broken. I think this is the singular key part of this test as well!

src/ut/scscf_test.cpp
```diff
@@ -2942,6 +2942,121 @@ TEST_F(SCSCFTest, ISCMultipleResponses)
   pjsip_tx_data_dec_ref(txdata); txdata = NULL;
 }
 
+// Test a 180 being overtaken by a 200
+// request.
+TEST_F(SCSCFTest, Overtaken180)
```
richardwhiuk: This isn't very pleasant. It's quite difficult to see what's different about this test compare to all the others because there's so much noise.... Having said that, it's continuing a long standing tradition in `scscf_test.cpp` which we need to fix as a larger whole I think.

src/sproutletproxy.cpp
kenferguson: That won't work because I need to execute every time.

src/sproutletproxy.cpp
kenferguson: I haven't done this.

src/ut/scscf_test.cpp
kenferguson: Yes - I continued down the path that other code had taken but it would be nice to have less code duplication and more clarity on what the function's doing that's different from usual.

PR #2182, opened 2018-03-15
Build IO trap library for sprout

AlexHockey:
This PR builds an IO trap library that can be used to check if sprout worker threads are doing IO without first calling `CW_IO_*`. 

This PR is not quite complete as I haven't decided how to run this going forwards. There are two options:

* Run sprout from the command line, and explicitly load the IO trap as described in https://github.com/Metaswitch/cpp-common/pull/779. In this case all that's needed is some documentation. 
* Add a shared config option that can be used to automatically load the IO trap when sprout starts. That needs more code and I'm not quite sure how to do it yet. 

However I still think it's worth gettting the code reviewed as it stands. I'll do one of the above either as a separate commit or a separate PR.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,4 +1,4 @@
-TARGETS := sprout call-diversion-as.so gemini-as.so sprout_bgcf.so sprout_icscf.so sprout_mmtel_as.so sprout_scscf.so mangelwurzel-as.so
+TARGETS := sprout call-diversion-as.so gemini-as.so sprout_bgcf.so sprout_icscf.so sprout_mmtel_as.so sprout_scscf.so mangelwurzel-as.so sprout_io_trap.so
 
 TEST_TARGETS := sprout_test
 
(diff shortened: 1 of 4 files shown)


src/Makefile
```diff
@@ -313,6 +313,10 @@ mangelwurzel-as.so_SOURCES := mangelwurzel.cpp mangelwurzelplugin.cpp
 mangelwurzel-as.so_CPPFLAGS := ${PLUGIN_COMMON_CPPFLAGS} -I../include/mangelwurzel
 mangelwurzel-as.so_LDFLAGS := ${PLUGIN_COMMON_LDFLAGS}
 
+sprout_io_trap.so_SOURCES := io_trap.cpp
+sprout_io_trap.so_CPPFLAGS := -fPIC -I../modules/cpp-common/include
+sprout_io_trap.so_LDFLAGS := -shared
```
richardwhiuk: In what way is this Sprout's IO Trap library? We want to use it in Sprout, but it's really part of cpp-common?

Long term, I think we should move to cpp-common building it's own artifacts (e.g. a single cpp_common.o object). It should probably also build this .so file.

AlexHockey: Completely agree that this so should be an artifact that is built from the cpp-common repo. I'm not going to do that under this PR though!

PR #2186, opened 2018-03-28
Add S-CSCF selector tests

KristaHyer:
Our S-CSCF selection UTs were incomplete.
The selecting on weighting functionality wasn't tested, so crass bugs could slip through - for example it wouldn't have been caught if, when choosing based on weightings, the first S-CSCF listed in s-cscf.json was always chosen.
I've added a test which more thoroughly checks this functionality.

I've run `make full_test`, which passed.


KristaHyer:
Don't merge!!
Instead, merge into git.datcon

KristaHyer:
Moved to git.datcon

